PR #2, opened 2019-07-03
Initial structure and tests agains SNAPSHOT

phillip-kruger:
This is now working and testing against the SNAPSHOT of microprofile-graphql-api.


.gitignore
```diff
@@ -0,0 +1,4 @@
+/implementation/target/
```
kenfinnigan: Should only need "target" as opposed to specific directories.

Take a look at the other projects for examples

docs/pom.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kenfinnigan: Can dispense with the whole content of /docs for now as we are in the process of determining a SmallRye wide docs solution

graphql-spqr/pom.xml
```diff
@@ -0,0 +1,302 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
```
kenfinnigan: Embedding `graphql-spqr` is because we need some changes upstream that aren't released?

Do they publish SNAPSHOTs anywhere?

implementation/src/main/java/io/smallrye/graphql/GraphQLExtension.java
```diff
@@ -0,0 +1,82 @@
...
+import io.leangen.graphql.GraphQLSchemaGenerator;
+import io.leangen.graphql.metadata.strategy.query.AnnotatedResolverBuilder;
+
+/**
+ * This is a CDI extension that detects GraphQL components
+ * and generate the GraphQL Schema
+ * 
+ * @author Phillip Kruger (phillip.kruger@redhat.com)
+ */
+public class GraphQLExtension implements Extension {
```
kenfinnigan: Is there a way to do this without needing a CDI Extension?

Thinking from the perspective of what's needed to have it running in Quarkus using ArC

graphql-spqr/pom.xml
phillip-kruger: Yes. There is no SNAPSHOT available. The changes to make it MicroProfile compatible is in a certain branch. This should be released soon, but for now, at least we can continue with creating the MicroProfile TCK

implementation/src/main/java/io/smallrye/graphql/GraphQLExtension.java
phillip-kruger: There might be. I just need to find all classes with a certain annotation on startup. I can look into that at some point.

.gitignore
phillip-kruger: I copied the gitignore from Health to make sure we have all the known ignores.

docs/pom.xml
phillip-kruger: Done

graphql-spqr/pom.xml
kenfinnigan: I think it depends on the definition of "soon".

Ideally we don't want to have an entire project embedded that's going to be removed if it's just a few weeks.

If the time period is reasonably short, I'd prefer we just have a SNAPSHOT dependency for now. CI won't work, but we wouldn't have huge commits for adding/removing the dependency code.

Do we know what the timeframe might be?

implementation/src/main/java/io/smallrye/graphql/GraphQLExtension.java
phillip-kruger: I looked at Jandex. I see that is used in OpenApi. I am going to need more time to see if that can work. And I might need to post on the forum for help. Can we merge like this, so then we can continue with the MP TCK ? Then I create a issue in GitHub to look at Jandex or another alternative. Le me know. Thanks

graphql-spqr/pom.xml
phillip-kruger: I'll find out and get back to you here.

phillip-kruger: Ok so this might still be a while before we see a release for spqr. However, I removed making a copy of all the source, and just added a checkout from the correct branch into target. At least is not committing a bunch of code just to delete it later. Now it's once pom.xml that will be removed later. Let me know if this is better.

phillip-kruger: @kenfinnigan - any feedback ?

implementation/src/main/java/io/smallrye/graphql/GraphQLExtension.java
phillip-kruger: @kenfinnigan - any feedback ?

Issue #36, opened 2020-01-24
Upgrade to latest parent


phillip-kruger:
Upgrading to that latest parent is breaking the build. Investigate and fix

PR #41, opened 2020-02-03
Added implementation for ZonedDateDate, OffsetDateTime and OffsetTime …

phillip-kruger:
…as per spec


phillip-kruger:
Added added fix for issue #36

Issue #64, opened 2020-02-27
Indexer doesn't work in JDK11 or with Wildfly


ribizli:
I tried to integrate this lib into my Wildfly JakartaEE project, but the indexer couldn't find my annotated `@GraphQLApi` classes.

The (open)JDK11 case is reproducible with the example project (https://github.com/phillip-kruger/graphql-example) too.

In Wildfly I've debugged the classpath resolution logic, and found, that only the `jboss-modules.jar` is processed by the indexer.

The only successful execution was with JDK8 and Thorntail.

phillip-kruger:
Thanks. I'll look at this a.s.a.p

jmartisk:
Perhaps including a pre-computed `META-INF/jandex.idx` in the deployment could solve this, but yeah we should probably make sure that the dynamic index computation works with wildfly too

ribizli:
@jmartisk I've tried that too, problem was: `getClass().getResourceAsStream(...)` still not correct, because it looks in the lib not in the war...

phillip-kruger:
@ribizli - You are welcome to fix it and do a PR :) We also need to add test cases that simulate all the scenarios. 
Also (hopefully) soon this will just be part of Wildfly and you would not have to add this lib to your project.

phillip-kruger:
Partially fixed (or at least a workaround) in PR #65. But we keep this open to do a propper fix where we scan the classpath (for jars) and the war or do the scanning in the war.

t1:
The Jandex workaround helps a bit, but when I add a `@Source` method, the `ReflectionDataFetcher` fails with a NPE, because the `methodInfo.declaringClass()` is `null`. Is this a separate bug or even a problem with Jandex?

I suppose fixing this issue #64 would also fix this problem. It seems to be non-trivial, but can I support somehow?

phillip-kruger:
So the fix for JDK 11 happened mostly in the MP Spec TCK. I think there is a bigger issue here with JDK 11 in Runtime. Once Spec 1.0.1 is released (soon) we should be able to at least build against JDK 11 and pass all tests. Then we can investigate why normal running is not working (I know the runner does not work in JDK 11).

phillip-kruger:
One of the things I want to do soon is "removing" Jandex and have an internal model. That would allow Quarkus to create the internal model during build. I should get to this soon, but that *might solve this issue, or at least we can then look at this in detail

jmartisk:
[...]

PR #65, opened 2020-02-27
fix: support META-INF/jandex.idx in the WAR (#64)

ribizli:
I've created a semi-fix: more robust reading of the `META-INF/jandex.idx` which uses the `ClassLoader` of this library,  which is mostly the loader of the WAR.

One needs to generate the `jandex.idx` file with the `org.jboss.jandex:jandex-maven-plugin` maven plugin (or manually).

Tested in Wildfly 18.


phillip-kruger:
Thanks @ribizli !

PR #74, opened 2020-03-13
fix config injection of optional black/whitelist

t1:
When I start `smallrye-graphql` in a WildFly 19, it logs the following warning:

```[io.smallrye.graphql.execution.error.ExceptionLists] (default task-1) Could not create instance of exception class [[]]. Can not do transitive black/whitelist check for this class```

I found that in `GraphQLConfig` the default for the black/whitelist is `[]`, which obviously doesn't work as expected: it creates a list containing the string `"[]"`. So I changed it to an `Optional<List<String>>` instead and it works just fine.

### DIFF
--- a/implementation/src/main/java/io/smallrye/graphql/execution/GraphQLConfig.java
+++ b/implementation/src/main/java/io/smallrye/graphql/execution/GraphQLConfig.java
@@ -15,7 +15,10 @@
  */
 package io.smallrye.graphql.execution;
 
+import static java.util.Collections.emptyList;
+
 import java.util.List;
+import java.util.Optional;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


t1:
Just saw that the TCK fails now... I'm checking.

phillip-kruger:
Awesome ! I also seen that in Quarkus. Would be awesome if you can fix that ! Let me know if I can help with the TCK

t1:
Even on the upstream master, I also get several failures on my machine... due to localization issues, obviously (that's why I skipped the tests before). I guess I'll have to fix them first :-/

phillip-kruger:
Yea, switch to JDK 8, I am busy fixing the locale issues on 11 as we speak

phillip-kruger:
Great ! Thanks very much !! Merging now

t1:
noooo... you where too quick... it still doesn't work :-(

phillip-kruger:
Ha ha. Ok i'll revert

t1:
Sorry... I shouldn't have pushed it ;-) I'll create a new PR

phillip-kruger:
No worries, yes please create a new PR. You might have to sync your fork

PR #76, opened 2020-03-13
fix config injection of optional black/whitelists (second approach)

t1:
second approach to #74 

I've also added logback so I can see the log output.

### DIFF
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -26,9 +26,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
-            <version>1.2.3</version>
+            <groupId>org.jboss.slf4j</groupId>
+            <artifactId>slf4j-jboss-logging</artifactId>
+            <version>1.2.0.Final</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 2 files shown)


t1:
I didn't find an easier way. There seems to be some discussion going on for MP Config 2.0 (https://github.com/eclipse/microprofile-config/issues/446).

tck/pom.xml
```diff
@@ -25,6 +25,12 @@
             <type>pom</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
phillip-kruger: We use jboss logger. Does that not work for you ?

t1: I see a warning in the logs that there is no slf4 binding

phillip-kruger: mmm, wonder what is pulling in slf4j ? Maybe we can just exclude that dep  from whoever is pulling it in ? (less dependencies is better :) )

phillip-kruger: seems like shrinkwrap-resolver-depchain

t1: It's used by shrinkwrap and graphql-java... I assume it's easier to add the jboss-logger-bridge (I'm sure there is one).

t1: `slf4j-jboss-logging`

t1: fixed

phillip-kruger: Cool !

phillip-kruger:
Thanks !

PR #80, opened 2020-03-18
fix #79: more robust creation of collection instances

### LINKED ISSUE #79: Error message logged in CollectionHelper
I see a lot of `ERROR [io.smallrye.graphql.bootstrap.schema.helper.CollectionHelper] (default task-1) Can not create new collection of [java.util.Collections$SingletonList]` and `[java.util.Collections$UnmodifiableRandomAccessList]` messages in my logs. That's because the `CollectionHelper` has a leaky logic for determining what type to create.

A fix should not be very hard. I don't see any tests except for the TCK, so there's nothing to update there, is it? Should I go for a PR?

t1:
It's tougher than I thought (as it always is ;-): I won't manage to make that interface heuristic bullet proof. There are many more interfaces that aren't recognized, e.g. `Queue` or `JsonArray`, and the application can add their own. But maybe it's good enough... at least if it's documented in the spec ;-)


phillip-kruger:
LGTM. Thanks !

PR #96, opened 2020-03-30
[#91] New metrics module for MP Metrics support

### LINKED ISSUE #91: MP Metrics Support
It would be nice to track statistics for GraphQL Query/Mutation invocations using MP Metrics.

andymc12:
This approach uses CDI interceptors, which I understand is less popular in SmallRye?  The reason for choosing this approach is to give users the option of whether or not to collect metrics.  If they want that ability (and have the MP Metrics dependency), then they just need to pull in this new module.  If they don't, then they keep operating as normal - just pulling in the implementation module.

If it comes to it, we could probably replace the interceptor approach by instrumenting the actual implementation code, but my preference would be to leave that separate so that users could run MP GraphQL without needing MP Metrics.

Fixes #91

### DIFF
--- a/metrics/src/main/java/io/smallrye/graphql/metrics/MetricsInterceptor.java
+++ b/metrics/src/main/java/io/smallrye/graphql/metrics/MetricsInterceptor.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2020 IBM Corporation and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.smallrye.graphql.metrics;
+
+import java.lang.reflect.Method;
+
+import javax.annotation.Priority;
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+import javax.interceptor.AroundInvoke;
+import javax.interceptor.Interceptor;
+import javax.interceptor.InvocationContext;
+
+import org.eclipse.microprofile.graphql.GraphQLApi;
+import org.eclipse.microprofile.metrics.MetricRegistry;
+import org.eclipse.microprofile.metrics.MetricRegistry.Type;
+import org.eclipse.microprofile.metrics.SimpleTimer;
+import org.eclipse.microprofile.metrics.SimpleTimer.Context;
+import org.eclipse.microprofile.metrics.annotation.RegistryType;
+import org.jboss.logging.Logger;
+
+@Dependent
+@GraphQLApi
+@Interceptor
+@Priority(Interceptor.Priority.PLATFORM_AFTER)
+public class MetricsInterceptor {
+    private final static Logger LOG = Logger.getLogger(MetricsInterceptor.class);
+    private final static String PREFIX = "mp_graphql_";
+
+    private MetricRegistry metrics;
+
+    @AroundInvoke
+    public Object captureMetrics(InvocationContext ctx) throws Exception {
+        Method m = ctx.getMethod();
+        String fqMethodName = PREFIX + m.getDeclaringClass().getName() + "." + m.getName();
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("invoking " + fqMethodName);
+        }
+        SimpleTimer timer = getMetrics().simpleTimer(fqMethodName);
+        try (Context c = timer.time()) {
+            return ctx.proceed();
+        } finally {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("invoked " + timer.getCount() + " times");
+            }
+        }
+    }
+
+    protected MetricRegistry getMetrics() {
+        return metrics;
+    }
+
+    @Inject
+    @RegistryType(type = Type.VENDOR)
+    protected void setMetrics(MetricRegistry metrics) {
+        this.metrics = metrics;
+    }
+}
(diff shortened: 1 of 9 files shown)


andymc12:
I'm seeing the CI Java 8 build fail - seems like an issue with the logging subsystem in the TCK?  I'm not too familiar with that code, but I don't think my changes should have had any impact on that (my changes shouldn't be involved in the execution of the TCKs tests).  @phillip-kruger any idea what that might be?  I was able to build successfully locally on both Java 8 and 11.  Thanks in advance for any help.

phillip-kruger:
@andymc12 - yes it was broken on the previous PR already. Not sure what is going on, I'll look at it soon. Sorry

andymc12:
@phillip-kruger no problem - I'm just glad to know that I didn't break anything! :-)

metrics/src/main/java/io/smallrye/graphql/metrics/MetricsInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+import org.eclipse.microprofile.graphql.GraphQLApi;
+import org.eclipse.microprofile.metrics.MetricRegistry;
+import org.eclipse.microprofile.metrics.MetricRegistry.Type;
+import org.eclipse.microprofile.metrics.SimpleTimer;
+import org.eclipse.microprofile.metrics.SimpleTimer.Context;
+import org.eclipse.microprofile.metrics.annotation.RegistryType;
+import org.jboss.logging.Logger;
+
+@Dependent
+@GraphQLApi
```
jmartisk: It's not clear to me how you expect the interceptors to get bound to the relevant methods.
And what is this ` @GraphQLApi` annotation for? From first look I thought it might be an interceptor binding, but it's not.

jmartisk:
What I had in mind by saying "without CDI interceptors" was that we could something like this:

Basically there would be a configuration option that turns metrics on/off, and in appropriate places where we want to update metrics, there will be an `if` condition :

- if metrics are off, do nothing
- if metrics are on, then look up the `MetricRegistry` and do whatever you need to update the metrics

This has the advantage that if metrics are turned off, you don't need any dependency on MP Metrics at runtime - that is if you keep all references to the MP Metrics API only inside the said `if` conditions - in this case the code can run without the depenendency and won't face any `NoClassDefFoundError`s.

So with this approach:
- you don't introduce the overhead of CDI interceptors
- you don't need to introduce a mock `MetricRegistry` or anything like that
- you don't need MP Metrics dependency at runtime if metrics are turned off
- you don't need a new module, just add some lines of code into the original one

I'm definitely not saying your approach is wrong, I just have a slight feeling this would be easier and more lightweight :)

metrics/src/main/java/io/smallrye/graphql/metrics/MetricsInterceptor.java
andymc12: Sorry - I tested this in OpenLiberty with the following code used to bind the interceptors:
```
AnnotatedType<GraphQLApi> bindingType = beanManager.createAnnotatedType(GraphQLApi.class);
        beforeBeanDiscovery.addInterceptorBinding(bindingType);

        AnnotatedType<GraphQLMetricsInterceptor> metricsInterceptorType = beanManager.createAnnotatedType(GraphQLMetricsInterceptor.class);
        beforeBeanDiscovery.addAnnotatedType(metricsInterceptorType, CDIServiceUtils.getAnnotatedTypeIdentifier(metricsInterceptorType, this.getClass()));
```

In OL's case, I extended the interceptor so that I could programmatically specify the MetricRegistry.  But I could add this code in a CDI extension that automatically registers it if the module is packaged in the application.  I could also add some sort of `register` method if we wanted users to "opt-in" even if the module is packaged with the app.  I lean toward automatic registration.  Wdyt?

andymc12:
@jmartisk I think that approach will work fine, but we add the risk of a NoClassDefFound if we make a mistake.  And iiuc, we would need to check whether metrics is enabled on every check (once before and again after each request) - if metrics is enabled, this cost is expected, but it would reduce performance unnecessarily if metrics is not enabled.

I still lean toward using the interceptor approach due to it's flexibility, but I can re-write it the other way if that's the group consensus.

metrics/src/main/java/io/smallrye/graphql/metrics/MetricsInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+public class MetricsInterceptor {
+    private final static Logger LOG = Logger.getLogger(MetricsInterceptor.class);
+    private final static String PREFIX = "mp_graphql_";
+
+    private MetricRegistry metrics;
+
+    @AroundInvoke
+    public Object captureMetrics(InvocationContext ctx) throws Exception {
+        Method m = ctx.getMethod();
+        String fqMethodName = PREFIX + m.getDeclaringClass().getName() + "." + m.getName();
```
jmartisk, line 48: This way the metric will be named after the Java method; the actual name of the query can be different (by applying `@Query`/`@Mutation`)

metrics/src/main/java/io/smallrye/graphql/metrics/MetricsInterceptor.java
jmartisk: Including a CDI extension for registering the interceptor would definitely be +1 if we go with the interceptor approach.

However, with that code you would apply the interceptor to *all* methods of the class, we need to detect just the methods with `@Query` or `@Mutation`, so maybe we should use these two annotations as bindings instead?

Registering metrics eagerly at boot would be a good thing too. The CDI extension could scan the annotated types, locate `@Query` and `@Mutation` methods and register metrics for them.

jmartisk:
> I think that approach will work fine, but we add the risk of a NoClassDefFound if we make a mistake. 

Theoretically yes, but I think it shouldn't be a problem to guard this. SmallRye Fault Tolerance does the same thing too.

> And iiuc, we would need to check whether metrics is enabled on every check (once before and again after each request)

The `GraphQLConfig` can just access the config once, and anywhere where we might need the value, we just get it from `GraphQLConfig` and potentially store it into a local boolean variable, because it's not supposed to change dynamically

andymc12:
Overridden by #111

PR #101, opened 2020-04-01
fix #100: introduce JUnit 5 (and assertj)

### LINKED ISSUE #100: Update to JUnit 5

t1:
`smallrye-parent` doesn't contain a version property or a managed dependency for JUnit 5, so I added a property and the dependency to the implementation module directly. The tck uses TestNG instead. The actual change is a different import and the test class and the test methods can be package private.

I also introduced [assertj](https://assertj.github.io/doc/) as that gives nice code and nice failure messages. What do you think?

### DIFF
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -13,6 +13,8 @@
 
     <properties>
         <graphql-java.version>14.0</graphql-java.version>
+        <junit-jupiter.version>5.6.1</junit-jupiter.version>
+        <assertj.version>3.15.0</assertj.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


phillip-kruger:
Hi @t1 - I like it, but I think we need to do this in smallrye-parent. @kenfinnigan can you comment ?

kenfinnigan:
It probably makes sense to add JUnit 5 to parent, so it's managed across all projects in one place.

Also, smallrye-parent defines AssertJ so you should be able to add a dependency without needing the version

phillip-kruger:
Thanks @kenfinnigan ! Does that mean we would need a release of smallrye-parent ?
@t1 would you mind doing it that way ? So a PR on parent with JUnit 5. Then a PR here without the version of both ?

kenfinnigan:
It would yes, but they're easy enough to do

t1:
I will give it a try

phillip-kruger:
@t1 - I am closing this one. Please open one again once parent is released. Thanks

PR #102, opened 2020-04-01
Changes for new Collection creation on transformation

phillip-kruger:

Changes:
1) Only create new collection if necessary (there is transformation on the objects)
2) Make sure that if we have to create a new collection, that we use the type from the method signature and not the object's type.
3) Centralized classloading.

### DIFF
--- a/implementation/src/main/java/io/smallrye/graphql/bootstrap/Classes.java
+++ b/implementation/src/main/java/io/smallrye/graphql/bootstrap/Classes.java
@@ -18,6 +18,9 @@ package io.smallrye.graphql.bootstrap;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.security.AccessController;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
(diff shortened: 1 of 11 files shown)


implementation/src/main/java/io/smallrye/graphql/bootstrap/Classes.java
```diff
@@ -38,59 +45,69 @@
...
+                                // Let's try this class classloader.
+                            }
+                        }
+                        Class<?> c = Class.forName(className, false, Classes.class.getClassLoader());
+                        loadedClasses.put(className, c);
+                        return c;
+                    });
+                } catch (PrivilegedActionException pae) {
+                    throw new ClassloaderException("Can not load class [" + className + "]", pae);
+                }
```
t1, line 79: maybe this block could be extracted into a method, while the `loadedClasses.put(className, c);` happens here.

implementation/src/main/java/io/smallrye/graphql/bootstrap/Classes.java
```diff
@@ -38,59 +45,69 @@
...
  */
 public class Classes {
+    private static final Map<String, Class> loadedClasses = new HashMap<>();
 
     private Classes() {
     }
 
+    public static Class<?> loadClass(String className) {
+
+        if (isPrimative(className)) {
```
t1, line 55: I think this is a type: should be 'primitive'

phillip-kruger: Thanks :) Fixing

implementation/src/main/java/io/smallrye/graphql/bootstrap/Classes.java
phillip-kruger: Yes cool. loadedClasses.put can actually also happen in the method

PR #105, opened 2020-04-01
Add inputTypeTodoList analogous to typeTodoList

ybroeker:
InputTypes for methods with `@Source`-Arguments were added as type-refs, but not created as Types, if not used otherwise.

This PR adds a todo-list for input-types, analogous to typeTodoList, to create those types.


phillip-kruger:
Thanks @ybroeker !
Would you also please add either just a ticket with a description at https://github.com/eclipse/microprofile-graphql/issues , or even better, a ticket with the description and a PR with a test case in the TCK ? Then we know that this is under test.

Thanks !

ybroeker:
Added tests with eclipse/microprofile-graphq#219

PR #107, opened 2020-04-02
Use fields and methods from superclasses while creating types

ybroeker:
Currently, all fields from superclasses are ignored while creating input-types and only the first superclass is considered while creating output-types.

With this PR, all fields and methods from a class and its parents are used.

### DIFF
--- a/implementation/src/main/java/io/smallrye/graphql/bootstrap/type/InputTypeCreator.java
+++ b/implementation/src/main/java/io/smallrye/graphql/bootstrap/type/InputTypeCreator.java
@@ -18,6 +18,7 @@ package io.smallrye.graphql.bootstrap.type;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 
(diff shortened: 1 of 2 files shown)


phillip-kruger:
Thanks @ybroeker - same comment as before, adding a ticket and even better a PR as well to the spec would be great !

Thanks

ybroeker:
Thanks for merging! Added testcases in eclipse/microprofile-graphql#215

phillip-kruger:
Thanks @ybroeker !

phillip-kruger:
Hi @ybroeker . If I test against the MicroProfile GraphQL 1.1-Snapshot I still fail ?!?
I have documented how you can run against the SNAPSHOT version in the README. 
Can you have a look ?

```
87 [SubTypes] (9) - Expecting non-null field from parent in OutputType expected [true] but found [false]
```

ybroeker:
That's the bug fixed in smallrye/smallrye-graphql#114. 

Followed the suggestion from @andymc12 in eclipse/microprofile-graphql#215 and added `@Name` to a field, which exposed this bug.

PR #108, opened 2020-04-02
Use source-arguments only if parameter is annotated as source

ybroeker:
Currently, a query for a field like `Baz getBaz(@Source Foo foo, Bar bar)` fails, if no value for `bar` is given, because the datafetcher tries to use the source for both foo and bar.

It should use the source only, if the parameter is annotated with `@Source`.


phillip-kruger:
Thanks @ybroeker - awesome contribution. Again, same as before, would be great to get this under test in the spec. Thanks !!

ybroeker:
Added some test with eclipse/microprofile-graphql#218

PR #114, opened 2020-04-03
Fix name generation on output-types

ybroeker:
Currently, `@Name` on fields is ignored while generating output-types, this PR fixes this.

(a single test is contained in eclipse/microprofile-graphql#215, I will add some more in a different PR)


ybroeker:
Added some testcases in eclipse/microprofile-graphql#216

PR #133, opened 2020-04-14
Use ISO_DATE_TIME-format for java.util.Date

ybroeker:
`java.util.Date` is mapped to DateTime, but `"yyyy-MM-dd"` is used as format. This PR changes the format to `"yyyy-MM-dd'T'HH:mm:ss"`.

There is no mention of `java.util.Date` in spec and no tests exist, so maybe this was intentionally, but I think the time-part shouldn't be omitted by default.


phillip-kruger:
Thanks @ybroeker  !

Issue #126, opened 2020-04-09
Transformation not working on java.util.Date and java.sql.Date


phillip-kruger:
SmallRye adds support for java.util and java.sql Date, however the transformation does not seem to work.
The test case (in SmallRye) is ignored for now, until this is fixed

PR #134, opened 2020-04-15
Fix date transformation

ybroeker:
This PR fixes smallrye/smallrye-graphql#126, changed both methods where dates are transformed.


phillip-kruger:
Hi @ybroeker - This is so awesome ! Thanks ! I just started looking at this last night in my attempt to replace the Jandex model with this new model, thanks for your help.

PR #137, opened 2020-04-15
Rename person to profile since profile is used in mp-graphql-tck

ybroeker:
Person is used in mp-graphql-tck (sorry, was added be me, didn't see this before), think it's easiest to rename it here?


phillip-kruger:
No worries, I just used the example from https://github.com/phillip-kruger/graphql-example, but even there is a person, so this is perfect.
Thanks !

PR #138, opened 2020-04-15
More complex executions working 

phillip-kruger:
(transformation on input and output).
Also merged the last few PR's


phillip-kruger:
@ybroeker - this PR contains some big changes and I had to manually merge your last few PRs.
Maybe you can have a look at the latest master if your stuff is still correct ? The new model is not yet being used in the execution as it's not passing all the tests in the TCK yet, but the idea is to eventually delete everything in the x package. Thanks for your help :)

ybroeker:
Most things work correctly as far as I see, just some small things.

* Date-format for `java.util.Date` is wrong (see #139).
* input-types are also generated for `@Source`-arguments of methods which are not annotated with  `@Query`, but the input-type is unused and empty (like `XYInput {}`)
* public fields without getter/setter are ignored

phillip-kruger:
Thanks @ybroeker !

- I merged the date format, I must have missed it when syncing my changes.
- That seems wrong, I'll have to look at that. 
- That is correct by the spec, however we can support it in SmallRye if required. Maybe with config to switch it on ?

ybroeker:
Maybe the spec should be clarified in regards to fields and getters/setters?

Currently it only mentions fields:

> Fields in GraphQL are similar to fields in Java in that they are a child of a single entity.  Thus, Java fields on entity classes are, by default, GraphQL fields of that entity.

phillip-kruger:
Good point. You can create an issue in MP Spec and then we can discuss that ?
You are also very welcome to join the spec meetings that is on fridays if you are keen ?

ybroeker:
Yes, I will create an issue there.

I would indeed be interested to join those meetings, it depends on their scheduled time. Could you please provide me with more details?

phillip-kruger:
Great ! You can find some details here. 
https://github.com/eclipse/microprofile-graphql/blob/master/CONTRIBUTING.adoc

It's today at 14h30 GMT

PR #139, opened 2020-04-16
Set correct date-format for `java.util.Date`

ybroeker:


ybroeker:
In the other PRs I only fixed two other places where the wrong format was used but overlooked this one, so nothing to apologies for :)

PR #146, opened 2020-04-17
add type-safe client (api and impl)

t1:


client-implementation/pom.xml
```diff
@@ -11,12 +14,114 @@
...
+            <groupId>org.eclipse.microprofile.graphql</groupId>
+            <artifactId>microprofile-graphql-api</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.12</version>
+            <scope>provided</scope>
+        </dependency>
```
phillip-kruger, line 58: Please remove Lombok (I know, it sucks)

client-implementation/pom.xml
```diff
@@ -11,12 +14,114 @@
...
+            <artifactId>microprofile-config-api</artifactId>
+            <version>1.4</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.ws.rs</groupId>
+            <artifactId>jakarta.ws.rs-api</artifactId>
+            <version>2.1.6</version>
+            <scope>provided</scope>
+        </dependency>
```
phillip-kruger, line 70: This might be temp (until the low level is in) right ?

client-implementation/pom.xml
```diff
@@ -11,12 +14,114 @@
...
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.12</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.microprofile.config</groupId>
+            <artifactId>microprofile-config-api</artifactId>
+            <version>1.4</version>
```
phillip-kruger, line 62: Version should be in the parent pom

client-implementation/pom.xml
```diff
@@ -11,12 +14,114 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.1</version>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
+                    <parameters>true</parameters>
```
phillip-kruger, line 26: Why is this needed ?

client-implementation/pom.xml
```diff
@@ -11,12 +14,114 @@
...
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.2</version>
```
phillip-kruger, line 32: version should be in the pom.

client-implementation/pom.xml
phillip-kruger: version might be in the parent

client-implementation/pom.xml
```diff
@@ -11,12 +14,114 @@
...
+            <version>2.0</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>5.6.1</version>
+            <scope>test</scope>
+        </dependency>
```
phillip-kruger, line 95: In general - check if the version of dependencies is in the parent pom. (Or parent's parent, smallrye-parent)

client-implementation/pom.xml
```diff
@@ -11,12 +14,114 @@
...
+            <artifactId>resteasy-client</artifactId>
+            <version>4.5.3.Final</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.2.3</version>
+            <scope>test</scope>
+        </dependency>
```
phillip-kruger, line 125: Please use Jboss Logger

client-api/src/main/java/io/smallrye/graphql/client/typesafe/api/GraphQlClientBuilder.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.net.URI;
+import java.util.Iterator;
+import java.util.ServiceConfigurationError;
+import java.util.ServiceLoader;
+
+public interface GraphQlClientBuilder {
+
+    static GraphQlClientBuilder newBuilder() {
+        ServiceLoader<GraphQlClientBuilder> loader = ServiceLoader.load(GraphQlClientBuilder.class);
```
phillip-kruger, line 11: Does RESTClient also use SPI ?

client-implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/cdi/GraphQlClientExtension.java
```diff
@@ -0,0 +1,35 @@
...
+import javax.enterprise.inject.spi.AfterBeanDiscovery;
+import javax.enterprise.inject.spi.Extension;
+import javax.enterprise.inject.spi.ProcessAnnotatedType;
+import javax.enterprise.inject.spi.WithAnnotations;
+
+import io.smallrye.graphql.client.typesafe.api.GraphQlClientApi;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class GraphQlClientExtension implements Extension {
```
phillip-kruger, line 16: Quarkus does not do CDI Extensions. We will need to find an alternative here (we can do that later)

client-implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/FieldInfo.java
```diff
@@ -0,0 +1,56 @@
+package io.smallrye.graphql.client.typesafe.impl.reflection;
```
phillip-kruger, line 1: Where possible, use Jandex (rather that your own type). Also think about usage in Quarkus, so what can happen during build and what can not. Example, can this be scanned during build ? Then Jandex is perfect for that. (We can change that later)

client-implementation/pom.xml
t1: done 😢

client-api/src/main/java/io/smallrye/graphql/client/typesafe/api/GraphQlClientBuilder.java
t1: yes. it didn't before, but as the api is in a different module, it was necessary.

client-implementation/pom.xml
t1: Parameter names are used as a fallback when a parameter is not annotated as `@Name`, which I really like. Too bad Oracle hasn't made this option the default.

client-implementation/pom.xml
t1: done

client-implementation/pom.xml
t1: yes

client-implementation/pom.xml
t1: done

client-implementation/pom.xml
t1: done

client-implementation/pom.xml
t1: done

client-implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/cdi/GraphQlClientExtension.java
t1: Maybe we need a Quarkus extension then?

client-implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/FieldInfo.java
t1: When we add Jandex support, I had a strategy pattern in mind: if it's available, use it; otherwise use reflection. The whole reflection package hides all of this already.

client-implementation/pom.xml
phillip-kruger: Thanks :)

client-implementation/pom.xml
phillip-kruger: Jandex should give you the name. But we can add that later.

client-implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/cdi/GraphQlClientExtension.java
phillip-kruger: Yes we would, let's merge like this and then see how we should change for Quarkus.

client-implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/FieldInfo.java
phillip-kruger: If you do it like the server, then the Jandex stuff is in a separate module, and can be included, even in not Quarkus apps. Still easier than using reflection

phillip-kruger: We can change this after merge

PR #150, opened 2020-04-19
Rewrite transforming

ybroeker:
I tried a few things with the transformation and replaced the "old" transfomer classes with a new approach.
Simplified a few things, especially in the helper classes.

Maybe this is interesting?


implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/helper/AbstractHelper.java
```diff
@@ -69,7 +69,7 @@ Object recursiveTransform(Object inputValue, Field field)
         } else if (Classes.isOptional(expectedType)) {
             // Also handle optionals
             return recursiveTransformOptional(inputValue, field);
-        } else if (shouldTransform(field)) {
+        } else {
```
phillip-kruger, line 72: Why did you remove this check ?? (Just so I understand)

implementation/src/main/java/io/smallrye/graphql/scalar/time/DateCoercing.java
```diff
@@ -16,12 +16,10 @@
 public class DateCoercing implements Coercing {
 
     private final Class[] supportedTypes;
-    private final Converter converter;
     private final String name;
 
-    public DateCoercing(String name, Converter converter, Class... supportedTypes) {
+    public DateCoercing(String name, Class... supportedTypes) {
```
phillip-kruger, line 21: Yea, this is here for future use, but we can add that back then. I was playing with fixing [#225](https://github.com/eclipse/microprofile-graphql/issues/225) in the spec.
At the moment Dates just fall through due to transformation, but if we change dates with transformation to be Strings, then this can create the correct date instance.

implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/helper/AbstractHelper.java
ybroeker: Transformation is also used for types that have no TransformInfo (for example `UUID`), and instead of extending shouldTransform accordingly, the logic is contained in `Transformer#transform`. If no transformation is needed, `PassThroughTransformer` is used where no transformation is performed, so this check is unnecessary.

phillip-kruger: Cool !

implementation/src/main/java/io/smallrye/graphql/scalar/time/DateCoercing.java
ybroeker: I think it should be enough to change the type of formatted dates in the schema to string, because dates are formatted in any case (either in the format given with `@DateFormat or with ISO_TIME, ISO_DATE, etc).

ybroeker: Tried this in https://github.com/ybroeker/smallrye-graphql/commit/06617144cc2d20faa585485dd15cd3b43b042e07, numbers and dates with `@DateFormat`-format (or equivalent) are included as Strings and are formatted correctly.

TCK currently fails because of eclipse/microprofile-graphql#225 and eclipse/microprofile-graphql#209, but once it is specified and tests are updated it should work.

PR #178, opened 2020-04-22
Include scalars in schema

ybroeker:
Currently, scalars such as 'Date' are not included in the schema, and therefore some tools cannot validate it, eg the GraphQL plugin for Intellij.

This PR adds scalars to the schema.


phillip-kruger:
Hi @ybroeker ! Thanks for this. How about we love that to config, with some default values. We can move more options in schema printer there.

ybroeker:
Something like this?

Added all options except `descriptionsAsHashComments` and `useAstDefinitions`, first one is deprecated in GraphQL-Spec and last one is not useful in this case.

phillip-kruger:
Perfect !! Thanks @ybroeker

PR #180, opened 2020-04-22
Add ctor for environments where CDI is unavailable

andymc12:
If packaging the servlet implementation as part of the Liberty runtime, CDI injection is not available, so constructors are necessary.  This change should allow the servlet to work with or without CDI.


phillip-kruger:
Why did the build fail ? @andymc12 can you have a look?

phillip-kruger:
@andymc12 - it's  formatting, please can you build on localhost then commit again

PR #193, opened 2020-04-25
[#187] First pass at QueryCache

### LINKED ISSUE #187: Query cache
Look into a query cache to optimize the query execution.
assigned: @andymc12

andymc12:
LRU-based cache with a default size of 2048 queries/mutations, but can be changed by setting the system property `io.smallrye.graphql.execution.queryCacheMaxSize=4096`.

The cache allows a query that has been validated to skip validation - since the validation does not check variables, that is the ideal use case (or repeated invocations of the same query with the same arguments).

Invalid queries are not cached, but queries or mutations that result in an exception will still be cached.  My reasoning for this is that some exceptions are point-in-time-related - for example, suppose I want to buy a new guitar that costs $500, so I invoke `mutation withdraw(500.00)`, but since I only have $100 in my account, the server throws an InsufficientFundsException.  Fortunately today is a pay day, and my company deposits $500 into my account, so when I go back and call `mutation withdraw(500.00)`, it succeeds - and that operation was cached! :-)

Please don't merge just yet - while the code changes don't break anything, I'm not convinced that they improve performance much, so I'd like to hear back from some benchmark tests before merging.  Thanks!

Fixes #187

### DIFF
--- a/implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
+++ b/implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
@@ -1,0 +1,98 @@
+package io.smallrye.graphql.execution;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+public class LRUCache<K, V> {
+
+    private final int maxSize;
+    private final Map<K, Entry<V>> cache = new HashMap<>();
+    private Entry<V> start;
+    private Entry<V> end;
+
+    LRUCache(int maxSize) {
+        this.maxSize = maxSize;
+    }
+
+    V get(K key) {
+        Entry<V> entry = cache.computeIfPresent(key, this::moveEntryToStart);
+        return entry == null ? null : entry.value;
+    }
+
+    V put(K key, V value) {
+        Entry<V> entry = cache.get(key);
+        V oldValue = null;
+        if (entry == null) {
+            entry = cachePutNew(key, value);
+        } else {
+            entry = removeEntry(entry);
+            oldValue = entry.value;
+            entry.value = value;
+        }
+        entry = addToStart(entry);
+        return oldValue;
+    }
+
+    V computeIfAbsent(K key, Function<K, V> valueFunction) {
+        Entry<V> entry = cache.get(key);
+        if (entry == null) {
+            entry = cachePutNew(key, valueFunction.apply(key));
+        } else {
+            entry = removeEntry(entry);
+        }
+        entry = addToStart(entry);
+        return entry.value;
+    }
+
+    private Entry<V> cachePutNew(K key, V value) {
+        Entry<V> entry = new Entry<V>(key, value);
+        cache.put(key, entry);
+        if (cache.size() > maxSize) {
+            cache.remove(end.key);
+            removeEntry(end);
+        }
+        return entry;
+    }
+
+    private synchronized Entry<V> moveEntryToStart(K key, Entry<V> entry) {
+        return addToStart(removeEntry(entry));
+    }
+
+    private synchronized Entry<V> removeEntry(Entry<V> entry) {
+        if (entry.left != null) {
+            entry.left.right = entry.right;
+        }
+        if (entry.right != null) {
+            entry.right.left = entry.left;
+        } else {
+            end = entry.left;
+        }
+        return entry;
+    }
+
+    private synchronized Entry<V> addToStart(Entry<V> entry) {
+        entry.right = start;
+        entry.left = null;
+        if (start != null) {
+            start.left = entry;
+        }
+        start = entry;
+        if (end == null) {
+            end = start;
+        }
+        return entry;
+    }
+
+    private class Entry<V> {
+        final K key;
+        V value;
+        Entry<V> left;
+        Entry<V> right;
+
+        Entry(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+    }
+}
--- a/implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
+++ b/implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
@@ -1,13 +1,13 @@
 package io.smallrye.graphql.execution;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
 public class LRUCache<K, V> {
 
     private final int maxSize;
-    private final Map<K, Entry<V>> cache = new HashMap<>();
+    private final Map<K, Entry<V>> cache = new ConcurrentHashMap<>();
     private Entry<V> start;
     private Entry<V> end;
 
@@ -48,9 +48,11 @@ public class LRUCache<K, V> {
     private Entry<V> cachePutNew(K key, V value) {
         Entry<V> entry = new Entry<V>(key, value);
         cache.put(key, entry);
-        if (cache.size() > maxSize) {
-            cache.remove(end.key);
-            removeEntry(end);
+        synchronized (cache) {
+            if (cache.size() > maxSize) {
+                cache.remove(end.key);
+                removeEntry(end);
+            }
         }
         return entry;
     }
--- a/implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
+++ b/implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
@@ -20,36 +20,18 @@ public class LRUCache<K, V> {
         return entry == null ? null : entry.value;
     }
 
-    V put(K key, V value) {
-        Entry<V> entry = cache.get(key);
-        V oldValue = null;
-        if (entry == null) {
-            entry = cachePutNew(key, value);
-        } else {
-            entry = removeEntry(entry);
-            oldValue = entry.value;
-            entry.value = value;
-        }
-        entry = addToStart(entry);
-        return oldValue;
-    }
-
     V computeIfAbsent(K key, Function<K, V> valueFunction) {
-        Entry<V> entry = cache.get(key);
-        if (entry == null) {
-            entry = cachePutNew(key, valueFunction.apply(key));
-        } else {
-            entry = removeEntry(entry);
-        }
-        entry = addToStart(entry);
+        Entry<V> entry = cache.computeIfAbsent(key, k -> {
+            return cachePutNew(k, valueFunction.apply(k));
+        });
+        entry = moveEntryToStart(key, entry);
         return entry.value;
     }
 
     private Entry<V> cachePutNew(K key, V value) {
         Entry<V> entry = new Entry<V>(key, value);
-        cache.put(key, entry);
         synchronized (cache) {
-            if (cache.size() > maxSize) {
+            if (cache.size() > maxSize - 1) {
                 cache.remove(end.key);
                 removeEntry(end);
             }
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
```diff
@@ -0,0 +1,102 @@
...
+
+        @Override
+        public void onDispatched(CompletableFuture<List<ValidationError>> result) {
+            // no-op
+        }
+
+        @Override
+        public void onCompleted(List<ValidationError> validationErrors, Throwable t) {
+            // at this point, we know the validation is complete - go ahead and add it to the cache if no errors
+            if (executionFunction != null && validationErrors == null || validationErrors.isEmpty()) {
```
ybroeker, line 93: Is the `executionFunction != null` necessary? 
ValidationInstrumentationContext is (currently) only created if the executionFunction is not `null`, and even if it were `null`, it would result in a NullPointer on line 95 if `validationErrors.isEmpty()` is true.

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
```diff
@@ -0,0 +1,102 @@
...
+    private final ThreadLocal<ExecutionFunction> executionFunctionTL = new ThreadLocal<>();
+
+    @Override
+    public PreparsedDocumentEntry getDocument(ExecutionInput executionInput,
+            Function<ExecutionInput, PreparsedDocumentEntry> computeFunction) {
+        String query = executionInput.getQuery();
+        PreparsedDocumentEntry entry = cache.get(query);
+        if (entry == null) {
+            ExecutionFunction executionFunction = new ExecutionFunction(computeFunction, executionInput);
+            functionCache.putIfAbsent(query, executionFunction);
```
ybroeker, line 38: entries are only added, but never queried or deleted?

implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
```diff
@@ -0,0 +1,98 @@
...
+        if (entry == null) {
+            entry = cachePutNew(key, valueFunction.apply(key));
+        } else {
+            entry = removeEntry(entry);
+        }
+        entry = addToStart(entry);
+        return entry.value;
+    }
+
+    private Entry<V> cachePutNew(K key, V value) {
```
ybroeker, line 48: Shouldn't these methods be synchronized like the others, because `HashMap#get`, `#put` and `#remove` aren't threadsafe?

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
andymc12: Good catch - I've removed this.

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
andymc12: The functionCache was an attempt to avoid the thread local, but ultimately wasn't feasible.  I removed most of that attempt, but left this in there.  Thanks for catching it.  I've removed it now too.

implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
andymc12: I've made `cache` to be a ConcurrentHashMap, and put a sync block around the remove case.  I think the rest can be left un-sync'd for better performance.  Wdyt?

ybroeker: lgtm :)

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
```diff
@@ -0,0 +1,98 @@
...
+
+        @Override
+        public void onDispatched(CompletableFuture<List<ValidationError>> result) {
+            // no-op
+        }
+
+        @Override
+        public void onCompleted(List<ValidationError> validationErrors, Throwable t) {
+            // at this point, we know the validation is complete - go ahead and add it to the cache if no errors
+            if (validationErrors == null || validationErrors.isEmpty()) {
```
jhanders34, line 89: Should
```
t != null
``` 
be part of the criteria to cache?

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
```diff
@@ -0,0 +1,98 @@
...
+import graphql.execution.preparsed.PreparsedDocumentProvider;
+import graphql.validation.ValidationError;
+
+public class QueryCache extends SimpleInstrumentation implements PreparsedDocumentProvider {
+    private static final Logger LOG = Logger.getLogger(QueryCache.class);
+    private static final int MAX_CACHE_SIZE = AccessController.doPrivileged((PrivilegedAction<Integer>) () -> {
+        return Integer.getInteger("io.smallrye.graphql.execution.queryCacheMaxSize", 2048);
+    });
+
+    private final LRUCache<String, PreparsedDocumentEntry> cache = new LRUCache<>(MAX_CACHE_SIZE);
```
jhanders34, line 25: is there ever only one instance of this class?  if not, should the ThreadLocal at least be static?  Usually they are to prevent multiple ones being created.  I would recommend making it static regardless.  Not sure if the LRUCache should be static. as well.

Have you measured the size of caching the objects?  Is 2048 an ok size?  If each String and PreparedDocumentEntry is 1K in size it is only 2 MB, 10K then 20 MB, etc.

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
```diff
@@ -0,0 +1,98 @@
...
+        return entry;
+    }
+
+    @Override
+    public InstrumentationContext<List<ValidationError>> beginValidation(
+            InstrumentationValidationParameters parameters) {
+
+        ExecutionFunction executionFunction = executionFunctionTL.get();
+        if (executionFunction != null) {
+            executionFunctionTL.remove();
```
jhanders34, line 49: Not sure if you want to move the remove above the if statement.  Even if it is null it will still keep an entry in the ThreadLocal with a value of null.

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
```diff
@@ -0,0 +1,98 @@
...
+        public PreparsedDocumentEntry apply(String s) {
+            return function.apply(executionInput);
+        }
+
+        String getQuery() {
+            return executionInput.getQuery();
+        }
+    }
+
+    private class ValidationInstrumentationContext implements InstrumentationContext<List<ValidationError>> {
```
jhanders34, line 74: If you made the cache static from the comment above, you could make this inner class static as well.

implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private synchronized Entry<V> moveEntryToStart(K key, Entry<V> entry) {
+        return addToStart(removeEntry(entry));
+    }
+
+    private synchronized Entry<V> removeEntry(Entry<V> entry) {
+        if (entry.left != null) {
+            entry.left.right = entry.right;
+        }
```
jhanders34, line 67: Do you need an else here?
```
else {
    start = entry.right;
}
```

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
andymc12: I can add that.  Thanks

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
andymc12: No, there will be one instance of the cache for every GraphQL application.  For containers like Quarkus, that means that there will only be one, but for containers like Open Liberty or Wildfly, etc. there could be more than one.  I think it needs to be non-static.

implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
```diff
@@ -0,0 +1,100 @@
...
+    LRUCache(int maxSize) {
+        this.maxSize = maxSize;
+    }
+
+    V get(K key) {
+        Entry<V> entry = cache.computeIfPresent(key, this::moveEntryToStart);
+        return entry == null ? null : entry.value;
+    }
+
+    V put(K key, V value) {
```
jhanders34, line 23: This method is not thread safe.  NPEs and other errors are possible.  It isn't used at all.  Should it just be removed?

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
andymc12: Will do

implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+            entry = removeEntry(entry);
+            oldValue = entry.value;
+            entry.value = value;
+        }
+        entry = addToStart(entry);
+        return oldValue;
+    }
+
+    V computeIfAbsent(K key, Function<K, V> valueFunction) {
```
jhanders34, line 37: This isn't thread safe either.  It would seem you should use 
call 
```
cache.computeIfAbsent(...);
```

implementation/src/main/java/io/smallrye/graphql/execution/LRUCache.java
andymc12: Maybe for completeness, but it shouldn't be required.  Every time that we remove and entry it is later added to the start of the list, so the `start` field is reset. The one exception to that is when the cache is full, then we always remove from the end - so the only time we would need to reset `start` in this method is if the max cache size were set to 0 or lower - but in that case, we would never use `start` or `end`.  I have no problem adding the else block, but I don't think it will ever be used.

jhanders34: If remove is called and it is the only thing in the cache, then start should be set to null.  Without the else, it will be left set to the one entry that was in the cache.

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
jhanders34: ThreadLocal should still be static.

implementation/src/main/java/io/smallrye/graphql/execution/QueryCache.java
```diff
@@ -86,7 +87,7 @@ public void onDispatched(CompletableFuture<List<ValidationError>> result) {
         @Override
         public void onCompleted(List<ValidationError> validationErrors, Throwable t) {
             // at this point, we know the validation is complete - go ahead and add it to the cache if no errors
-            if (validationErrors == null || validationErrors.isEmpty()) {
+            if (t != null && (validationErrors == null || validationErrors.isEmpty())) {
```
jhanders34, line 90: I was wrong it should be if (t == null)

andymc12:
@ybroeker @phillip-kruger I ran this change by the performance team - and implemented a few more of their suggestions - both for performance and for thread-safety - and now it should be ready to merge.  Can you take another (hopefully final) look?  Thanks!

phillip-kruger:
Hi @andymc12 - Thanks for this. Was there a significant performance improvement ?

andymc12:
> Thanks for this. Was there a significant performance improvement ?

Ultimately it will depend on the usage, but in one benchmark (where the same query is invoked), the performance improved 30% (transactions per second).  

To take the most effective use of the query cache, clients should try to re-use queries/mutations as much as possible - using the same query with different variables will still take advantage of the cache but use different arguments.

PR #199, opened 2020-04-27
Refactor MetricsService to own interface

andymc12:
- refactored MetricRegistry into it's own service (from LookupService)
- req'd changes in CDI implementation module and impl
- new test cases

These changes should make it easier for containers (like Open Liberty) to separate out the Metrics implementation.


ybroeker:
tried something similar in https://github.com/ybroeker/smallrye-graphql/tree/feat/opentracing, mainly to use other metrics-versions (unfortunately we have to use an older version) and to connect OpenTracing/OpenTelemetry.

Maybe there is a way to merge them?

phillip-kruger:
Thanks @andymc12 !

phillip-kruger:
@ybroeker - we should be able to follow the same pattern for tracing, so have a Tracing service interface ?

phillip-kruger:
Thanks @andymc12 !

PR #202, opened 2020-04-28
Added Gradle plugin

marceloverdijk:
This PR adds a Gradle plugin to generate the schema (similar and with the same config options as the Maven plugin).

Note this is the first time I created a Gradle plugin (created Gradle project tasks before).
And in this case specifically a Gradle plugin in a Maven based project.

The code might be a bit rough on the edges so there might be cases that won't work yet.
Note I installed the plugin in my local Maven repo and generated the schema file successfully from an [example](https://github.com/marceloverdijk/graphql-example) project. The Gradle build file is [here](https://github.com/marceloverdijk/graphql-example/blob/master/person-example/build.gradle) and it can be run with `./gradlew clean compileJava generateSchema`.

Some notes:

- The `gradle-plugin` module contains both a `pom.xml` and a Gradle build setup. This is needed to be able to publish the plugin to the Gradle plugin portal. Note that the setup is the same as the Quarkus Gradle plugin setup and also how Spring Boot was doing it before they fully switched to Gradle as build tool.
- For publishing to the Gradle plugin portal, I don't know if we have to create an account or something like that. Maybe we can ask help from Quarkus devs? It would be good if someone could review the code + setup anyway.
- There is some code duplication with Maven plugin so we could maybe introduce a `devtools-commons` module to share code between both plugins.
- Currently both the Maven and Gradle plugin provide a single classesDir (`File`). In Gradle it is possible to define multiple sourcesets per module, so we might need to support that as well.
- At least it is a start.


jmartisk:
Very cool!

PR #204, opened 2020-04-28
[#187] Followup

andymc12:
@jhanders34 found another couple issues with the LRUCache and suggested these changes - related to #187.
- additional thread safety
- ensuring the cache never exceeds max size


jhanders34:
The major issue that is fixed is in the cache.computeIfAbsent lambda.  moveToStart was called instead of addToStart.  As such the queue was all messed up making it looks like there was nothing we could remove from the cache because we thought there was only one thing in it.

PR #207, opened 2020-04-29
Added two more options when including dependencies in the maven plugin

phillip-kruger:

@marceloverdijk - maybe we want to do this in the gradle plugin as well ? Not sure.

### DIFF
--- a/maven-plugin/README.adoc
+++ b/maven-plugin/README.adoc
@@ -27,6 +27,8 @@ The schema will appear as `target/generated/schema.graphql` by default.
 
 - `destination` - To override the default `target/generated/schema.graphql` destination
 - `includeDependencies` - Scan project's dependencies for GraphQL model classes too. Off (false) by default.
+- `includeDependenciesScopes` - If the above `includeDependencies` is true, you can control what scopes should be included. Default is `compile, system`
+- `includeDependenciesTypes` - If the above `includeDependencies` is true, you can control what types should be included. Default is `jar`
 - `includeScalars` - Include scalars in the schema. Default false.
 - `includeDirectives` - Include directives in the schema. Default false.
 - `includeSchemaDefinition` - Include the schema definition. Default false.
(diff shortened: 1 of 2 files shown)


marceloverdijk:
@phillip-kruger Yes I will add something similar for the Gradle plugin.
In Gradle terms this will be something like `includeDependenciesConfigurations`. I will create a PR this evening.

phillip-kruger:
THanks @jmartisk. Thanks @marceloverdijk .

marceloverdijk:
@phillip-kruger I looked into adding these options to the Gradle plugin as well, but it is not that easy. In Gradle it seems difficult to get all dependencies for a certain scope/configuration with a handle to the actual file.
I will continue, but it will take a bit more time unfortunately.

phillip-kruger:
No worries :) I just wanted to make sure you know. We can create an issue and do this later

marceloverdijk:
I think I found a solution and work it out this this evening.

But probably I will also add another option `includeDependenciesRecursive` (or  `includeDependenciesTransitive`) so only direct specified dependencies will be index (and not the transitive ones). Maybe good to add to the Maven plugin as well ;-)

phillip-kruger:
Sounds good ! Thanks :)

marceloverdijk:
In Gradle this will provide a nice option to do:

```
configurations {
    schema // we add a custom configuration
}

dependencies {
    implementation "foo:bar:1.0"
    implementation "com:example-with-types-to-be-indexed:1.0"
    schema "com:example-with-types-to-be-indexed:1.0"
}

```

and then generate the schema something like 
```
./gradlew generateSchema --include-dependencies 
            --include-dependencies-configrations=schema 
            --include-dependencies-transitive=false 
```

marceloverdijk:
Creates https://github.com/smallrye/smallrye-graphql/pull/211 for the additional Gradle options + a bit of refactoring.

PR #208, opened 2020-04-29
Some metrics changes

ybroeker:
* include all queries/fields which are updated by MetricDecorator in initial registration
* include type-name in metric-name, prevents a single metric for multiple fields
* fix the test which tested metrics on mutations


phillip-kruger:
Thanks @ybroeker ! Awesome as always !

PR #215, opened 2020-05-02
add client-generator

t1:
I've built a PoC of an annotation processor to generate Type-Safe Client APIs. An example is `client-generator-test/src/main/java/io.smallrye.graphql.client.generator.test.SuperHeroes`, which looks like this:

```java
@GraphQlSchema("resource:schema.graphql")
@GraphqlQuery("{ teams { name } }")
@GraphqlQuery("query heroesLocatedIn($location: String) { heroesIn(location: $location) { name realName superPowers } }")
public class SuperHeroes {
}
```

And the annotation processor generates a class `SuperHeroesApi ` like this:

```java
public interface SuperHeroesApi {
    List<Team> teams();
    @Query("heroesIn") List<SuperHero> heroesLocatedIn(String location);
}
```

Plus the POJOs `Team` and `SuperHero`.

What do you think?

IMPORTANT: Please don't merge, yet!


phillip-kruger:
Interesting ! So when does the code generate  ? On build ?

t1:
Yeah, it's a annotation processor, so it runs whenever the class is compiled

t1:
Ooops I did it again 🤪 I accidentally used JUnit 5 again. I will fix it before we merge this or build on #100... whichever comes first ;-)

ybroeker:
I only see the problem (apart from some bugs) that the types are not generated appropriately for the respective query.

For example:
```java
@GraphqlQuery("query teamNames {team { name }}")
@GraphqlQuery("query teamSizes {team { size }}")
```

would generate this class:
```java
class Team {
  @NonNull Integer size;
  String name;
}
```

Depending on the query, however, only one of the two fields is used. The other field is `null`, even if the annotation says otherwise.


But honestly, I don't know a suitable solution for this. Generating exactly matching types would result in too many classes with the same name and would probably be unusable.



Apart from that, I would like to see `Optional` and primitives (for nonnull Int/Float/Boolean) used, but this is just personal preference.

t1:
@ybroeker : Thanks for the feedback. Currently the second query overwrites the `Team` class generated for the first... they wouldn't be merged, which, as you said, would be even worse.

Actually we have exactly the same problem with a hand written Client API. Different queries that return different variations of the same type need to have different Java types. See https://github.com/t1/graphql-java-experiment/blob/ecc779b14d4ac2ec95063c8d68f0f5afd810be58/src/test/java/it/SuperHeroesIT.java#L39 for an example.

At first I was thinking about some automatic type name generation when required, i.e. `TeamNamesTeam` and `TeamSizesTeam` or so. But that would not be stable, i.e. if both return the same fields, they should be the same class, but if you then add a field to one, the code using the other would break... all alternatives I thought about where not very pleasing, either. And I think names are just too important for clean code to let them be generated like that.

My current plan is to a) mark it as an error if two return types are the same but with different fields, and b) to allow you to explicitly set the type name. I was considering to add it to the query, e.g. `query teamNames {teams { name }}: TeamWithName`, or as a separate method in the annotation, i.e. `@GraphqlQuery(value = "query teamNames {team { name }}", type = "TeamWithName")`. What would you prefer?

ybroeker:
> Currently the second query overwrites the Team class generated for the first... they wouldn't be merged

Sorry, I forgot to mention that it doesn't work yet because the second query overwrites it. The rest was about how it "should be" if it wouldn't just overwrite it.

> At first I was thinking about some automatic type name generation when required, i.e. TeamNamesTeam and TeamSizesTeam or so. But that would not be stable, i.e. if both return the same fields, they should be the same class, but if you then add a field to one, the code using the other would break... all alternatives I thought about where not very pleasing, either. And I think names are just too important for clean code to let them be generated like that.
>
> My current plan is to a) mark it as an error if two return types are the same but with different fields, and b) to allow you to explicitly set the type name. I was considering to add it to the query, e.g. `query teamNames {teams { name }}: TeamWithName`, or as a separate method in the annotation, i.e. `@GraphqlQuery(value = "query teamNames {team { name }}", type = "TeamWithName")`. What would you prefer?


I thought of that at first, but that would not be nice for such simple queries and would not work for nested ones.

For something like this:
```graphql
{
  teams { name, size}
  heroes {teamAffiliations } { name }
  teamsLargerThan(size: 3) { size }
}
```

three different "team" classes would have to be generated, depending on the context in the one query.


This also speaks against a in my opinion. That a type is used several times with different fields is a valid use case, and that would mean that you either specify more fields than needed, or such queries are simply not possible.

I am not sure how far proposal b would be possible. Perhaps directives could be used for this?

ybroeker:
two random ideas:

"Abuse" Optionals for this. 
`null` would mean "not contained", `Optional.empty` would mean "contained, but `null`".
For non-null fields, instead of Optionals the normal type would be used.


Or use (java-) interfaces with getters and thrown an exception if an not-requested field is accessed.

For this query:
```graphql
{
  teams { name, size}
  heroes {teamAffiliations} { name }
  teamsLargerThan(size: 3) { size }
}
```

One interface would be generated:
```java
interface team {
  String getName();
  int getSize();
}
```

And  three different implementations depending on the context, for example:

```java
class TeamsLargerThanTeam implements Team {
  int size;
  String getName() {
    //throw an exception
  }
  int getSize() {
    return size;
  }
}
```

Names would not really matter, users would only work with the interface. Wouldn't help at compile time of course, only at runtime.

t1:
I must admit that I hadn't thought enough about nested types. But I still think it's important to name types explicitly... also for your multi-query example; even if it ends up with something as generic as `MainPage`, it does convey a lot of information to the reader.

The type name from the schema would be a fallback, as long as all non-explicitly-typed queries select the same fields. And if they don't, an error would be raised (I actually added that in the last commit e9bba28) and you have to rename all but one.

The syntax I'm thinking of would look like this for a slightly fixed version of your example:

```
{
  teams { name size }
  heroes { teamAffiliations { name }}: TeamAffiliations
  teamsLargerThan(size: 3) { size }: TeamSizes
}: MainPage
```

The `SuperHero` types in line 2 is unique, but doesn't make a lot of sense, so it's better to rename it. After renaming the `Team` type in line 3 to `TeamSizes`, it's unique for line 1. Line 4 doesn't have a type in the Schema, so it always has to be renamed.

As this is an extension to the query syntax it can't be parsed by `graphql-java`; I guess we'll have to write our own parser. This is more work than I hoped for the first iteration. Maybe we can start with a very limited generator? You'd have to use multiple API classes in case the selected fields don't match everywhere. What do you think?

t1:
> "Abuse" Optionals for this.
null would mean "not contained", Optional.empty would mean "contained, but null".
For non-null fields, instead of Optionals the normal type would be used.

This logic wouldn't be obvious. I prefer to stick to the principle of least surprise :-)

t1:
Should we and how should we continue with this?

phillip-kruger:
I am not sure. Maybe move it to client/generator and then we can merge. If we see it's not being used we can remove it later ?

t1:
I've rebased it to the current master and added a README to be explicit about this being experimental.

t1:
no, the integration test has to be a separate module, as the annotation processor has to run

phillip-kruger:
@t1 I broke this .

t1:
no problem ;-)

phillip-kruger:
Thanks ! :)

PR #218, opened 2020-05-04
Fixes for Spec update on formatting.

phillip-kruger:
This PR aligns with https://github.com/eclipse/microprofile-graphql/pull/240

@ybroeker please have a look as I have changed some stuff that you have done.

All logic on what happens on transformation moved to the implementation (and out of the model builder).
This will allows us to change the decision on making the schema types String scalars to something else in the future.

I also removed the "null" transformers that translate to default and made the default decision in the implementation.

This is still a draft PR until 1.0.2 of the spec release.


ybroeker:
lgtm :)

implementation/src/main/java/io/smallrye/graphql/transformation/Transformer.java
```diff
@@ -16,31 +16,49 @@
...
 
     static Transformer transformer(Field field) {
-        if (field.hasTransformInfo()) {
-            TransformInfo format = field.getTransformInfo();
-            if (format.getType().equals(TransformInfo.Type.NUMBER)) {
-                if (format.getFormat() != null || format.getLocale() != null) {
-                    return new FormattedNumberTransformer(field);
-                }
-                return new NumberTransformer(field);
+        if (shouldTransform(field)) {
```
andymc12, line 19: I would remove this check.  The following if/else if blocks cover all of the conditions that `shouldTransform` checks, so it is essentially checking those conditions twice.

pom.xml
```diff
@@ -17,7 +17,7 @@
 
     <properties>
         <version.eclipse.microprofile.config>1.4</version.eclipse.microprofile.config>
-        <version.eclipse.microprofile.graphql>1.0.1</version.eclipse.microprofile.graphql>
+        <version.eclipse.microprofile.graphql>1.0.2</version.eclipse.microprofile.graphql>
```
andymc12, line 20: We'll have to wait for 1.0.2 release before we can merge this - but hopefully that will be very soon!

PR #228, opened 2020-05-05
JUnit 4 -> JUnit 5 Vintage

t1:
It still depends on smallrye-parent 19-SNAPSHOT, so it won't build on GitHub and shouldn't be merged before that was released.


t1:
Should fix #100

t1:
If I can find the time for it, I may update some tests to JUnit 5 (and assertj)

phillip-kruger:
@t1 - There is some Conflicting files

t1:
sorry... I just don't grok it :-/

PR #238, opened 2020-05-06
Add OpenTracing support to server

ybroeker:
A first try for OpenTracing support.

Tracing can be activated via `smallrye.graphql.tracing.enabled`.
The entire execution of a query and the fetching of the respective fields is then traced. Informations about current field, current Type, path to field etc are added to the span. Spans of `@Source' methods are added as children to the span of the respective parent object.

jaeger_tracing_example

Currently, the SpanContext of the request is not yet used. If the GraphQL service is called by another service, it will not be assigned appropriately.

Comments are welcome!

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionDecorator.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionDecorator.java
@@ -1,0 +1,18 @@
+package io.smallrye.graphql.execution;
+
+import graphql.ExecutionInput;
+import graphql.ExecutionResult;
+
+public interface ExecutionDecorator {
+    default void before(ExecutionInput executionInput) {
+
+    }
+
+    default void after(ExecutionInput executionInput, ExecutionResult executionResult) {
+
+    }
+
+    default void onError(ExecutionInput executionInput, Throwable throwable) {
+
+    }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,8 @@
         <version.eclipse.microprofile.graphql>1.0.2</version.eclipse.microprofile.graphql>
         <version.eclipse.microprofile.metrics>2.3.1</version.eclipse.microprofile.metrics>
         <version.smallrye.metrics>2.4.1</version.smallrye.metrics>
-        
+        <version.opentracing>0.31.0</version.opentracing>
+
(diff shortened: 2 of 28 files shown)


ybroeker:
MicroProfile OpenTracing defines only behavior for jax-rs-applications and provides '@Traced' annotation, which is not usable here.

Instead io.opentracing is used directly, which is also used by smallrye-opentracing.

phillip-kruger:
Hi @ybroeker - I clean up the config in CDI module. Can you resolve this conflict ? Thanks :)

ybroeker:
> Hi @ybroeker - I clean up the config in CDI module. Can you resolve this conflict ? Thanks :)

Done :)

phillip-kruger:
Awesome ! Thanks ! Let's see if @andymc12 can have a look...

server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionDecorator.java
```diff
@@ -0,0 +1,18 @@
+package io.smallrye.graphql.execution;
+
+import graphql.ExecutionInput;
+import graphql.ExecutionResult;
+
+public interface ExecutionDecorator {
+    default void before(ExecutionInput executionInput) {
```
andymc12, line 7: Out of curiosity, why are these methods all `default`?

Also, not necessary, but maybe we could make `before` and `after` non-default and then make `onError` default and implement it by just delegating to `after`?  For Tracing the decorator definitely wants to distinguish between a normal result and an error result, but maybe in the future other decorators won't care - and just want to log entry/exit (regardless of the success of the execution).  Just a thought - it's completely fine the way it is.

ybroeker: Had the same thought that some decorators don't need everything, but to have all of them as default could indeed be too much.

Maybe it is also useful to add an additional `finally` method to have a single exit point?Normally it will call `before`, `after` (maybe renamed) and `finally`, and in case of an error `before`, `onError` and `finally`. `after` and `onError` would then be default. 

The same might also be done for `DataFetcherDecorator`.

andymc12: It's probably fine the way it is now. One nice thing about all methods being default is that if somebody were to add a new ExecutionDecorator that only needed to do something before the invocation or only in an error scenario, etc., then they only need to write one method.

If we did need to add a new decorator that always executed regardless of error or successful execution, we could add that in later - either a `finally` method or change the default implementation of `onError` to delegate to `after`.  But until we need to handle that scenario, I don't think we need to change anything with this PR.  

Nice job!

PR #251, opened 2020-05-10
Support asynchronous methods

ybroeker:
This PR adds support for asynchronous methods that return CompletionStage or CompletableFuture, such as MP Fault Tolerance `@Asynchronous`.

Originally it was only intended as a small change, but for a correct implementation with OpenTracing and Metrics more changes were necessary.

To support OpenTracing and metrics, both the end of the work in the current thread and the end of the asynchronous method must be handled.

The previous DataFetcherDecorator did not support this, and adding it would probably have made it too complex.
Therefore the whole thing was changed to a model similar to CDI interceptors. (But maybe this is too much and the old model should be kept or CDI interceptors should be used directly?)

The new model would also allow BeanValidation and Security if they were to be solved at this level.

Most of the changes are covered with unit tests (mostly MetricDecorator and OpenTracingDecorator), also some asynchronous methods are added to the tck (currently without `@Asynchronous`, just with CompletionStage completing in the same thread, but can be easily changed).

I would like to hear feedback, but if it's rejected, that's no problem :)


ybroeker:
Does anyone know why the last check fails? I can't see the details...

phillip-kruger:
@ybroeker - if you are keen you can also update the model diagram using this file (https://github.com/smallrye/smallrye-graphql/blob/master/common/schema-model/classdiagram.xml) and draw.io. And then export to png to update the README at https://github.com/smallrye/smallrye-graphql/blob/master/common/schema-model/README.adoc

server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/AsyncDataFetcher.java
```diff
@@ -0,0 +1,105 @@
...
+
+import graphql.GraphQLContext;
+import graphql.execution.DataFetcherResult;
+import graphql.schema.DataFetchingEnvironment;
+import io.smallrye.graphql.execution.datafetcher.decorator.DataFetcherDecorator;
+import io.smallrye.graphql.schema.model.Operation;
+import io.smallrye.graphql.transformation.TransformException;
+
+/**
+ * Fetch data using some bean lookup and Reflection
```
jmartisk, line 20: Could use a better description for async fetcher

jmartisk:
Great, async collections now seem to work, thanks! Now I see the formatter plugin complaining about formatting

phillip-kruger:
Thanks @jmartisk ! Thanks @ybroeker !

phillip-kruger:
@ybroeker - once you solve the formatting issue, I think we can merge .

ybroeker:
Can you wait with merging? Would just improve nonnull-handling at CompletionStage

phillip-kruger:
Cool. Let me know when you are ready

ybroeker:
It’s ready :)

phillip-kruger:
Thanks @ybroeker for this and thanks @jmartisk for your review.

phillip-kruger:
We had a request for support for smallry-mutiny:

https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/GraphQL.20-.20reactive

w.d.y.t ?

ybroeker:
Currently, one could use Smallrye Mutiny by creating a CompletionStage from the Multi:

```java
@Query("allFilms")
@Description("Get all Films from a galaxy far far away")
public CompletionStage<List<Film>> getAllFilms() {
    return service.getAllFilms().collectItems().asList().subscribeAsCompletionStage();
}
```

I don't know if this would have any disadvantages, if there are any, it would certainly make sense to support it directly.


It would probably be more interesting for graphql subscriptions.

tpenakov:
Hi @ybroeker ,
That will do fine. Thank you!
I was with "feeling" that the `Smallrye Mutiny` will become widely used than `CompletionStage` in Quarkus. That's why I've asked for it. Any non-blocking framework is OK for me.

PR #262, opened 2020-05-20
fix #258: add headers via annotations instead of CDI

### LINKED ISSUE #258: Typesafe Client: Explicit Headers
The behavior was that with CDI, all implementations of `GraphQlClientHeader` would be added as headers to all API interfaces. This is completely intransparent and unexpected.

The new behavior is that headers have to be annotated on the query method or the API interface.

This aligns it more to the MP Rest Client [ClientHeaderParam](https://download.eclipse.org/microprofile/microprofile-rest-client-2.0-m1/microprofile-rest-client-2.0-m1.html#_specifying_additional_client_headers).

t1:


client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/GraphQlClientBuilder.java
```diff
@@ -1,6 +1,7 @@
 package io.smallrye.graphql.client.typesafe.api;
 
 import java.net.URI;
+import java.net.URL;
```
andymc12, line 4: I don't see where URL is used in this class.  Maybe this was accidentally added?  Or was there an intended `endpoint(URL url)` method?

client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/Header.java
```diff
@@ -0,0 +1,28 @@
...
+ * or the name of a `method` for dynamic values. It can't be both or neither.
+ * The method is always a static method, either on the interface itself
+ * or on a fully qualified class.
+ *
+ * In rare cases, you may want to pass the value of a header as a parameter;
+ * neither `constant` nor `method` will be allowed then.
+ */
+@Retention(RUNTIME)
+@Repeatable(Headers.class)
+public @interface Header {
```
andymc12, line 22: I'd suggest adding a `@Target` annotation.  IIUC, the target should include PARAMETER, METHOD, and maybe TYPE?

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/HeaderBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private Object resolveValue(Header header) {
+        if (!header.method().isEmpty()) {
+            if (!header.constant().isEmpty())
+                throw new GraphQlClientException("Header with 'method' AND 'constant' not allowed: " + header);
+            return resolveMethodValue(header.method());
+        }
+        if (header.constant().isEmpty())
+            throw new GraphQlClientException("Header must have either 'method' XOR 'constant': " + header);
```
andymc12, line 53: Just a thought here - what do you think about using the constant as a fallback in case the method either throws an exception or returns null?

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/HeaderBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+        return header.constant();
+    }
+
+    private Object resolveMethodValue(String methodName) {
+        TypeInfo declaringType = method.getDeclaringType();
+        MethodInfo method = new MethodResolver(declaringType, methodName).resolve();
+        if (!method.isStatic())
```
andymc12, line 60: MP Rest Client does something similar with annotation-based header declaration (`@ClientHeaderParam`).  It allows users to call static methods on any accessible class or default method on the interface itself.  That might be something to consider here for consistency.

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/RequestBuilder.java
```diff
@@ -16,10 +19,13 @@ public RequestBuilder(MethodInfo method) {
...
         request.append(method.getName());
-        if (method.getParameterCount() > 0) {
+        List<ParameterInfo> parameters = method.parameters()
+                .filter(parameterInfo -> !parameterInfo.isAnnotated(Header.class))
+                .collect(toList());
+        if (parameters.size() > 0) {
             request.append("(");
             Repeated repeated = new Repeated(", ");
-            for (ParameterInfo parameterInfo : method.getParameters()) {
+            for (ParameterInfo parameterInfo : parameters) {
```
andymc12, line 28: Maybe we could replace the `Repeated` code with `StringJoiner` or `Collectors.joining()` as suggested here: https://stackoverflow.com/a/669165/6575578

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/RequestBuilder.java
```diff
@@ -16,10 +19,13 @@ public RequestBuilder(MethodInfo method) {
 
     String build() {
         request.append(method.getName());
-        if (method.getParameterCount() > 0) {
+        List<ParameterInfo> parameters = method.parameters()
+                .filter(parameterInfo -> !parameterInfo.isAnnotated(Header.class))
```
andymc12, line 23: This won't collect parameters annotated with `AuthorizationHeader`.  Is that intended?

client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/GraphQlClientBuilder.java
t1: JAX-RS uses URI; URL was a mistake. Thanks!

client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/Header.java
t1: done

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/HeaderBuilder.java
t1: Hmmm... I'd say that should be the responsibility of the method to handle, shouldn't it?

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/HeaderBuilder.java
t1: I had taken a look at the implementation, and it's quite complex. I still want to take a closer look into the whole `MethodHandle` API, so I wanted to start more restrictive and open it up later.

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/RequestBuilder.java
t1: I had that first, but it made the code more complex: appending to the StringBuilder in the different methods is easier to grasp.

client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/Header.java
t1: good point BTW!

t1: I've added it to the `AuthorizationHeader`, too (but not `PARAMETER`)

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/RequestBuilder.java
t1: Yes. Username/Password would require two fields, so the `AuthorizationHeader` annotation doesn't work on a single parameter. It also doesn't work with methods, just with config. They are quite different concepts.

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/HeaderBuilder.java
t1: I looked into MethodHandles but I'm really not sure if it's worth the hassle to call `default` methods. Users can declare their methods to be `static` methods just as well (hey, it's even one char less ;-) I'm looking for use-cases for default methods, but even when testing code accessing the API interface by mocking it: there's not need to override the method, as it won't be called.

Maybe there is a use-case I don't see?

t1: Another update: it's allowed to access private or protected methods in enclosing classes, even without qualifying it with the class name; this is a more common use-case. MP Rest Client doesn't support that, though. I'm not sure how to align them. What do you think?

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/HeaderBuilder.java
andymc12: either way works for me - if you allow the constant to be the fallback, that might simplify code where different interfaces could specify different fallbacks - maybe something like this:

```
@Header(name="ImportantHeader", method="getHeaderFromDB", constant="123")
Widget getWidget(String criteria...);

@Header(name="ImportantHeader", method="getHeaderFromDB", constant="456")
Widget getADifferentWidget(String criteria...);

static String getHeaderFromDB() {
    if (canAccessDB()) {
        return headerFromDB();
    }
    return null; // or throw exception
}
```

But we can do it either way.

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/HeaderBuilder.java
andymc12: I think that is a good approach.  We can always open it up later - but it's a lot harder to restrict it later if we find out that it doesn't add much value.

t1:
Is this okay now?

phillip-kruger:
Thanks :)

PR #272, opened 2020-06-01
fix #270: don't assume String-only constructors to define a scalar wrapper

### LINKED ISSUE #270: Bug: API Objects with a Single Scalar Field Are Interpreted As Scalars
This may be related to the use of Kotlin data classes in my code.

For the following API and model
```
@GraphQlClientApi(endpoint = "https://api.com/graphql")
interface Client {
    fun query(foo : String): Response
}

data class Project( var apiObject : ApiObject)

data class ApiObject(
    var id: String = "",
    var name: String = "" // This second field must be added or graphql client thinks that `Project` is a `String` due to its "scalar constructor"
) 

```

In this case, if `ApiObject` only has `id`, graphql request with be `apiObject`, instead of `apiObject{ id }`. Adding a second field will result in `apiObject{ id name }`

Seems caused by this chain: https://github.com/smallrye/smallrye-graphql/blob/a9f73137977d09ea325c762dc6899f1edf7d4bb8/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/GraphQlClientProxy.java#L86

https://github.com/smallrye/smallrye-graphql/blob/a9f73137977d09ea325c762dc6899f1edf7d4bb8/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/TypeInfo.java#L122

https://github.com/smallrye/smallrye-graphql/blob/a9f73137977d09ea325c762dc6899f1edf7d4bb8/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/TypeInfo.java#L136

t1:


phillip-kruger:
Thanks @t1 !

PR #279, opened 2020-06-04
Preparing gradle plugin release

marceloverdijk:
This is only a part of getting the Gradle plugin being published to the Gradle Plugin Portal.

This PR:

- Updates the Gradle version
- Updates the Gradle plugin-publish plugin version
- Added version to gradle.properties --> __this might be error prone as the version is also specified in the main pom.xml__ (Quarkus build setup has same issue I think)

What still needs to be done:

- Add the `gradle publishPlugins -Pgradle.publish.key=<key> -Pgradle.publish.secret=<secret>` step to Github release action
- Create an account on the Gradle Plugin Portal: See https://guides.gradle.org/publishing-plugins-to-gradle-plugin-portal/#create_an_account_on_the_gradle_plugin_portal
- Secure the key and secret

I've never looked into Github actions so I'm not sure I'm the best person to do this.
I think also creating the Gradle Plugin Portal account and key/secret might be better done by someone from SmallRye team itself.

### DIFF
--- a/tools/gradle-plugin/build.gradle
+++ b/tools/gradle-plugin/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'com.gradle.plugin-publish' version '0.11.0'
+    id 'com.gradle.plugin-publish' version '0.12.0'
     id 'java-gradle-plugin'
 }
 if (JavaVersion.current().isJava9Compatible()) {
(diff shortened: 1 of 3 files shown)


jmartisk:
Updating `gradle.properties` automatically sounds like something we could add after the `release:prepare` goal: 
https://github.com/smallrye/smallrye-graphql/blob/a4b43c2d0ddceb0da45b2bdf20f38cc7479ec7bd/.github/workflows/release.yml#L42

phillip-kruger:
@jmartisk sounds like a plan !

PR #280, opened 2020-06-04
fix #100: go to JUnit 5

### LINKED ISSUE #100: Update to JUnit 5

t1:
This time it should work ;-)

And it's not only junit-vintage, it's full JUnit 5 now... except for the Arquillian tests.

### DIFF
--- a/common/schema-builder/pom.xml
+++ b/common/schema-builder/pom.xml
@@ -35,9 +35,8 @@
         
         <!-- Test -->
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse</groupId>
--- a/server/implementation-cdi/pom.xml
+++ b/server/implementation-cdi/pom.xml
@@ -51,11 +51,10 @@
 
         <!-- Testing -->
         <dependency>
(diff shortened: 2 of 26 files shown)


common/schema-builder/pom.xml
```diff
@@ -35,9 +35,8 @@
         
         <!-- Test -->
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
```
phillip-kruger, line 40: Why is the scope not test anymore ?

server/implementation-cdi/pom.xml
```diff
@@ -51,11 +51,10 @@
 
         <!-- Testing -->
         <dependency>
-             <groupId>junit</groupId>
-             <artifactId>junit</artifactId>
-             <scope>test</scope>
```
phillip-kruger, line 56: Same here. Scope ?

server/implementation/pom.xml
```diff
@@ -70,9 +70,8 @@
 
         <!-- Testing -->
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
```
phillip-kruger, line 75: Same ?

common/schema-builder/pom.xml
t1: The scope is inherited from the managed dependency. I agree that it's clearer to still add the scope everywhere, but I've seen this in this project already, so I thought it would be more consistent.

Issue #69, opened 2020-03-03
Implement all "Serving over HTTP" use cases (e.g. HTTP GET, HTTP POST application/graphql)


marceloverdijk:
Looking at the source code `smallrye-graphql` currently only supports the basic `POST` use case.

As described on https://graphql.org/learn/serving-over-http/ there are more use cases:

- When receiving an HTTP GET request, the GraphQL query should be specified in the "query" query string.
- For HTTP POST requests, if the "query" query string parameter is present (as in the GET example above), it should be parsed and handled in the same way as the HTTP GET case.
- For HTTP POST requets, if the "application/graphql" Content-Type header is present, treat the HTTP POST body contents as the GraphQL query string.

E.g. in the `micronaut-graphql` implementation we support all these use cases: https://github.com/micronaut-projects/micronaut-graphql/blob/master/graphql/src/main/java/io/micronaut/configuration/graphql/GraphQLController.java

If you are interested I could help out implementing these use cases.

Let me know.

phillip-kruger:
Hi @marceloverdijk .
This is already added to master (or maybe just my fork) and will be in the next release. We Also discussed this in the last meeting (https://docs.google.com/document/d/1gb3jirFGrJwDZSbrtnFPVTNjPNe3Y0dUYfm-HkU1c3U/edit?usp=sharing) and will add more info on this in the specification and add some test cases for this. We will also allow disabling GET using config.

Thanks !

phillip-kruger:
I am keeping this open, will close one this is released

marceloverdijk:
Ah good to hear. I was on holiday last week, and looking at the code from my mobile and then it wasn't there yet :-) But indeed I see GET was added 4 days ago to `master` 👍 

I don't see the POST variants in `master` btw.

phillip-kruger:
Yea, I'll discuss the POST Options in the next spec meeting. At the moment we only expect a body in the POST. That is also why the GET is configurable, as having the info in the query might be a security concern. (Caching proxies etc) We might change the configuration from allowGet to allowQueryParam or something ?

marceloverdijk:
or both `allowGet` and `allowPostQueryParam`?
But good to hear it will be discussed.

phillip-kruger:
Cool. Good idea. Thanks :)

marceloverdijk:
I think I will add this configuration option to the `micronaut-graphql` integration as well :-)

phillip-kruger:
Cool !

[...]

PR #289, opened 2020-06-15
#69: simplistic first try to support GET queries

t1:
This is currently just a PoC for https://github.com/eclipse/microprofile-graphql/pull/266:

* It's a breaking change, as the current double-nested syntax for GET is not supported any more.
* It's not very clean, as it uses concatenation to build a JSON string, which may even be a security concern.
* It fails for 38 tck tests, and I don't understand why 😢

Please remember to set the `version.eclipse.microprofile.graphql` property to `1.1-SNAPSHOT` after building eclipse/microprofile-graphql#266


pom.xml
```diff
@@ -17,7 +17,7 @@
 
     <properties>
         <version.eclipse.microprofile.config>1.4</version.eclipse.microprofile.config>
-        <version.eclipse.microprofile.graphql>1.0.2</version.eclipse.microprofile.graphql>
+        <version.eclipse.microprofile.graphql>1.1-SNAPSHOT</version.eclipse.microprofile.graphql>
```
phillip-kruger, line 20: We can not build (in CD/CI) against snapshot

t1: I know... one of the reasons for a draft PR :-)

Just went back to 1.0.2, but please remember to go to the 1.1-SNAPSHOT from https://github.com/eclipse/microprofile-graphql/pull/266 before testing

Issue #282, opened 2020-06-07
Add ability for users to add error code and other extensions


nandorholozsnyak:
Hello there,

I'm a new person to GraphQL but in the last few days I was experimenting with it within the Quarkus framework, and I'm liking it so far. 
When it comes to writing code sometimes error can occur and I like to specify the error via error codes or with reasons which are straightforward and can be consumed easily by the caller. 
In the official GraphQL specification there are a few section about errors and one of them is specifying how the programmers could define "custom" error codes in their responses. Take a look at here:
https://spec.graphql.org/June2018/#example-fce18

So what I see there if the programmer would like to indicate any kind of custom behavior in the error response they just have to put this information into the `extension ` section in a key-value format. Easy. 
The problem with that in the SmallRye implementation this kind of stuff is not working properly (yet). 

An example:
I have a GraphQL API like the following:
```
@GraphQLApi
public class PackageResource {

    @Query("getPackages")
    @Description("Returns all available packages")
    public List<Package> getPackages() {
        throw GraphqlErrorException.newErrorException()
                .message("Something went wrong")
                .extensions(Map.of("code","CUSTOM_ERROR_CODE"))
                .build();
    }

}
```
I found the `GraphqlErrorException` class and it looks fine, because it has an `extensions` variable in it and it can be populated, so far so good, btw this class is coming from the `com.graphql-java:graphql-java:14.0` dependency right now.
So when this exception is being thrown it will be handled by the `io.smallrye.graphql.execution.error.ExceptionHandler` where it will be wrapped as a `io.smallrye.graphql.execution.error.GraphQLExceptionWhileDataFetching` exception which calls it's super class constructor `graphql.ExceptionWhileDataFetching` and it sets any kind of incoming extensions in it.
```
public ExceptionWhileDataFetching(ExecutionPath path, Throwable exception, SourceLocation sourceLocation) {
        this.path = assertNotNull(path).toList();
        this.exception = assertNotNull(exception);
        this.locations = Collections.singletonList(sourceLocation);
        this.extensions = mkExtensions(exception);
        this.message = mkMessage(path, exception);
    }
```
At this point if a newly created exception is containing extra extensions then it will be populated there. The problems comes when this exception is being serialized by the `io.smallrye.graphql.execution.error.ExecutionErrorsService#toJsonError` method.
It basically "replaces" or just simply ignores the content of the `extensions` field.
[...]

PR #295, opened 2020-06-18
#282: get error extension `code` from exception name or annotation

t1:


t1:
@andymc12 sorry to add changes to an already approved PR, but I thought that TCK tests would be a good idea to make it easier to later migrate that to MP.

client/generator-test/pom.xml
```diff
@@ -45,9 +45,8 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>io.smallrye</groupId>
-            <artifactId>smallrye-graphql-api</artifactId>
-            <version>${project.version}</version>
+            <groupId>org.eclipse.microprofile.graphql</groupId>
+            <artifactId>microprofile-graphql-api</artifactId>
```
phillip-kruger, line 49: Why did this change ?

server/api/src/main/java/io/smallrye/graphql/api/GraphQlErrorCode.java
```diff
@@ -0,0 +1,14 @@
...
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Retention(RUNTIME)
+@Target({ TYPE, ANNOTATION_TYPE })
+public @interface GraphQlErrorCode {
```
phillip-kruger, line 12: Remove GraphQl (to be consistent with other annotations) So just `@ErrorCode`

server/implementation/src/main/java/io/smallrye/graphql/execution/error/ExecutionErrorsService.java
```diff
@@ -86,12 +86,25 @@ private JsonObject toJsonError(GraphQLError error) {
...
         addKeyValue(objectBuilder, EXCEPTION, exception.getClass().getName());
         addKeyValue(objectBuilder, CLASSIFICATION, error.getErrorType().toString());
+        addKeyValue(objectBuilder, CODE, toErrorCode(exception));
         Map<String, Object> extensions = error.getExtensions();
         populateCustomExtensions(objectBuilder, extensions);
 
         return Optional.of(objectBuilder.build());
     }
 
+    private String toErrorCode(Throwable exception) {
```
phillip-kruger, line 96: Please add a TODO that this should change to use the Model and the model needs to get this using Jandex.

t1: done

server/api/src/main/java/io/smallrye/graphql/api/GraphQlErrorCode.java
t1: done

client/generator-test/pom.xml
t1: reverted

PR #302, opened 2020-06-21
Add ErrorInfo to the model so we do not need to look at annotation du…

phillip-kruger:
…ring runtime


phillip-kruger:
@t1 - F.Y.I

PR #311, opened 2020-07-01
Initial proposal of `@Context` feature

phillip-kruger:
This is facading some information from graphql-java's `DataFetchingEnviroment ` and also allow leaky abstraction to the `DataFetchingEnviroment ` via `unwrap`.

I am sure is more we need to move into the `@Context` interface, so please comment here.

I think a good test would be to see if we can create custom SQL or JPA queries using the information in `Context` (without having to use unwrap)

Please look at the code in `API` as this is what will eventually move to the spec.

Let me know what you think.


phillip-kruger:
Usage is as simple as injecting the Context anywhere you need it:

```
@Inject
Context context;
```

server/api/src/main/java/io/smallrye/graphql/api/Context.java
```diff
@@ -0,0 +1,248 @@
...
+     */
+    public JsonArray getSelectedFields();
+
+    /**
+     * This leaky abstraction allows falling down to the underlying implementation
+     * 
+     * @param <T> the implementation class
+     * @return instance of the implementation class
+     */
+    public <T> T unwrap();
```
andymc12, line 184: I think I would allow this method to pass in the Class type - so that we have more flexibility in the future to allow other types - i.e. 
```suggestion
    public <T> T unwrap(Class<T> wrappedType);
```

Then the method should probably throw an `IllegalArgumentException` for any passed-in type that we can't handle - so for now, anything other than `DataFetchingEnvironment`, but it allows us to add more context types in the future.

phillip-kruger: Sounds good. Thanks. I think this is what @t1 also wanted.

PR #347, opened 2020-07-30
Try to fix gradle

phillip-kruger:

### DIFF
--- a/tools/gradle-plugin/build.gradle
+++ b/tools/gradle-plugin/build.gradle
@@ -1,4 +1,5 @@
 plugins {
+    id'java'
     id 'com.gradle.plugin-publish' version '0.12.0'
     id 'java-gradle-plugin'
 }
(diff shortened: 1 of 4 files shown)


phillip-kruger:
@marceloverdijk - please help.

marceloverdijk:
@phillip-kruger I'm not sure this is the right direction to fix the issue to be honest.
See my comments in https://github.com/smallrye/smallrye-graphql/pull/351 which I think is the root issue of why the build started failing.

phillip-kruger:
Yea I have no idea what I am doing. Grade is not something I know at all. Good catch on the parent pom. Should we open a new issue and continue there ?

marceloverdijk:
Yes maybe good to create a separate issue and check about the parent Pom. Maybe we can reconfigure the compiler plug-in in the gradle-plugin module to not compile the sources..

PR #351, opened 2020-08-04
Make builder factory static

### LINKED ISSUE #350: Perf: 17% drop in throughput between 1.0.2 and 1.0.5 (and presumably 1.0.6)
The Open Liberty performance team found a performance issue in 1.0.5 and logged it at https://github.com/OpenLiberty/open-liberty/issues/13301

The issue is that a new JsonBuilderFactory is created each time a new instance of `SmallRyeContext` is created.  This is an expensive operation as it performs classloading, etc. Instead of creating one for each instance, we should create one static instance and re-use it.

andymc12:
This avoids the performance penalty of creating a new builder factory on
each init.

Fixes #350

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
@@ -191,5 +191,5 @@ public class SmallRyeContext implements Context {
         return field.getQualifiedName().contains("/");
     }
 
-    private final JsonBuilderFactory jsonbuilder = Json.createBuilderFactory(null);
+    private static final JsonBuilderFactory jsonbuilder = Json.createBuilderFactory(null);
 }


phillip-kruger:
Builds are failing due to gradle. I tried to fix that but are about to give up and ask @marceloverdijk to help :)

marceloverdijk:
I will try to have a look this evening.

marceloverdijk:
It seems not related to this PR/commit, but https://github.com/smallrye/smallrye-graphql/commit/0d9ff4e1e10baff2a1236dceddbb03bd4b567161

Maven tries to compile the `tools\gradle-plugin` sources which should not be done as packaging is set to `pom` for that module. I will dig deeper....

marceloverdijk:
When I change the 

```
    <parent>
        <artifactId>smallrye-parent</artifactId>
        <version>23</version>
    </parent>
```

back to 

```
    <parent>
        <artifactId>smallrye-parent</artifactId>
        <version>22</version>
    </parent>
```

on `master` branch the project compiles again.

I will have to look what changed in that parent version.

marceloverdijk:
OK, `smallrye-parent` `v23` added this via https://github.com/smallrye/smallrye-parent/commit/6acb118e87c2d882661b79a4522bbb3d595ae840

```
      <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <executions>
          <execution>
            <id>default-compile</id>
            <phase>compile</phase>
            <goals>
              <goal>compile</goal>
            </goals>
            <!-- Ensure that the logging generator does not generate invalid annotations -->
            <configuration>
              <compilerArgs>
                <arg>-Aorg.jboss.logging.tools.addGeneratedAnnotation=false</arg>
              </compilerArgs>
            </configuration>
          </execution>
        </executions>
      </plugin>
```

and I *think* this might cause the Maven compiler to compile sources in `src` even for modules with packaging `pom`.
Now in most cases modules with packaging `pom` will not have Java sources in `src/main/java` but that's not the case for the `gradle-plugin` module. What I mean to say is that this changed behaviour might not influence the build for many projects, but in this case it does.

Issue #353, opened 2020-08-05
Parent pom breaks gradle


phillip-kruger:
As @marceloverdijk figured out, the new parent pom (23) breaks the gradle module.

We need to see how we can fix that.

phillip-kruger:
From @marceloverdijk comments in the PR

OK, smallrye-parent v23 added this via smallrye/smallrye-parent@6acb118

      <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <executions>
          <execution>
            <id>default-compile</id>
            <phase>compile</phase>
            <goals>
              <goal>compile</goal>
            </goals>
            
            <configuration>
              <compilerArgs>
                <arg>-Aorg.jboss.logging.tools.addGeneratedAnnotation=false</arg>
              </compilerArgs>
            </configuration>
          </execution>
        </executions>
      </plugin>

and I think this might cause the Maven compiler to compile sources in src even for modules with packaging pom.
Now in most cases modules with packaging pom will not have Java sources in src/main/java but that's not the case for the gradle-plugin module. What I mean to say is that this changed behaviour might not influence the build for many projects, but in this case it does.

PR #354, opened 2020-08-05
Fix gradle issue

phillip-kruger:
Not really fix, more sidestep.
I don't think there is a way to fix this in the parent pom, so this "fix" just move the source folder in gradle to another place that is not the default for maven, so maven does nothing but gradle still builds.

@marceloverdijk please comment.


phillip-kruger:
Fix #353

marceloverdijk:
Yes that could work as workaround.

I have no access to the code at the moment to try, but maybe this is an option as well by adding the maven compiler plugin to the `/tools/gradle-plugin/pom.xml` like:

```
<build>
   <plugins>
       <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-compiler-plugin</artifactId>
            <version>3.8.1</version>
            <configuration>
                <excludes>
                    <exclude>**/io/smallrye/*</exclude>
                </excludes>
             </configuration>
        </plugin>
    </plugins>
</build>
```

phillip-kruger:
Thanks @marceloverdijk - I think it's much or a muchness :) I am merging this PR, we can also change later if we believe above approach is better

PR #365, opened 2020-08-14
Batch Source 

phillip-kruger:
This PR contains the initial changes to support batch source methods.
So this is now valid:
```
    @Name("timestamp")
    public List<TestSource> getTestSources(@Source List<TestObject> testObjects) {
         // Here do batch operation
         return batched;
    }
```
This PR also contains a refactor on the decorators used to add support for Metrics, OpenTracing and Bean validation.
It's changed to use an event/visitor that decouple the integration with the implementation, where the impl just fire a notification that can be used by the integration.

This also means that all integration is neatly in one place and that users can now also take part in this by using the CDI Events.


jmartisk:
Quick look at a +2,333/−2,497 line PR :smiley:  yeah sure I'll try to give this a spin today

phillip-kruger:
Thanks @jmartisk !

jmartisk:
This breaks the WildFly subsystem, trying to find out why.

phillip-kruger:
That's not good. Try looking at the SPI stuff, as that has changed

server/implementation/src/main/java/io/smallrye/graphql/execution/event/EventEmitter.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    private static final ServiceLoader<EventingService> eventingServices = ServiceLoader.load(EventingService.class);
+
+    private final List<EventingService> enabledServices = new ArrayList<>();
+
+    private EventEmitter(Config config) {
+        Iterator<EventingService> it = eventingServices.iterator();
+        while (it.hasNext()) {
+            EventingService eventingService = it.next();
```
jmartisk, line 124: This is bad, because when you call `next()` it creates an instance of the service, so this will force the creation of all services that are listed in `META-INF/services/io.smallrye.graphql.spi.EventingService`, even if that service is disabled (for example, it creates an instance of `TracingService` when `smallrye.graphql.tracing.enabled=false`, which causes the whole deployment to break if tracing library is not on the classpath).

We could wrap this in a try-catch block and ignore services that can't be constructed - that is quite stupid, but I don't see a better way right now, because you can't call `getConfigKey()` without building an instance of the service :disappointed:

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Config.java
```diff
@@ -54,10 +50,22 @@ default boolean isTracingEnabled() {
...
+
     default boolean isValidationEnabled() {
         return false;
     }
 
+    default boolean isEventsEnabled() {
+        return false;
+    }
+
+    default boolean shouldEmmitEvents() {
```
jmartisk, line 65: emmit -> emit (not only  here, the same typo is present in more places)

jmartisk (changes requested):

server/implementation/src/main/java/io/smallrye/graphql/execution/event/EventEmitter.java
```diff
@@ -0,0 +1,196 @@
...
+    private static Optional<EventEmitter> emitter() {
+        EventEmitter eventEmitter = current.get();
+        if (eventEmitter != null) {
+            return Optional.of(eventEmitter);
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private static final ServiceLoader<EventingService> eventingServices = ServiceLoader.load(EventingService.class);
```
jmartisk, line 117: This must not be static. In WildFly, the `EventEmitter` class will be loaded by the same class loader for all deployments, therefore the value of static field in it would be shared between all deployments, which is wrong.

In general, to work with WildFly, we need to avoid static fields as much as possible.

jmartisk (changes requested):

jmartisk:
WildFly integration looks good now, I'm looking at the batch source functionality. Not sure if I'm using everything correctly, but it seems when you have a batch-source method that takes additional arguments, it breaks:
```
@Name("secretToken")
    public List<String> generateSecretToken(@Source List<Person> persons,
                                      @DefaultValue("true")
                                      @Name("maskFirstPart") boolean maskFirstPart) {
```

With this a query fails with 
```
2020-08-24 14:24:36,368 ERROR [io.sma.graphql] (ForkJoinPool.commonPool-worker-11) SRGQL012000: Data Fetching Error: java.util.concurrent.CompletionException: java.lang.RuntimeException: java.lang.IllegalArgumentException: wrong number of arguments
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1702)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1692)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
Caused by: java.lang.RuntimeException: java.lang.IllegalArgumentException: wrong number of arguments
	at io.smallrye.graphql.bootstrap.Bootstrap$2.doSourceListCall(Bootstrap.java:660)
	at io.smallrye.graphql.bootstrap.Bootstrap$2.lambda$load$0(Bootstrap.java:651)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1700)
	... 6 more
Caused by: java.lang.IllegalArgumentException: wrong number of arguments
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at io.smallrye.graphql.execution.datafetcher.helper.ReflectionHelper.invoke(ReflectionHelper.java:43)
	at io.smallrye.graphql.bootstrap.Bootstrap$2.doSourceListCall(Bootstrap.java:657)
	... 8 more
```

phillip-kruger:
That seems like a bug. Can we merge this, then I'll open an issue for that and fix in another PR, as this PR is becoming very big. I'll fix before we release.

jmartisk (approved):
agreed ;)

phillip-kruger:
Ok, merging this, created https://github.com/smallrye/smallrye-graphql/issues/369 for the extra parameter issue. Thanks @jmartisk !!

PR #381, opened 2020-08-27
Fix NPE (from Quarkus)

phillip-kruger:


phillip-kruger:
Hi @jmartisk  - Got this while trying to upgrade Quarkus to use 1.0.8. - will have to do a new release :(

jmartisk (approved):
I didn't check that it indeed fixes Quarkus integration, but I don't see any obvious issue here, so +1

PR #382, opened 2020-08-27
Check if we should transform before we transform.

phillip-kruger:
@andymc12 - I am not convinced that this is the cause of the slower speed between versions, but this code can do no harm. Let's do a perf test again against 1.0.9


jmartisk:
I realize that the bad news is that after we merged https://github.com/smallrye/smallrye-graphql/pull/383, wildfly tests will always fail until we cut a graphql release and update wildfly with it :/ 
Or we could make the feature pack depend on `1.0.9-SNAPSHOT` and apply the metrics changes in it right away, but I don't like depending on snapshots. So I guess we'll need to do a release soon.

phillip-kruger:
No worries. I want to release soon, just trying to get this PR in.

Issue #281, opened 2020-06-05
Are Generics not supported?


marceloverdijk:
I'm having:

```
public interface PageType<T> {
    List<T> getItems();
    boolean isHasNextPage();
    boolean isHasPreviousPage();
    long getTotalCount();
}

@Type("ContinentPage")
public class ContinentPageType implements PageType<ContinentType> {

    private final List<ContinentType> items;
    private final boolean hasNextPage;
    private final boolean hasPreviousPage;
    private final long totalCount;

    // ctor, getters, setters omitted to keep issue text shorter
}

@GraphQLApi
public class MyGraphQLApi {

    @Query("continents")
    @Description("Get a page of continents")
    public ContinentPageType getContinentPage() {
        return new ContinentPageType(..);
    }
}
```

But this gives: 

```
Caused by: io.smallrye.graphql.schema.SchemaBuilderException: Don't know what to do with [T] of kind [TYPE_VARIABLE]
        at io.smallrye.graphql.schema.creator.ReferenceCreator.getReference(ReferenceCreator.java:214)
        at io.smallrye.graphql.schema.creator.ReferenceCreator.getReference(ReferenceCreator.java:202)
        at io.smallrye.graphql.schema.creator.ReferenceCreator.createReferenceForInterfaceField(ReferenceCreator.java:118)
        at io.smallrye.graphql.schema.creator.FieldCreator.createFieldForInterface(FieldCreator.java:55)
        at io.smallrye.graphql.schema.creator.type.InterfaceCreator.addFields(InterfaceCreator.java:78)
        at io.smallrye.graphql.schema.creator.type.InterfaceCreator.create(InterfaceCreator.java:57)
        at io.smallrye.graphql.schema.creator.type.InterfaceCreator.create(InterfaceCreator.java:31)
        at io.smallrye.graphql.schema.SchemaBuilder.createAndAddToSchema(SchemaBuilder.java:148)
        at io.smallrye.graphql.schema.SchemaBuilder.addTypesToSchema(SchemaBuilder.java:107)
        at io.smallrye.graphql.schema.SchemaBuilder.generateSchema(SchemaBuilder.java:88)
        at io.smallrye.graphql.schema.SchemaBuilder.build(SchemaBuilder.java:60)
```

Note that in the `@GraphQLApi` I don't directly refer to the `PageType<T>` interface.

It would be nice to have some sort of Page contract in my Java classes.

phillip-kruger:
[...]

PR #385, opened 2020-08-31
#281 - support for generics

velias:
Support for generics as proposed in issue #281. 
Added mainly to schema generation process. Unit tests for schema generation added also.

Runtime tested using example application as I do not see any automated tests for this. 
Everything looks OK, I only see this log message when generics is used as input parameter in mutation `WARNING [org.eclipse.yasson.internal.ReflectionUtils] (default task-1) Generic bound not found for type T declared in class com.github.phillipkruger.user.graphql.ClassWithOneGenericsParam.`, but it looks to work well.


phillip-kruger:
@velias - This is awesome !!!
Thanks very much !!

phillip-kruger (approved):
Thanks !

PR #397, opened 2020-09-03
Support async batching

### LINKED ISSUE #396: Async Batch Source not working
I have a `Person` with an a nullable attribute `client`, which is a `Client`, which I'm batching. This works:

    public List<Optional<Client>> client(@Source List<Person> people)

and it gives this schema fragment:

    type Person {
       client: Client
    }

Making this async, breaks it though. The following:

    public CompletableFuture<List<Optional<Client>>> client(@Source List<Person> people)

gives the following schema fragment:

      type Person {
          client: [Client]
      }

Now, client is unexpectedly a list. It also breaks at runtime. It appears that `ArrayCreator.createArray` incorrectly marks the Operation as an array type.

eamelink:
Fixes #396

Updates the Schema Builder to work correctly with the combination of batching and async @Source methods.

Updates the Source Batch Loader to differentiate between normal methods that are invoked with `supplyAsync`, and async methods that are invoked as-is.

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/ArrayCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/ArrayCreator.java
@@ -47,18 +47,23 @@ public class ArrayCreator {
     }
 
     public static Optional<Array> createArray(Type fieldType, Type methodType, boolean batched) {
-        if (Classes.isCollectionOrArray(methodType) && !batched) {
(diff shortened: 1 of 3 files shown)


phillip-kruger (approved):
Very nice ! Thanks @eamelink !!

phillip-kruger:
@Thanks @eamelink ! How 'urgent' is this for you ?

eamelink:
Not urgent, I can continue running a snapshot.

PR #403, opened 2020-09-11
Avoid initializing multiple EventEmitter instances

andymc12:
While debugging an issue with Metrics and Open Liberty, I noticed that the EventEmitter class is initialized four times using the same Config object.  Each time it does that it uses the ServiceLoader to find classes and attempt to register them.  This can get expensive, and we really only need one instance of the EventEmitter class.  This change makes one per Config (assuming that there could be some case where there will be multiple Configs - each with different class loaders allowing a different set of EmitterServices).

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/event/EventEmitter.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/event/EventEmitter.java
@@ -1,9 +1,13 @@
 package io.smallrye.graphql.execution.event;
 
 import java.util.ArrayList;
+import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.ServiceLoader;
+
+import org.jboss.logging.Logger;
 
 import graphql.schema.GraphQLSchema;
 import io.smallrye.graphql.api.Context;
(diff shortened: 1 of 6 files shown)


phillip-kruger:
Thanks for this @andymc12 - It looks good to me, and seems like the WF build are passing, so it should be ok. I just want @jmartisk to also have a look at static variables has been problematic in the past in cases where multiple apps run (like WF)

phillip-kruger (dismissed):
It's per config so should be ok in multiple apps. ? @jmartisk  ?

server/implementation/src/main/java/io/smallrye/graphql/execution/event/EventEmitter.java
```diff
@@ -19,22 +23,31 @@
...
+                eventingService = it.next();
                 String configKey = eventingService.getConfigKey();
                 boolean enabled = config.getConfigValue(configKey, boolean.class, false);
                 if (enabled) {
                     enabledServices.add(eventingService);
                 }
             } catch (Throwable t) {
                 // Ignore that service...
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("Failed to register eventing service, " + eventingService, t);
```
jmartisk, line 49: Shouldn't this rather be `WARN`? This looks like quite a serious problem (metrics and tracing won't work, for example)

jmartisk (approved):
I think from WildFly side it should be good. Added just one (not very important) suggestion about logging, otherwise LGTM

jmartisk:
I just realized we will need to implement some proper cleanup; in the current state, there will be a class loader leak when an application is undeployed (we don't ever clear its reference in the `Map<Config, EventEmitter> INSTANCES`) 

but as there are most definitely more similar issues in other places and we don't have a concise strategy how to perform cleanup, I submitted https://github.com/smallrye/smallrye-graphql/issues/405 and plan to look into that from a more general PoV at some point

Issue #418, opened 2020-09-24
Some GraphQL types for generics in 'type' section of the scheme have incorrect names


velias:
Follow up on #281, GraphQL type in the `type` section has sometimes name like `ClassWithOneGenericsParam_Int` instead of `ClassWithOneGenericsParam_Integer` - cause looks to be that appended name is GraphQL type name instead of Java type name. Type in `query` is correct in this case, so whole schema is invalid as a result.

Probably affects `interface` section too.

velias:
@phillip-kruger I found this while patching quarkus deployment. I'll improve tests and patch this problem.

velias:
Looks like the problem is related to translation of Java classes to GraphQL SCALARs. One part of the code always uses name from the java class, second one uses GraphQL name which is different due to this translation

phillip-kruger:
Cool ! Are you preparing a PR ?

velias:
Sure

phillip-kruger:
@velias - I want to do a release to fix the ParametrizedTypeArguments issue. Would be good to get this in as well. When are you planning a PR ?

velias:
I plan to do PR of this now. I found one more error in the schema generation for one specific generics structure (when field in class is generic class which takes generic type resolution from the class owning this field). I'll create error report for it, but PR for it probably sometimes during next week as I'm not sure how complicated it will be to resolve it. 
So feel free to release once I provide PR for this.

phillip-kruger:
Thanks !

velias:
PR created - #422

PR #422, opened 2020-09-25
#418 - patched type names generation for some java generics cases

velias:
full test of schema types consistency added


phillip-kruger (approved):
Thanks @velias !!

PR #430, opened 2020-09-29
#429 - patched Bootstrap.generateGraphQLSchema when generic type with

velias:
two distinct resolutions exists in the schema. Test added to cover this problem.

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
@@ -237,7 +237,7 @@ public class Bootstrap {
         GraphQLInterfaceType graphQLInterfaceType = interfaceTypeBuilder.build();
         // To resolve the concrete class
         this.codeRegistryBuilder.typeResolver(graphQLInterfaceType, new InterfaceResolver(interfaceType));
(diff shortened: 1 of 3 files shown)


server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -237,7 +237,7 @@ private void createGraphQLInterfaceType(InterfaceType interfaceType) {
         GraphQLInterfaceType graphQLInterfaceType = interfaceTypeBuilder.build();
         // To resolve the concrete class
         this.codeRegistryBuilder.typeResolver(graphQLInterfaceType, new InterfaceResolver(interfaceType));
-        this.interfaceMap.put(interfaceType.getClassName(), graphQLInterfaceType);
+        this.interfaceMap.put(interfaceType.getName(), graphQLInterfaceType);
```
phillip-kruger, line 240: Why did this change from classname to name ? (Just out of interest, seems like everything is still working...)

velias: Even interfaces can be generic, so graphql name is unique, but class name may not be. So better to use name here

PR #431, opened 2020-09-30
Add support for Grouping

### LINKED ISSUE #426: Allow using @Named on @GraphQLApi resource
Hi,

following discussion with @phillip-kruger in Quarkus issue https://github.com/quarkusio/quarkus/issues/12302, we agreed to create a seperate issue here. 

Basically, we are concerned with how the maintainability of the code scales as the number of graphql endpoints grow. It would be nice to have an equivalent to @Path in JAX-RS, but that is a change to the microprofile spec, which I've also created an issue on here: 
https://github.com/eclipse/microprofile-graphql/issues/320

If the spec cannot be changed or the issue is rejected, then we would like to test out the following addition to the @GraphQLApi annotation:

`@GraphQLApi @Name("Books")`

phillip-kruger:
Fix #426

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
@@ -2,6 +2,7 @@ package io.smallrye.graphql.schema;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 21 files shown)


mejlholm (approved):

phillip-kruger:
I am leaving it open for a day or so. Hoping to get some input from @kaqqao

Issue #448, opened 2020-10-07
Graphql client throws exception when errors array is empty.


erikhh:
I'm using smallrye-graphql-client-all version 1.0.12.

The API I'm working with returns results like this:
```
{
  "errors": [],
  "data": {
    "field": {
      "foo": "some value"
    }
  },
  "extensions": null,
  "dataPresent": true
}
```

And even tough there are no errors in the error list, the client throws a `GraphQlClientException`, with the message `errors from service: []`

Apart from this issue. It would be useful to have a more fine grained error handling mechanism than a 'all or nothing exception'. So you can deal with situations where some fields of the request give errors but others still return results.

phillip-kruger:
@t1 - should be an easy fix. I think we can assume no errors if the error field is not there OR empty.

erikhh:
Proposed fix got accepted.

PR #449, opened 2020-10-07
Assume no error if the error fiels is not there or empty

erikhh:
A fix for #448


phillip-kruger (approved):
LGTM ! Thanks @erikhh (for solving your own problem :))

PR #452, opened 2020-10-08
Array out of bounds

erikhh:
Test case and fix for an `ArrayIndexOutOfBoundsException` situation I ran into. 
Not all that sure about what I'm doing here, please check. But it does sovle my problem.


phillip-kruger:
@t1 - please review. Thanks :)

client/implementation/src/test/java/test/unit/TypeInfoBehaviour.java
```diff
@@ -0,0 +1,27 @@
...
+
+import static org.assertj.core.api.BDDAssertions.then;
+
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import io.smallrye.graphql.client.typesafe.impl.reflection.TypeInfo;
+
+class TypeInfoBehaviour {
```
t1, line 11: This has to be the American spelling `Behavior`, or it won't be picked up by the config in the `maven-surefire-plugin`.

t1 (commented):
The `TypeInfo` class is not part of the public API. Can you elaborate a bit on the actual use-case you ran into problems with? I'd prefer to have tests on the use-case level, not on implementation details.

erikhh:
I've replaced my original test case for one in `AnnotationBehavior` that reproduces the issue from the client API side. You're right, it is an improvement to do it like this.

t1 (approved):
super, now I understand your use-case :-)

PR #485, opened 2020-10-22
consider `@Input` annotations

t1:


client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/ParameterInfo.java
```diff
@@ -45,6 +47,10 @@ private String graphQlInputTypeName(TypeInfo type) {
     }
 
     private String simpleInputTypeName(TypeInfo type) {
+        if (type.isAnnotated(Input.class))
+            return type.getAnnotation(Input.class).value();
+        if (type.isAnnotated(Type.class))
```
phillip-kruger, line 52: Input should not use `@Type`. First look `@Input`, then `@Name` then `className`. On the server we also have a NamingStrategy when using ClassName. Default is just the ClassName, then you also MergeInnerClass (so Inner class will do parentClassNameInnerClassName, Or you can do Full then it use the full packageName. Naming Strategy can be set via config.

phillip-kruger (changes requested):

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/ParameterInfo.java
t1: You're absolutely right. I just fixed that.

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/ParameterInfo.java
```diff
@@ -45,6 +46,8 @@ private String graphQlInputTypeName(TypeInfo type) {
     }
 
     private String simpleInputTypeName(TypeInfo type) {
+        if (type.isAnnotated(Input.class))
+            return type.getAnnotation(Input.class).value();
         if (type.isAnnotated(Name.class))
             return type.getAnnotation(Name.class).value() + "Input";
```
phillip-kruger, line 52: Only append 'Input' on classname. same as server

phillip-kruger (approved):

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/ParameterInfo.java
t1: Okay... I'm not really convinced that this is the best way to go, but consistency with the server beats any argument ;-)

phillip-kruger (approved):
Thanks @t1!

PR #487, opened 2020-10-22
[486] Ensure we process JARs inside WARs

### LINKED ISSUE #486: Servlet module does not scan WEB-INF/lib JARs for MP GraphQL components
If annotated POJOs are located in a WAR's WEB-INF/classes directory, then everything works fine. But when running in Open Liberty with a WAR file that contains query/mutation/entity classes, etc in a JAR file in the WEB-INF/lib directory, those classes are not processed.  Since these JARs are part of the classpath of the web application, they should also be scanned for MP GraphQL components.

andymc12:
Fixes #486

### DIFF
--- a/server/implementation-servlet/src/main/java/io/smallrye/graphql/servlet/IndexInitializer.java
+++ b/server/implementation-servlet/src/main/java/io/smallrye/graphql/servlet/IndexInitializer.java
@@ -133,7 +133,7 @@ public class IndexInitializer {
         if (fileName.endsWith(DOT_CLASS)) {
             SmallRyeGraphQLServletLogging.log.processingFile(fileName);
             indexer.index(is);
-        } else if (fileName.endsWith(DOT_WAR)) {
+        } else if (fileName.endsWith(DOT_WAR) || fileName.endsWith(DOT_JAR)) {
             // necessary because of the thorntail arquillian adapter
             processJar(is, indexer);
         }


t1 (commented):
The change looks good to me, but aren't there any tests for this? And I assume we have the same problem with `.ear` files, which are less common, but still in use sometimes.

server/implementation-servlet/src/main/java/io/smallrye/graphql/servlet/IndexInitializer.java
```diff
@@ -133,7 +133,7 @@ private void processFile(String fileName, InputStream is, Indexer indexer) throw
         if (fileName.endsWith(DOT_CLASS)) {
             SmallRyeGraphQLServletLogging.log.processingFile(fileName);
             indexer.index(is);
-        } else if (fileName.endsWith(DOT_WAR)) {
+        } else if (fileName.endsWith(DOT_WAR) || fileName.endsWith(DOT_JAR)) {
             // necessary because of the thorntail arquillian adapter
```
t1, line 137: this comment doesn't make a lot of sense any more, does it?

jmartisk:
Just FYI I'm trying this with the WildFly feature pack - created a deployment where the GraphQL API class is in a WAR, and model classes are in a `WEB-INF/lib/` jar and it works, even without this fix. Perhaps WF does something differently that it works already. I'll add a test there to be sure it doesn't break.

phillip-kruger (approved):

PR #495, opened 2020-10-29
Initial refactor to support more Async types.

phillip-kruger:
This initial refactor moves the datafetching to SPI, allowing us to "plug-in" other datafetchers based on the operation. At the moment the default will be synchronous reflection calls. And build in is a datafetcher provider that can handle CompletableFuture reflection calls.

Next step is to plug in Mutiny as a provider. 

There is still some refactoring needed in the model and model builder where we detect if this is async (based on name=CompletableFuture). We either need to make that plugable as well or, ideally, remove that logic and do that in the core module. That will be some refactoring on how we do generics as the model figure that out by checking not collection and not async. I think that logic should move to the core module as well. And the model only knows about the structure.

I decided to do a PR now, before I start with the above, else the PR will become very big.


server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/PlugableDataFetcher.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public PlugableDataFetcher(Operation operation, Config config) {
+        this.operation = operation;
+        this.dataFetcherService = DataFetcherService.getDataFetcherService(operation);
+        this.dataFetcherService.init(operation, config);
+    }
+
+    @Override
+    public T get(final DataFetchingEnvironment dfe) throws Exception {
+        SmallRyeContext.setDataFromFetcher(dfe, operation);
```
cescoffier, line 29: Does that use some kind of thread local?

cescoffier: or is it just a static singleton.

server/implementation/src/main/java/io/smallrye/graphql/spi/DataFetcherService.java
```diff
@@ -0,0 +1,147 @@
...
+    String getName();
+
+    String[] forClasses();
+
+    void init(final Operation operation, Config config);
+
+    <T> T get(final DataFetchingEnvironment dfe,
+            final DataFetcherResult.Builder<Object> resultBuilder) throws Exception;
+
+    <T> CompletionStage<List<T>> batch(final BatchLoaderEnvironment ble, final List<Object> keys);
```
cescoffier, line 74: Do you want to keep a CS here?

server/implementation/src/main/java/io/smallrye/graphql/spi/DataFetcherService.java
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            return (T) resultBuilder.build();
+        }
+
+        @Override
+        public <T> CompletionStage<List<T>> batch(BatchLoaderEnvironment ble, List<Object> keys) {
+            Object[] arguments = batchLoaderHelper.getArguments(keys, ble);
+            final ClassLoader tccl = Thread.currentThread().getContextClassLoader();
+            return CompletableFuture.supplyAsync(() -> (List<T>) reflectionHelper.invokePrivileged(tccl, arguments));
```
cescoffier, line 144: Don't use supplyAsync without an executor. It picks a thread from the fork/join pool, you don't have any control.

server/implementation/src/main/java/io/smallrye/graphql/spi/datafetcher/CompletableFutureDataFetcherService.java
```diff
@@ -0,0 +1,126 @@
...
+            eventEmitter.fireOnDataFetchError(dfe.getExecutionId().toString(), graphQLException);
+        } catch (SecurityException | IllegalAccessException | IllegalArgumentException ex) {
+            //m.invoke failed
+            eventEmitter.fireOnDataFetchError(dfe.getExecutionId().toString(), ex);
+            throw ex;
+        } finally {
+            eventEmitter.fireAfterDataFetch();
+        }
+
+        return (T) CompletableFuture.completedFuture(resultBuilder.build());
```
cescoffier, line 116: Is this really what you want? It seems that something wrong happened and you still returned a future. Shouldn't you send the failure (in a future) ?

server/implementation/src/test/java/io/smallrye/graphql/execution/AsyncTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Test a basic async endpoint
+ * 
+ * @author Phillip Kruger (phillip.kruger@redhat.com)
+ */
+public class AsyncTest extends ExecutionTestBase {
+
+    protected IndexView getIndex() {
+        return Indexer.getTestIndex("io/smallrye/graphql/test/async");
```
cescoffier, line 19: don't you want to rename it to be more specific (CompletationStage?)

cescoffier (changes requested):
I don't know much about the codebase, so I focused on the _async_ part only.

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/GenericsInfo.java
```diff
@@ -0,0 +1,28 @@
+package io.smallrye.graphql.schema.model;
+
+import java.io.Serializable;
+
+/**
+ * If the type is wraped in a generics bucket, keep the info here.
+ * 
+ * @author Phillip Kruger (phillip.kruger@redhat.com)
+ */
+public class GenericsInfo implements Serializable {
```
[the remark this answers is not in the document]
phillip-kruger: All classes in the model is Serializable, we record that in Quarkus build.

server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/PlugableDataFetcher.java
phillip-kruger: Yes local thread.

server/implementation/src/main/java/io/smallrye/graphql/spi/DataFetcherService.java
phillip-kruger: The underlying graphql-java wants a CS. We use that to to the batch.

server/implementation/src/test/java/io/smallrye/graphql/execution/AsyncTest.java
phillip-kruger: Yes :)

server/implementation/src/main/java/io/smallrye/graphql/spi/datafetcher/CompletableFutureDataFetcherService.java
phillip-kruger: graphql can do partial responses, so the errors are part of a valid response.

server/implementation/src/main/java/io/smallrye/graphql/spi/DataFetcherService.java
andymc12: That sounds like we might need some sort of config mechanism for users to be able to specify the executor service they want to use.

server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/helper/ReflectionHelper.java
```diff
@@ -38,6 +41,39 @@ public ReflectionHelper(Operation operation, EventEmitter eventEmitter) {
...
+                        public Object run() {
+                            ClassLoader originalTccl = Thread.currentThread()
+                                    .getContextClassLoader();
+                            Thread.currentThread().setContextClassLoader(classLoader);
+
+                            try {
+                                return invoke(arguments);
+                            } catch (Exception ex) {
+                                // TODO: Handle this better.
+                                throw new RuntimeException(ex);
```
andymc12, line 64: You may be able to avoid the catch block entirely and have the run method declare that it throws Exception.  Any exception thrown from here should be wrapped in the PrivilegedActionException.

server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/helper/ReflectionHelper.java
```diff
@@ -38,6 +41,39 @@ public ReflectionHelper(Operation operation, EventEmitter eventEmitter) {
...
+                            } finally {
+                                if (originalTccl != null) {
+                                    Thread.currentThread().setContextClassLoader(classLoader);
+                                }
+                            }
+                        }
+                    });
+        } catch (PrivilegedActionException e) {
+            throw new RuntimeException(e);
```
andymc12, line 73:
```suggestion
            throw new RuntimeException(e.getCause());
```

In my experience the PrivilegedActionException is mostly useless for problem solving and it adds noise to the stacktrace/logs, but is necessary in order to wrap the real exception.  So, I'd suggest just using the original exception.

server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/helper/ReflectionHelper.java
```diff
@@ -38,6 +41,39 @@ public ReflectionHelper(Operation operation, EventEmitter eventEmitter) {
...
+
+                            try {
+                                return invoke(arguments);
+                            } catch (Exception ex) {
+                                // TODO: Handle this better.
+                                throw new RuntimeException(ex);
+                            } finally {
+                                if (originalTccl != null) {
+                                    Thread.currentThread().setContextClassLoader(classLoader);
+                                }
```
andymc12, line 68:
```suggestion
                                Thread.currentThread().setContextClassLoader(classLoader);
```

I would suggest always setting it back to the original loader - even if it was null.  A null classloader (as weird as it sounds) actually has a meaning - it is the JDK's bootstrap classloader.  I can't imagine a case where the thread's context classloader would be null before coming here, but it's always better to set it back to what it was before.

andymc12 (changes requested):
I think the biggest thing is the ServiceLoader wrt environments with multiple applications (Wildfly/Liberty/etc.), but mostly this change looks good to me.

server/implementation/src/main/java/io/smallrye/graphql/spi/DataFetcherService.java
phillip-kruger: I think we need to open a separate issue for this. (the current code is exactly the same as above.) 
I opened https://github.com/smallrye/smallrye-graphql/issues/496 for this, we need to change this as part of another PR.

server/implementation/src/main/java/io/smallrye/graphql/spi/DataFetcherService.java
```diff
@@ -0,0 +1,147 @@
...
+ * Plugable way to do data fething.
+ * We will provide some built in providers.
+ * 
+ * @author Phillip Kruger (phillip.kruger@redhat.com)
+ */
+public interface DataFetcherService {
+
+    ServiceLoader<DataFetcherService> dataFetcherServices = ServiceLoader.load(DataFetcherService.class);
+
+    Map<String, DataFetcherService> dataFetcherServicesMap = load();
```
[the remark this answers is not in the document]
phillip-kruger: Hi @andymc12 - Would the same not apply to ClassloadingService ?

server/implementation/src/main/java/io/smallrye/graphql/spi/datafetcher/CompletableFutureDataFetcherService.java
```diff
@@ -0,0 +1,126 @@
...
+        this.fieldHelper = new FieldHelper(operation);
+        this.reflectionHelper = new ReflectionHelper(operation, eventEmitter);
+        this.argumentHelper = new ArgumentHelper(operation.getArguments());
+        this.partialResultHelper = new PartialResultHelper();
+        this.batchLoaderHelper = new BatchLoaderHelper();
+    }
+
+    @Override
+    public <T> T get(DataFetchingEnvironment dfe, DataFetcherResult.Builder<Object> resultBuilder) throws Exception {
+        eventEmitter.fireBeforeDataFetch();
```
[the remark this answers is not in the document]
phillip-kruger: With above, I refactored to have an Abstract DataFetcher that should solve both concerns.

server/implementation/src/main/resources/META-INF/services/io.smallrye.graphql.spi.DataFetcherService
```diff
@@ -0,0 +1 @@
+io.smallrye.graphql.spi.datafetcher.CompletableFutureDataFetcherService
```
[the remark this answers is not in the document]
phillip-kruger: You mean if they want to override how to handle CompletableFuture and not use the out of the box version ?

server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/helper/ReflectionHelper.java
phillip-kruger: Cool, done !

server/implementation/src/main/java/io/smallrye/graphql/spi/DataFetcherService.java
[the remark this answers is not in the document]
phillip-kruger: Also, the fields (dataFetcherServices and dataFetcherServicesMap) are not static ?

server/implementation/src/main/resources/META-INF/services/io.smallrye.graphql.spi.DataFetcherService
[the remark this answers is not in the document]
phillip-kruger: Ok, I now load the Built in onces manually, if it's not already loaded by SPI.

server/implementation/src/main/java/io/smallrye/graphql/spi/DataFetcherService.java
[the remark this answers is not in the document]
phillip-kruger: I created https://github.com/smallrye/smallrye-graphql/issues/497 for this. Merging as is, and after we confirm the issue, we can do a PR. Hope that is ok with you.

phillip-kruger:
@cescoffier @andymc12 thanks for the review, I resolved most things, and create 2 new issues to look at separate from this. Merging this.fixed

Issue #493, opened 2020-10-27
Support any map type.


patope:
Test case below 

```
class ReferenceCreatorTest {

    public static class MapImpl extends AbstractMap<String,String> implements Map<String,String> {
        @Override
        public Set<Entry<String, String>> entrySet() {
            return null;
        }
    }

    @Test
    public void test() throws Exception {
        try {
            ReferenceCreator referenceCreator = new ReferenceCreator(TypeAutoNameStrategy.Full);
            // Modified IndexCreator to accept variable number of arguments
            Index index = IndexCreator.index(Map.class, MapImpl.class);
            ScanningContext.register(index);
            Reference reference = referenceCreator.createReference(Direction.OUT, index.getClassByName(DotName.createSimple(Map.class.getName())));
            // reference should be?
        } finally {
            ScanningContext.remove();
        }
    }
}
```

throws exception
```
java.lang.IllegalArgumentException: Not a type variable!

	at org.jboss.jandex.Type.asTypeVariable(Type.java:241)
	at io.smallrye.graphql.schema.creator.ReferenceCreator.createReference(ReferenceCreator.java:203)
	at io.smallrye.graphql.schema.creator.ReferenceCreator.createReference(ReferenceCreator.java:161)
	at io.smallrye.graphql.schema.creator.ReferenceCreatorTest.test(ReferenceCreatorTest.java:31)
       ...
```

I have panache jpa model that triggers this issue. `io.smallrye.config.ConfigSourceMap` is direct implementer of Map.

phillip-kruger:
Thanks for this. We will need to see how we can accommodate this.

PR #502, opened 2020-11-03
Ignore non type parameters of specialized implementation of generic type

patope:
Proposed fix for https://github.com/smallrye/smallrye-graphql/issues/493

This issue is not specific to Map, but to all generic interface types that have specialized implementation.


phillip-kruger:
Hi @patope - Thanks for this. I will look at this a.s.a.p. I might break this with some refactoring I am busy with, let's wait and see

phillip-kruger (approved):

Issue #316, opened 2020-07-11
Add Stereotypes and Mixins (and more)


t1:
I added [power-annotations](https://github.com/t1/power-annotations) to the typesafe client as a PoC.

There are two tests that show using a [Stereotype](https://github.com/t1/smallrye-graphql/blob/power-annotations/client/implementation/src/test/java/test/unit/StereotypeBehavior.java) and a [Mixin](https://github.com/t1/smallrye-graphql/blob/power-annotations/client/implementation/src/test/java/test/unit/MixinBehavior.java).

Maybe you can take a look?

It's in a [branch](https://github.com/t1/smallrye-graphql/tree/power-annotations) in my fork. I didn't want to create a PR to master, yet, as I think we should work on a branch first; and it seems like I can't create a PR for a new branch in the GitHub UI.

t1:
This fixes https://github.com/eclipse/microprofile-graphql/issues/111

phillip-kruger:
I like it. I spoke to @radcortez to see if we should move power-annotations to SmallRye. @radcortez see examples mentioned above

phillip-kruger:
@t1 - we can move power annotation to smallrye commons. Let's chat on gitter to discuss

t1:
Another option would be to merge it directly into Jandex. In that way, all implementations using Jandex would benefit without needing to change their code.

t1:
Now it already works without any change to the Smallrye GraphQL code base! I have tested it with a class that uses a Stereotype for the `@GraphQLApi` annotation and a Mixin for the `@Query` annotation [here](https://github.com/t1/power-annotations/tree/trunk/demos/demo-war).

t1:
We may want to update all MP GQL annotations to allow them to be used in Stereotypes by adding `@Target(ANNOTATION_TYPE)`.

phillip-kruger:
Will you log an issue in mp ?

t1:
> Will you log an issue in mp ?

sure :-)

t1:
> How would this work if you do not use maven to get the Jandex index ? (i.e jandex index is available, like with WF and Quarkus?)

I've already extracted the actual resolution code to a `commons` module, so we can easily add it to a runtime scanner.

t1:
#569 still keeps the implementation separate, i.e. you'll have to use the `power-jandex-maven-plugin` to see any effect.

[...]

PR #524, opened 2020-11-20
typesafe client: `@OutputOnly`

### LINKED ISSUE #522: `@NonInput`
If a java class is used an `Input` as well as a `Type`, but there are resolver methods in the GraphQL service, then some fields are returned as output but not allowed as input (see also #517). Instead of defining two separate classes with most of the fields duplicated, it would be convenient to mark some fields as `@NonInput`, so the typesafe client checks that they are actually `null` and then doesn't send them at all to the service.

t1:
fixes #522 and a minor bug when an error has a path to a nested field but the container value is null

### DIFF
--- a/client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/OutputOnly.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/OutputOnly.java
@@ -1,0 +1,14 @@
+package io.smallrye.graphql.client.typesafe.api;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+
+/**
+ * Allow this field only when the containing class is used as a <code>Type</code>s, but not
+ * when used as an <code>Input</code>. This happens when the GraphQL service has resolver
+ * methods (see the <code>@Source</code> annotation).
+ */
+@Retention(RUNTIME)
+public @interface OutputOnly {
+}
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/GraphQlClientProxy.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/GraphQlClientProxy.java
@@ -1,5 +1,6 @@
 package io.smallrye.graphql.client.typesafe.impl;
 
+import static javax.json.JsonValue.ValueType.NULL;
 import static javax.ws.rs.client.Entity.entity;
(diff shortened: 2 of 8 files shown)


phillip-kruger (commented):
As I said in the issue, I do not think we should introduce this new annotation, but rather use `@Ignore` as on the server.

t1:
I just non-suggestively asked the developers in my team. All 5 of them said the same thing: using `@Ignore` on the setter needs a lot of explanation and learning; a separate `@OutputOnly` is much easier to understand. We probably should think about adding this to the server, too.

phillip-kruger:
We have discussed this and decided to align with JsonB.
See https://javaee.github.io/jsonb-spec/users-guide.html#ignoring-properties

phillip-kruger:
@t1 - The client should also support the JsonB annotation (as the server do) `@JsonbTransient` and other JsonB annotations.

phillip-kruger:
There are also other annotations that work this way (field = both, setter = input and getter=output). 
`@Name` , `DateFormat`, `NumberFormat` , `Description` and so on. The client should work the same way.

See for example https://download.eclipse.org/microprofile/microprofile-graphql-1.0.3/microprofile-graphql.html#naming

t1:
@phillip-kruger : yes, you're right with everything you say. This is just not where the client currently is!

Annotating a setter as `@JsonbTransient` makes a lot of sense from the perspective of JSON-B where you think about de/serialization. But in the context of a class used as a client side GraphQL Input, it's strange to annotate a setter as `@Ignore`... there is a valid reasoning behind that, but it doesn't come naturally. A separate annotation would make things much easier to grasp.

phillip-kruger:
@andymc12 - how does RESTClient do this ? Since JAX-RS also support JsonB annotation, does RESTClient do the same ?

@t1 - We should rather put effort into getting the client where it needs to be, than add annotations that later potentially will have to be removed.

t1:
Let's talk about this in the next meeting

phillip-kruger:
Would #316 solve this in another way (so you can now in your project define this an an alias ?)

t1:
> Would #316 solve this in another way (so you can now in your project define this an an alias ?)

We would need way to move a `@OutputOnly` field-level Stereotype annotation to a `@Ignore` annotation on the setter... possible, but not here, yet.

phillip-kruger:
Closing, as we decided to do @Ignore(on=INPUT) and this will first need some changes in MicroProfile API

PR #535, opened 2020-11-24
More resilient error handling in GQL client

bamling:
Some GraphQL server implementations return `errors` with the value `null`, which causes a `ClassCastException` during parsing of the response. This small change makes the parsing of the `errors` more resilient.

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/ResultBuilder.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/ResultBuilder.java
@@ -55,6 +55,8 @@ public class ResultBuilder {
     }
 
     private void readErrors() {
+        if (!response.containsKey("errors") || response.isNull("errors"))
+            return;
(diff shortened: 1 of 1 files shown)


phillip-kruger (approved):
LGTM

bamling:
Cleaned up the commit history. Somehow messed something up there, but now it looks better!

phillip-kruger (approved):

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/ResultBuilder.java
```diff
@@ -55,6 +55,8 @@ private JsonObject readData() {
     }
 
     private void readErrors() {
+        if (!response.containsKey("errors") || response.isNull("errors"))
+            return;
         JsonArray jsonErrors = response.getJsonArray("errors");
         if (jsonErrors == null)
             return;
```
t1, line 62: these two line are superfluous, then, aren't they?

t1 (approved):
It's okay by me, but I would have added a test, too.

PR #539, opened 2020-11-25
#316: add power annotations with power-jandex-maven-plugin

t1:
Rewrite of the power-annotations idea (currently `@Stereotype` and `@MixinFor`) with a maven plugin to replace the Jandex maven plugin and do the annotation resolution directly in the Jandex file. In this way, power annotations can already be used by anything based on Jandex!


phillip-kruger:
How would this work if you do not use maven to get the Jandex index ? (i.e jandex index is available, like with WF and Quarkus?)

phillip-kruger (approved):

PR #565, opened 2020-12-09
Get ValidatorFactory from CDI

phillip-kruger:

### DIFF
--- a/server/implementation-cdi/src/main/java/io/smallrye/graphql/cdi/validation/ValidationService.java
+++ b/server/implementation-cdi/src/main/java/io/smallrye/graphql/cdi/validation/ValidationService.java
@@ -3,7 +3,6 @@ package io.smallrye.graphql.cdi.validation;
 import java.lang.reflect.Method;
 import java.util.Set;
 
-import javax.enterprise.inject.spi.CDI;
 import javax.validation.ConstraintViolation;
 import javax.validation.Validation;
 import javax.validation.ValidatorFactory;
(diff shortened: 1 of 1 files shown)


server/implementation-cdi/src/main/java/io/smallrye/graphql/cdi/validation/ValidationService.java
```diff
@@ -39,4 +40,13 @@ public void beforeInvoke(InvokeInfo invokeInfo) throws Exception {
     public String getConfigKey() {
         return ConfigKey.ENABLE_VALIDATION;
     }
+
+    private ValidatorFactory getValidatorFactory() {
+        try {
+            ValidatorFactory validatorFactory = CDI.current().select(ValidatorFactory.class).get();
```
andymc12, line 46: We discussed this in the meeting - could we use the CDI Lookup service to get the ValidatorFactory bean?

server/implementation-cdi/src/main/java/io/smallrye/graphql/cdi/validation/ValidationService.java
```diff
@@ -39,4 +40,13 @@ public void beforeInvoke(InvokeInfo invokeInfo) throws Exception {
...
+            return validatorFactory;
+        } catch (Throwable t) {
```
andymc12, line 48:
```suggestion
        } catch (Exception ex) {
```

Also mentioned in the meeting, maybe catch Exception or RuntimeException instead of Throwable.

andymc12 (commented):
retroactively, LGTM! :)

Issue #579, opened 2020-12-21
Schema build for Interface inheritance sets wrong parent interface


velias:
Discussions in #574 and my related investigation found that schema build for Interface inheritance sets wrong parent interface, it sets interface itself as its parent interface. Problem is in `io.smallrye.graphql.schema.creator.type.InterfaceCreator.addInterfaces(InterfaceType, ClassInfo)`.
I'm going to patch it.

velias:
@phillip-kruger I hope it is OK for you if I add some interface hierarchy into "Profile" part of the TCK to cover this problem. I want to add `IUser` interface with `getUsername()` and let `Profile` object to implement it. Then add `IEntity` interface with `getId()` and let `IUser` to extend it.

velias:
@phillip-kruger implementing given structure of interfaces in TCK, I found another topic to talk about as it is not covered in the MP spec (interface inheritance is not covered there at all ;-) 
I have to add `getId()` method into `IUser` interface also, because I get GraphQL schema validation error `interface type 'IUser' does not implement interface 'IEntity' because field 'id' is missing` if it is not there.
I'd say that schema builder itself should add all methods/fields from all parent interfaces automatically not to require developer to do it manually. WDYT?

velias:
Aha, there was an attempt to add fields from super interfaces, but it was implemented incorrectly. I patched it and will provide it in the PR for this issue.

velias:
Closing as PR has been merged

PR #580, opened 2020-12-21
Schema generation patch for interface inheritance 

velias:
patched list of parent interfaces and fields from parent interfaces as desciebed in issue #579


phillip-kruger (approved):
Thanks @velias !!

PR #595, opened 2021-01-14
Create GraphQL on every request

### LINKED ISSUE #592: Batched queries are always cached by underlying implementation
graphql-java caches queries resolved by a BatchLoader by default, which caused me much confusion working on a prototype using smallrye until I started digging through the call stack when debugging to figure out why my queries were only getting execution once. Smallrye should either disable the caching entirely when creating the BatchLoader or make it configurable.

Some personal input: I think having the GraphQL implementation handle caching at all puts this logic at completely the wrong level, outside of developer control, so I believe disabling caching entirely and making this non-configurable is the ideal solution.

phillip-kruger:
Proposal to fix #592

@snuxoll - please have a look.

@andymc12 - can you also have a look? Might be worth it to run a performance test ?

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionService.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionService.java
@@ -62,7 +62,7 @@ public class ExecutionService {
 
     private final EventEmitter eventEmitter;
 
-    private GraphQL graphQL;
+    private GraphQL.Builder graphqlBuilder;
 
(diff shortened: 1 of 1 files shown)


snuxoll:
@phillip-kruger Thanks, I'll test my sample against your branch here today and see if it does the trick.

andymc12 (approved):
LGTM

phillip-kruger:
Closing in favour of a better fix coming.

PR #601, opened 2021-01-19
Update to OpenTracing 0.33

radcortez:


radcortez:
Requires #600. Added just to see if we pass tests.

phillip-kruger:
LGTM, but I would rather that @jmartisk also have a look, as he knows more about this than I do.

jmartisk (commented):
I can't say I fully understand all the changes, and the WF feature pack doesn't yet properly support tracing, so there's not much I can test on that side, but I don't see any obvious problem

phillip-kruger (approved):

PR #602, opened 2021-01-19
Make sure the batch loader is not cached between requests.

phillip-kruger:
New proposed fix for #592

@snuxoll - please have a look and let me know. I have confirmed that this is now working as expected using your example, but it would be good to have another pair of eyes on it.

@andymc12 - can you also have a look? Might be worth it to run a performance test ?

Please note that this PR is against the 1.0.x branch (as this is what is used in Quarkus at the moment) and the master (1.1.x) contains MicroProfile 4.0 updates.

Once we are happy with this I will open a PR against master and release 1.0.21 to get this fix into Quarkus.


andymc12 (approved):
I haven't run the changes through the performance benchmarks, but they look good to me.

phillip-kruger:
I'll wait a bit longer so that @snuxoll can comment, before I merge. Then I'll apply this to master and get it into Quarkus. Thanks @andymc12 !

PR #619, opened 2021-01-26
Batch exception fix when unwrapping exception

### LINKED ISSUE #590: Batch: CompletionException can not be unpacked
I have a batch resolver that throws an exception. The response then contains for every entry the same error (which is okay), but it's only a `java.util.concurrent.CompletionException` with error code `completion`, but I need access to the exception that caused the failure. It doesn't help to add the `CompletionException` to the `smallrye.graphql.unwrapExceptions` config.

phillip-kruger:
Fix #590 


phillip-kruger:
@t1 - added a test and now also a fix ! I'll do a back port for 1.0.x

phillip-kruger:
@t1 this PR is waiting for you .... please review :)

server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/AbstractDataFetcher.java
```diff
@@ -48,10 +43,6 @@ public AbstractDataFetcher(Operation operation, Config config) {
         this.argumentHelper = new ArgumentHelper(operation.getArguments());
         this.partialResultHelper = new PartialResultHelper();
         this.batchLoaderHelper = new BatchLoaderHelper();
-        if (config != null && config.getUnwrapExceptions().isPresent()) {
```
t1, line 51: I would do this with an `ifPresent` block, but that's just a matter of taste, I guess.

server/implementation/src/main/java/io/smallrye/graphql/execution/error/ExceptionHandler.java
```diff
@@ -19,15 +24,21 @@
...
     private final ExceptionLists exceptionLists;
+    protected List<String> unwrapExceptions = new ArrayList<>();
 
     public ExceptionHandler(Config config) {
         this.config = config;
         this.exceptionLists = new ExceptionLists(config.getHideErrorMessageList(), config.getShowErrorMessageList());
+        if (config.getUnwrapExceptions().isPresent()) {
+            this.unwrapExceptions.addAll(config.getUnwrapExceptions().get());
+        }
+        this.unwrapExceptions.addAll(DEFAULT_EXCEPTION_UNWRAP);
```
t1, line 35: Except for the `DataFetcherException`, this (and the constant) is a code duplication with `AbstractDataFetcher`. Maybe it would be better to push that into the `Config` class, so this could would be only `this.unwrapExceptions = config.getAllUnwrapExceptions()`.

server/tck/src/test/java/io/smallrye/graphql/test/apps/batch/api/BatchApi.java
```diff
@@ -0,0 +1,60 @@
...
+        List<BatchPojo> l = new ArrayList<>();
+        l.add(new BatchPojo(1));
+        l.add(new BatchPojo(2));
+        l.add(new BatchPojo(3));
+        l.add(new BatchPojo(4));
+
+        return l;
+    }
+
+    // Nornal Source
```
t1, line 26:
```suggestion
    // Normal Source
```

t1 (approved):
With some minor remarks, LGTM, but I don't understand a lot of the context (Uni, etc.)

server/implementation/src/main/java/io/smallrye/graphql/execution/error/ExceptionHandler.java
phillip-kruger: It's been removed (or rather moved) from `AbstractDataFetcher.` So it's not in `AbstractDataFetcher` anymore

server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/AbstractDataFetcher.java
phillip-kruger: This is being deleted :)

PR #653, opened 2021-02-24
Correct reactive error handling in UniDataFetcher

Andro999b:
fix for https://github.com/quarkusio/quarkus/issues/15271

### DIFF
--- a/server/implementation/src/test/java/io/smallrye/graphql/execution/ExecutionTestBase.java
+++ b/server/implementation/src/test/java/io/smallrye/graphql/execution/ExecutionTestBase.java
@@ -5,12 +5,7 @@ import java.io.StringWriter;
 import java.util.HashMap;
 import java.util.Map;
 
-import javax.json.Json;
-import javax.json.JsonObject;
-import javax.json.JsonObjectBuilder;
-import javax.json.JsonValue;
-import javax.json.JsonWriter;
-import javax.json.JsonWriterFactory;
+import javax.json.*;
 import javax.json.stream.JsonGenerator;
 
 import org.jboss.jandex.IndexView;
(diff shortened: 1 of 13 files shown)


cescoffier (dismissed):
The mutiny part is good.

When the function was throwing an exception, mutiny created a composite exception with the original failure and the thrown exception.

server/implementation/src/test/java/io/smallrye/graphql/execution/ExecutionTestBase.java
```diff
@@ -5,12 +5,7 @@
 import java.util.HashMap;
 import java.util.Map;
 
-import javax.json.Json;
-import javax.json.JsonObject;
-import javax.json.JsonObjectBuilder;
-import javax.json.JsonValue;
-import javax.json.JsonWriter;
-import javax.json.JsonWriterFactory;
+import javax.json.*;
```
phillip-kruger, line 8: Keep the imports specific (as it was)

phillip-kruger (changes requested):
Nice !! Thanks very much @Andro999b . One small change and maybe you can add a test in the TCK module ?

Java code here: https://github.com/smallrye/smallrye-graphql/blob/master/server/tck/src/test/java/io/smallrye/graphql/test/apps/error/api/ErrorApi.java

And graphQL query here: https://github.com/smallrye/smallrye-graphql/tree/master/server/tck/src/test/resources/tests/errors

Thanks !

phillip-kruger (approved):
Thanks @Andro999b !!!

phillip-kruger:
@Andro999b - Thanks for this. Can you please also add this to the 1.0.x branch (master  is 1.1.x that has not yet been included in any runtime, it's waiting for MP4) 1.0.x is the version currently in Quarkus

PR #660, opened 2021-02-26
fix #634: don't reuse rest client & more

### LINKED ISSUE #634: Typesafe Client: don't reuse RestClient instance and close when closeable
The `GraphQlClientBuilderImpl` uses a constant for `ClientBuilder.newClient()`, but that means registered filters will be reused for other clients.

And the MP Rest Client closes some things when the interface is closeable; we should do the same.

t1:
don't invoke methods inherited from Object
close client when closeable api is closed

Signed-off-by: Rüdiger zu Dohna <snackbox@sinntr.eu>


phillip-kruger (approved):

phillip-kruger:
Are you also backporting this ?

PR #680, opened 2021-03-14
667 render.js path

t1:


phillip-kruger:
I think this will break Quarkus. Give me a few days to check. In Quarkus we basically override the first two line:

```
const api = '/graphql';
const ui = '/graphql-ui';
```

Both is configurable and not always relative. I think WF will have to do the same. We can maybe make that easier, so create a java api that can be called with parameters that return the bytes of render.js.

We do something like that for OpenAPI

phillip-kruger:
Closing as discussed

PR #682, opened 2021-03-14
fix #624: typesafe client: `@Name` annotation on methods, too

### LINKED ISSUE #624: Typesafe Client: use `@Name` on method
When a method is named with `@Name`, use that name instead of the method name.

t1:
(but not `@JsonbProperty`, yet)

Signed-off-by: Rüdiger zu Dohna <snackbox@sinntr.eu>

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/ParameterInfo.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/ParameterInfo.java
@@ -3,8 +3,6 @@ package io.smallrye.graphql.client.typesafe.impl.reflection;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Parameter;
 
-import javax.json.bind.annotation.JsonbProperty;
-
 import org.eclipse.microprofile.graphql.Id;
 import org.eclipse.microprofile.graphql.Input;
 import org.eclipse.microprofile.graphql.Name;
@@ -106,12 +104,9 @@ public class ParameterInfo {
     public String getName() {
         if (parameter.isAnnotationPresent(Name.class))
             return parameter.getAnnotation(Name.class).value();
-        JsonbProperty jsonbProperty = parameter.getAnnotation(JsonbProperty.class);
-        if (jsonbProperty != null && !jsonbProperty.value().isEmpty())
-            return jsonbProperty.value();
         if (!parameter.isNamePresent())
             throw new GraphQlClientException("Missing name information for " + this + ".\n" +
-                    "You can either annotate all parameters with @Name or @JsonbProperty, " +
+                    "You can either annotate all parameters with @Name, " +
                     "or compile your source code with the -parameters options, " +
                     "so the parameter names are compiled into the class file and available at runtime.");
         return parameter.getName();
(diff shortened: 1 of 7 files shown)


client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/ParameterInfo.java
```diff
@@ -104,9 +106,12 @@ public Object getValue() {
     public String getName() {
         if (parameter.isAnnotationPresent(Name.class))
             return parameter.getAnnotation(Name.class).value();
+        JsonbProperty jsonbProperty = parameter.getAnnotation(JsonbProperty.class);
```
phillip-kruger, line 109: Can you not use Jandex here ? Then we do not need the optional dependency on Jakarta json bind

t1: Yeah, I already thought about that. But that would only make sense if _all_ reflection would be done via Jandex. I want to do that anyways, so I can fully use Power-Annotations, but that would mean a larger rewrite.

Should I remove the dependency and thereby the support for JSON-B for now? We currently don't need it in our project.

phillip-kruger: Maybe we just add a ticket so that we remember to do this later ? Also, maybe change to provided ? And maybe handle classnotfound ? I am not sure to be honest. Maybe we can discuss this afternoon ?

phillip-kruger:
@t1 As discussed - waiting on you to decide how to proceed. Let me know .

t1:
I've removed the dependency on `JSON-B`. Everything else would have been half-hearted.

client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/ParameterInfo.java
t1: Filed the Jandex topic as #693

pom.xml
```diff
@@ -177,7 +178,12 @@
                 <groupId>jakarta.validation</groupId>
                 <artifactId>jakarta.validation-api</artifactId>
                 <version>${version.jakarta-validation}</version>
-                <scope>provided</scope>
+            </dependency>
+
+            <dependency>
+                <groupId>jakarta.json.bind</groupId>
+                <artifactId>jakarta.json.bind-api</artifactId>
+                <version>${version.jsonb}</version>
```
phillip-kruger, line 186: We can remove this now right ?

phillip-kruger (changes requested):

pom.xml
t1: Yes, of course. Sorry. I just had seen that there are other modules that also depend on json-b, but they inherit the version from the `smallrye-parent-pom`, so it was wrong anyway.

pom.xml
```diff
@@ -177,7 +177,6 @@
                 <groupId>jakarta.validation</groupId>
                 <artifactId>jakarta.validation-api</artifactId>
                 <version>${version.jakarta-validation}</version>
-                <scope>provided</scope>
```
phillip-kruger, line 180: Sorry, I should have asked this in the prev review, why are we removing provided ?

phillip-kruger (changes requested):

pom.xml
t1: sorry, that was an accidental change

phillip-kruger (approved):
Thanks !

phillip-kruger:
I am planning a 1.0.x release soon, do you want this in there too ?

t1:
No, we're on 1.1 already.

PR #699, opened 2021-03-19
fix #627: typesafe client: multi query

### LINKED ISSUE #627: Support for multiple queries in same request (Client)
Is there a possibility to make multiple queries in same request ? 
Example: 
`{
    findHero(id: 1) {
         name
    }
    findVillain(id: 2) {
        name
    }
}`

In this case there are 2 queries that are made in the same http request. Is this possible with current client API ? 

P.S. This is a great API, great work. :)

t1:


phillip-kruger (changes requested):
there is a Multi in Mutiny .... I wonder if we should not name this something else to not confuse the two ? How about writing the full name (Multiple) ?

t1:
>  How about writing the full name (Multiple) ?

Sure!

phillip-kruger (approved):

PR #715, opened 2021-03-30
Fix #702 correctly identify get/setters…

### LINKED ISSUE #702: Getters aren't recognised correctly
Currently, all methods that start with `get` and a following capital letter are recognised as getters, regardless of parameters. Something like, for example, (Map's) `getOrDefault(K key, V value)` is added to the schema as `orDefault`, but at runtime it is (correctly) ignored.
Such methods shouldn't be added to the schema.

ybroeker:
…by name, return type and parameters


phillip-kruger (approved):
Thanks @ybroeker !

phillip-kruger:
If you want this in Quarkus soon (i.e before Quarkus 2) then please backport to the 1.0.x branch

PR #719, opened 2021-03-31
Fix #700, fix #701

### LINKED ISSUE #700: `static` fields shouldn't be added to Types & Inputs
Currently, `static` fields are added to the schema on types and inputs, which means constants must be explicitly ignored. 
Those fields should be ignored by default.

Especially for inputs, as these field cannot be used at runtime.

### LINKED ISSUE #701: `final` fields should be ignored in inputs
Currently these are added to the schema but ignored at runtime.

ybroeker:
Check modifiers while considering a field and/or method as a graphql field.


phillip-kruger (approved):
Thanks @ybroeker !

PR #729, opened 2021-04-06
Fix #728: Use WeakHashMap to prevent memory leaks

### LINKED ISSUE #728: Memory leak in `SmallRyeBatchLoaderContextProvider.INSTANCES`
With each request, objects are stored in the `SmallRyeBatchLoaderContextProvider.INSTANCES`, but never deleted. 

Depending on the number of batch operations and available memory, this leads more or less quickly to extremely slow requests (and almost 100% GC) and at some point to an OutOfMemoryError.

ybroeker:
Fixes the memory leak by using WeakHashMap.

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeBatchLoaderContextProvider.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeBatchLoaderContextProvider.java
@@ -1,7 +1,6 @@
 package io.smallrye.graphql.execution.context;
 
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


jmartisk (commented):
Changes LGTM. Please rebase

jmartisk (approved):
Maybe we could remove the map entry for the `DataLoader` ourselves at some point, but I'm not familiar with the lifecycle of DataLoader objects so I'm not sure when exactly we can do this, and I'd rather not assume too much about how graphql-java will behave. The solution with a weak map, while not necessarily the most efficient, looks at least safe enough, so I'm +1 on this

phillip-kruger (approved):
Nice ! Thanks @ybroeker !

Issue #704, opened 2021-03-21
Support for Java 16 records


ybroeker:
Records would be super useful for Types (and maybe also inputs).

Currently, a (actually superfluous) explicit getter has to be added for each record component. It would be better if records were supported "out of the box".

Only the schema builder would have to be adapted, the Java version would not have to be adapted and could remain at 8.

----

An alternative (or an additional possibility) would be to be able to mark methods with annotations as fields, e.g. with `@Query`. A method `@Query public field()` (in a non-`@GraphQLApi` class) would then add the field `field`. This could be used together with records, but would also offer the possibility to be used in "normal" classes.

phillip-kruger:
Agree. The alternative has been discussed before (I know SPQR works like that).

t1:
I just had a second thought about this: is `@Query` the right annotation for this? Why not also `@Mutation`? Or shouldn't those two be reserved for the root `Query`/`Mutation` types? We could add a new `@Field` annotation. I think this could be more clear for users.

phillip-kruger:
I think we can close this one. If I am wrong please re-open

robp94:
The support for records is only for the server types, inputs and on the client side they seem not to be supported. Could this be added?

t1:
Yeah, that makes sense. Could you open a new issue?

PR #732, opened 2021-04-06
Allow `@Query` on POJO-methods, refs #704

ybroeker:
This PR allows to use `@Query` on POJO-methods, so methods don't need to be prefixed with `get`/`is`.

#704 should not be closed yet, native record support would still be desirable. But at least it allows to annotate components with `@Query` instead of implementing additional getters.

I'm not sure what the backport policy is, if desired I can add that to 1.0.x as well.


phillip-kruger (approved):

phillip-kruger:
Yes please backport as well
Thanks @ybroeker !

phillip-kruger:
Hi @ybroeker, hope you are well, seems this PR broke Quarkus in native mode :( 
see  https://github.com/quarkusio/quarkus/issues/17349

I'll look at this a.s.a.p, but if you have an idea what the problem is, please share :)

ybroeker:
Apparently MethodHandles have to be enabled via an explicit flag in Graalvm 20.3 (at least the part that is used), and full support has only been added with 21.0 (https://github.com/oracle/graal/issues/2761).

It should be easy to replace MethodHandles in `FieldDataFetcher`, they are only used to be able to access fields and methods in the same way.

Can create a pull request later

phillip-kruger:
Great ! Thanks :)

Issue #628, opened 2021-02-03
Support custom directive


phillip-kruger:
see https://github.com/eclipse/microprofile-graphql/issues/344

t1:
@KSmigielski : I need this for #521. Have you already started? Otherwise I will start working on this now.

t1:
We don't have full support, yet; just type directives. We at least need also field directives. I haven't considered the other possible locations and if they make sense in MP GraphQL. Reopening.

t1:
I think we can close this when #810 has been merged. If locations other than type and field become necessary in the future, we can open a new issue.

PR #736, opened 2021-04-06
Enable usage of `@JsonbCreator`

### LINKED ISSUE #705: Support for @JsonbCreator (or similar mechanisms)
Currently, a public field/setter must exist for each field of inputs.

It would be nice if `JsonbCreator` or a similar mechanism was supported. Instead of explicit setters, a constructor or a factory method would then suffice.

Also, until explicit support is available, it would allow to use records as inputs.

ybroeker:
Allows to use constructor or factory methods instead of just setters or public fields for input types.

Can backport this to 1.0.x, but would wait for a backport of #628 if any is planned?

----
Closes #705.

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/Annotations.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/Annotations.java
@@ -102,6 +102,16 @@ public class Annotations {
         }
     }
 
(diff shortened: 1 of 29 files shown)


phillip-kruger (approved):
Thanks @ybroeker

phillip-kruger:
@t1 - what do you think ? Can we (you ;) ) backport #628 ?

PR #741, opened 2021-04-11
Move typesafe client tests to tck

t1:
and lazy init JAX-RS client, so the tests don't create one.

Building on this I will create another PR to reverse the dependency from tck→impl to impl→tck


client/tck/src/test/java/test/unit/ConfigBehavior.java
```diff
@@ -102,37 +97,6 @@ void shouldLoadAnnotatedKeyConfig() {
...
-        DummyClientRequestFilter instance = new DummyClientRequestFilter();
-        Api api = ((JaxRsTypesafeGraphQLClientBuilder) fixture.builder())
-                .register(instance)
-                .build(Api.class);
-
-        api.foo();
-
-        verify(fixture.client()).register(instance);
-    }
-
```
t1, line 135: These tests don't make any sense any more, as this is not part of the public API any more

phillip-kruger (approved):

phillip-kruger:
Nice @t1 !

PR #747, opened 2021-04-17
#627: change `@Multiple` target from `METHOD` to `TYPE`

t1:
As discussed in the last meeting. This is actually a breaking change, but IIRC it hasn't been released yet.

And more sensible test, which discovered a bug: we need to use raw field names for variable names (important for multiple parameters/fields mapped to the same name)

And some typos

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/MethodInvocation.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/MethodInvocation.java
@@ -1,14 +1,16 @@
 package io.smallrye.graphql.client.typesafe.impl.reflection;
 
+import static java.util.stream.Collectors.toList;
+
 import java.io.Closeable;
 import java.lang.annotation.Annotation;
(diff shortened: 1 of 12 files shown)


client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/reflection/MethodInvocation.java
```diff
@@ -116,11 +119,13 @@ public boolean hasNestedParameters(String path) {
...
-                .mapToObj(i -> new ParameterInfo(this,
-                        parameters[i],
-                        parameterValues[i]));
+        if (this.parameters == null)
+            this.parameters = IntStream.range(0, method.getParameterCount())
+                    .mapToObj(i -> new ParameterInfo(this,
+                            method.getParameters()[i],
+                            parameterValues[i]))
+                    .collect(toList());
+        return parameters.stream();
```
t1, line 128: The query will be cached, but the parameters are accessed many times, so this should improve warmup times

phillip-kruger (approved):

PR #775, opened 2021-05-05
Add support for graphql extensions

ybroeker:
Allows to use extensions like [Apollo Tracing](https://github.com/apollographql/apollo-tracing).

Such extensions can currently be added with a custom `EventingService`. Direct support for apollo-tracing (which is included in graphlql-java) could be added with another PR.


phillip-kruger:
Nice ! Maybe we can introduce a new annotation, that allows devs to create classes that will be extensions ? Then we scan for them when we build the schema ?

Issue #713, opened 2021-03-29
Duplicate selectedFields/ fields when using @Context?


janknobloch:
Hey there,

im using the quarkus-graphql extension and seeing duplicates in the fieldList (selectedFields) and hope here is the right place to post this issue as its not quarkus specific I guess. 

Using the following setup on **quarkus 1.12.2**

```
@Inject
Context context;

@Query("someName")
 public ComplexType someName(){

 DataFetchingEnvironment dfe = context.unwrap(DataFetchingEnvironment.class);
     
        for (SelectedField f : dfe.getSelectionSet().getFields())
        {
            LOG.info(f);
        }
        
        for (JsonValue jvalue : context.getSelectedFields()) {
            LOG.info(jvalue);
        }

}
```

sending a query: 

 ```
{
  someName( ) {
    pageType
    stopReason
    incomingUrl
   }
}
```

I receive duplicate entries for all 3 fields queried for (example above of course 4 times each - just to highlight this is happening for both for the DFE and context versions):

pageType
pageType
stopReason
stopReason
incomingUrl
incomingUrl

I am not using any fragments in my query so I assume retrieving _mergedField_ should not be necessary at this point.

Expected behaviour would be to receive each element only once?

Maybe its a usage error on my side, thanks for your support.

Let me know if I can assist further.

Best
Jan

**Comment by phillip-kruger** [comment]
@jmartisk ^^^^ - maybe this got introduced during our latest context fixes ?

**Comment by jmartisk** [comment]
I'll try to investigate and let you know shortly

**Comment by janknobloch** [comment]
I'll did a fallback to quarkus 1.11.2 where the issue is also present (for me) - maybe this helps.

**Comment by jmartisk** [comment]
This looks related to the upgrade to graphql-java 16.0 and a weird behavior that it exhibits. It's related to this change: https://github.com/graphql-java/graphql-java/commit/afaa9aa1f321ef4aed7ef6ccac3e2efdd8413874#diff-368edec8757011bd7775450e7ad220541847811f6b6710db206d60a15ea6955e

Basically if there's a query like `{ allPeople { name }}`, then the `SelectionSet` basically contains the `name` field twice - once as "unqualified", once as "qualified" by including the enclosing type:

`dfe.getSelectionSet().contains("Person.name")  // true`
`dfe.getSelectionSet().contains("name")  // true`

If you call `getFields()`, both versions of the field are included separately in the result:
```
[...]

PR #780, opened 2021-05-06
Hotfix for https://github.com/smallrye/smallrye-graphql/issues/713 - …

janknobloch:
Feel free to add this until the guys from _java-graphql_ figured out how to remove field duplicates.

I haven't seen any issues in regards of using a Set over a List, only drawback however is the conversion itself, which should be removed anyways when solved inside the `DataFechingEnvironment`.

Best
Jan


phillip-kruger (approved):
Thanks @janknobloch ! LGTM

PR #783, opened 2021-05-08
Fix/Refactor CollectionCreator

ybroeker:
CollectionCreator contained a few bugs (wrong collection was created in some cases) and had bad performance in some cases (e.g. when `List` was used in an input type, the reflective call failed, which is quite slow).

This PR should fix both.


phillip-kruger (approved):
Thanks @ybroeker !

Issue #770, opened 2021-05-04
performance issue using nested elements in query


janknobloch:
Hi all,

im facing serious performance drops using a nested query for a tree like structure using quarkus 1.13.2 and its graph-ql extension:

when using:

 ```
navigation {
      name
      children {
        name
        children {
          name
          children {
            name
          }
        }
      }
    }
```
im able to reach about 200 req/s via k6s load testing, whereas:

using the exact same query but removing the nested query part and replacing it with a single key:value pair as follows:

navigation : objectMapperOfyourChoice.writeValueAsString(navigation); 

im able to increase throughput from about roughly 200 req/s to 800 req/s using k6s running 1000 VUs and 1000 Req/s aimed for.

To clarify I was able to pinpoint this issue down to the nested vs. non-nested query part for my use-case.
 My navigation can be rather large with up to depth = 8, and about 5-10 properties inside each navigation element, however its always computed in the same way no matter how I return it via graphql.

Any clues on this behaviour? As a workaround I'm using the json string response type without nested query for now, but I would love to eliminate this issue and can try to run some more load testing potential fixes in the upcoming days, if that helps.

I assume the nested query element should not have such a big impact on performance if im not mistaken?

Best
Jan

**Comment by phillip-kruger** [comment]
Hi @janknobloch . Can you share your code ? Or a reproducer ? Are you using `@Source` ? And if so, are you using batch ?

**Comment by janknobloch** [comment]
hey @phillip-kruger, thanks for your fast reply.

hope to share a reproducer for the issue by the weekend at the latest.

in the meantime, a little more information.

Im not using @Source, at least not that im aware of.
Im using @Context to remove some elements not existent in a concrete query (but this should be independent from the issue as the speed drop only occurs when using the query stated).

My Navigation POJO "TreeWrapper" looks like this in the raw form:

```
CoreNavigationElement {
    String val;
    String val2;
    @Ignore
    String val3;
    List<CoreNavigationElement> children;
}
```

Basically when I remove the navigation part of the query the speed drop is gone, but the speed drop is also gone if I forcefully set my navigation object to null and keep the query in place.

It looks like that the model transformation done inside graphql-java seems to be very slow when it comes to recursive lookups or tree like structures? / and or maybe my @ignore field has something to do with it.

I'll try to setup a plain project as a reproducer on the weekend but maybe this issue is actually somehow related to :

https://github.com/graphql-java/graphql-java/issues/2185

[...]

PR #784, opened 2021-05-09
Build json 'by hand'

ybroeker:
Formatting json with JsonB und parse afterwards was slow, formatting by hand should be much faster.
GraphQL returns a limited set of values (Collections, Maps, Numbers, Booleans, Enums), so
building the json by hand is easy. JsonB is used just in case if a different type is encountered.

See #770 for context


t1 (approved):
LGTM

PR #807, opened 2021-05-19
Replace MethodHandles to allow building with graalvm native image 20.3

ybroeker:
See quarkusio/quarkus#17349 and smallrye/smallrye-graphql#732


phillip-kruger (approved):
Awesome ! Thanks @ybroeker ! I think we need to do a release to get this into Quarkus. I'll do that soon. Thanks

PR #810, opened 2021-05-25
#628: custom field directives

t1:
I need these for GraphQL Federation


phillip-kruger (approved):
LGTM ! Thanks @t1 ! Can you take us through this tomorrow (in the meeting)?

PR #815, opened 2021-05-27
Add support for records as graphql types

ybroeker:
Allows to use records as GraphQL-Types without further annotations:

```java
public record SomeType(String someField){}
```

would be mapped to
```graphql
type SomeType {
  someField: String
}
```

Tests are currently not included, as these would have to be built and run with Java 16.

----

If getters are present, they are preferred over record component accessors for backwards compatibility:

```java
public record SomeType(String someField){
  @Description("Description...")
  public String getSomeField() {
    return someField;
  }
}
```

would be mapped to
```graphql
type SomeType {
  "Description..."
  someField: String
}
```

.----

See also #704


t1:
> Nice ! Thanks @t1

T'wasn't me ;-) Thanks go to @ybroeker !!!

phillip-kruger:
Apologies, thanks @ybroeker

Issue #790, opened 2021-05-12
Exposing SmallRyeContext.documentSupplier 


computerlove:
Is there any reasons `documentSupplier` is not exposed through the leaky abstraction? 
We are doing some custom logging of the incoming query that needs the parsed query.

Is there any reasons why the already parsed Document, or its supplier, should not be available through `SmallRyeContext.unwrap(Document/Supplier)`?

phillip-kruger:
No real reason. We can definitely add that.

computerlove:
That would be great :)
Should I make a PR?

Would the best approach be supporting `unwrap(Document.class)`? 
`unwrap(Document.class)` feels a little bit wrong because of type erasure and `Supplier` could be anything.

phillip-kruger:
Yea Let's add Document to the classes we support ....

phillip-kruger:
PR very welcome !

computerlove:
PR #800 pushed. 
Not really sure if the access to the supplier needs to be thread safe, but it's not that complex.
Any ideas for other tests that should be added?

PR #818, opened 2021-05-28
Use QueryCache when supplying Document from Context

computerlove:
#790 added support for getting the parsed query document with `context.unwrap(Document.class)`.
The current and previous version of `SmallRyeContext` reparse the query when using Document. 
This PR attempts to use the QueryCache such that a given query is parsed only once.


computerlove:
Getting alot of `Caused by: java.lang.ClassNotFoundException: javax.ws.rs.core.MultivaluedMap` when building the TCK-module. 
Don't understand the reason.

phillip-kruger (approved):
Nice ! Thanks @computerlove !

PR #819, opened 2021-05-31
Prevent NPE in SmallRyeContext.DocumentSupplier when queryCache not set

computerlove:
#818 added `SmallRyeContext.withDataFromExecution(ExecutionInput, QueryCache)`, leaving the old method `withDataFromExecution(ExecutionInput)`. 
As `SmallRyeContext.unwrap(Document.class)` is right now it requires queryCache being set. 
This PR falls back to parsing the query if queryCache is not set.


phillip-kruger (approved):

phillip-kruger:
Thanks @computerlove !

PR #827, opened 2021-06-01
Invoke eventEmitter.fireOnDataFetchError for all exceptions in AbstractDataFetcher

computerlove:
All classes extending `AbstractDataFetcher` call `eventEmitter.fireOnDataFetchError` for all exceptions, but the default in `AbstractDataFetcher` only calls `eventEmitter.fireOnDataFetchError` for certain exceptions.


phillip-kruger (approved):
LGTM

PR #883, opened 2021-06-29
emit descriptions for enum values using `@Description`-annotation

Felk:


Felk:
I'm not sure if this should be added anywhere to the microprofile-graphql tck, so I went ahead and did only this for now.

phillip-kruger (approved):
LGTM

Issue #519, opened 2020-11-17
`mp.graphql.showErrorMessage` - wildcard


t1:
For larger projects it would be very convenient to configure a wildcard like `mp.graphql.showErrorMessage=com.superheroes.*` to match all exceptions in a package and sub-packages.

Analogous for `mp.graphql.hideErrorMessage`.

noon2a:
.

parsharma:
Hi @t1,
Are you working on this issue? Could I give it a try?

Thanks

t1:
No, I don't... just go ahead!

parsharma:
Thanks, Could you please give me some more details about it?

phillip-kruger:
So a.f.a.i.k, at the moment, to add exceptions on the hide or show list, you need the full package class name(s).
This feature will allow you to use wildcards. Possibly we should just support regex and use pattern matching ? @t1 - does that some this up ?

t1:
I hadn't considered regex. It's definitively more powerful and could come in handy, but it could also lead to confusion, e.g. that `foo.*` also matches classes in a package `foobar`. Do you see regex often in config strings? We also could start simple and add some form of expressions only when we really need it; something like `foo.${regex:.*}`. What do you think?

phillip-kruger:
I am easy, agree we can add regex later. @parsharma - Is this enough so you can work on this ?

PR #935, opened 2021-07-28
mp.graphql.showErrorMessage - wildcard

parsharma:
Issue: https://github.com/smallrye/smallrye-graphql/issues/519

I have written the test cases but that's not working correctly as Runner is not picking config properties.


server/tck/src/test/resources/META-INF/microprofile-config.properties
```diff
@@ -2,6 +2,7 @@ smallrye.graphql.printDataFetcherException=true
 smallrye.graphql.tracing.enabled=true
 smallrye.graphql.allowGet=true
 smallrye.graphql.logPayload=true
-mp.graphql.showErrorMessage=java.security.AccessControlException
-mp.graphql.hideErrorMessage=java.io.IOException
+
+mp.graphql.showErrorMessage=java.security.AccessControlException,io.smallrye.graphql.test.apps.exceptionlist.MyUncheckedException
```
phillip-kruger, line 6: `mp.graphql.showErrorMessage=java.security.AccessControlException,io.smallrye.graphql.test.apps.exceptionlist.*`

server/tck/src/test/resources/META-INF/microprofile-config.properties
```diff
@@ -2,6 +2,7 @@ smallrye.graphql.printDataFetcherException=true
...
+mp.graphql.hideErrorMessage=java.io.IOException,org.eclipse.microprofile.graphql.tck.apps.superhero.api.*,io.smallrye.graphql.test.apps.exceptionlist.MyCheckedException
```
phillip-kruger, line 7: `mp.graphql.hideErrorMessage=java.io.IOException,io.smallrye.graphql.test.apps.exceptionlist.*`

phillip-kruger (changes requested):
Change those two config value, then we test both hide and show for wildcard

server/implementation/src/main/java/io/smallrye/graphql/spi/config/Config.java
```diff
@@ -86,8 +86,13 @@ default boolean isListed(Class throwableClass, List<String> classNames) {
...
 
         // Check that specific class
-        if (classNames.contains(throwableClass.getName())) {
-            return true;
+        for (String configuredValue : classNames) {
+            if (configuredValue.equals(throwableClass.getName())) {
+                return true;
+            } else if (configuredValue.endsWith("*")) {
+                String values = configuredValue.substring(0, configuredValue.length() - 2);
+                return throwableClass.getName().startsWith(values);
```
phillip-kruger, line 94: Here only return if true. If false, we still want to iterate the rest of the for loop

parsharma: Thanks, you are right. We need to iterate the loop if it is false.

phillip-kruger:
@parsharma - seems like the TCK is passing now ?

phillip-kruger:
Ok, but there are not test cases to test the wildcard ?

parsharma:
Yes, tck is passing now. I will add the test cases, I am just checking why the runner is not picking values.

phillip-kruger (approved):

PR #939, opened 2021-08-02
Allow interfaces to be mapped as types or inputs

### LINKED ISSUE #926: Java interface should only be mapped to GraphQL interface if `@Interface` is annotated
Hi,

I tried to work further on the Blaze-Persistence integration with MP GraphQL but I hit another roadblock because I get the following error:

```
interface type 'SomeInterface' cannot implement 'SomeInterface' because that would result on a circular reference
```

That is because I have a query that returns the type 'SomeInterface' which is a Java interface. But why would SmallRye GraphQL even care what kind of class this is for the GraphQL schema? IMO a Java interface should be modeled as GraphQL object type unless it was annotated as `@Interface`.

I tried using SmallRye GraphQL 1.0.26.Final with Quarkus 1.13.7.Final

phillip-kruger:
Fix #926

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/ReferenceCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/ReferenceCreator.java
@@ -177,8 +177,10 @@ public class ReferenceCreator {
         // we figure out this is actually an enum or interface
         ReferenceType referenceType = getCorrectReferenceType(direction);
 
+        Annotations annotationsForClass = Annotations.getAnnotationsForClass(classInfo);
+
         // Now check if this is an interface or enum
-        if (Classes.isInterface(classInfo)) {
+        if (isInterface(classInfo, annotationsForClass)) {
             // Also check that we create all implementations
             Collection<ClassInfo> knownDirectImplementors = ScanningContext.getIndex()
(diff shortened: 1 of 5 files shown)


robp94:
Server. For example, the added test. https://github.com/phillip-kruger/smallrye-graphql/blob/fc3ad76f8618ca89817399a4d9273fdb83058d4e/server/tck/src/test/java/io/smallrye/graphql/test/apps/interfaces/api/FoodResource.java

phillip-kruger:
So are you asking how you would know what the concrete class is ?

robp94:
When I add a mutation to the linked api like this:
```
    @Mutation
    public Eatable inputEatable(Eatable eatable) {
        return eatable;
    }
```

I get the following error
`Class org.acme.Eatable is used as input, but does neither have a public default constructor nor a JsonbCreator method`

I don't see how to provide an JsonBCreator at the interface, for example, like what the concrete class is. Could you provide an example in the test, or is this not supported, and I misunderstood the pull request?

phillip-kruger:
mmm, so the schema generate ok (that is what this PR did) but it fails at runtime ? Please open a bug, referencing this and if possible add a reproducer.

robp94:
Quarkus fails to start.

Issue #855, opened 2021-06-15
Returning a SortedSet will return unsorted values


sacnoth:
From an `@GraphQLApi` endpoint I want to return a `SortedSet` of comparable data.  
(Reason for choosing a `SortedSet` over a `List`: Wanting to prevent duplicates in the result, but still having an order in the result)

Unfortunately the originally sorted set will end up being unsorted once it's queried via GraphQL.

```Java
    @Query
    public SortedSet<@NonNull MySortableType> sortedValues() {
        return new TreeSet<>(List.of(
                new MySortableType(1, 999),
                new MySortableType(2, 666),
                new MySortableType(3, 333),
                new MySortableType(4, 999),
                new MySortableType(5, 666),
                new MySortableType(6, 333),
                new MySortableType(7, 999),
                new MySortableType(8, 666),
                new MySortableType(9, 333)
        ));
    }
```

Query:
```GraphQL
query{
  sortedValues {value}
}
```

Expected result:
```JSON
{
  "data": {
    "sortedValues": [
      {
        "value": 1
      },
      {
        "value": 2
      },
      {
        "value": 3
      },
      {
        "value": 4
      },
      {
        "value": 5
      },
      {
        "value": 6
      },
      {
        "value": 7
      },
      {
        "value": 8
      },
      {
        "value": 9
      }
    ]
  }
}
```

Actual result:
```JSON
{
  "data": {
    "sortedValues": [
      {
        "value": 1
      },
      {
        "value": 4
      },
      {
        "value": 7
      },
      {
        "value": 2
      },
      {
        "value": 5
      },
      {
        "value": 8
      },
      {
        "value": 3
      },
      {
        "value": 6
      },
      {
        "value": 9
      }
    ]
  }
}
```

As far as I can tell the culprit lies in the `io.smallrye.graphql.execution.datafetcher.CollectionCreator` where:
```Java
    private static Collection<?> newCollection(Class<?> type) {
        try {
            return (Collection<?>) type.getDeclaredConstructor().newInstance();
        } catch (Exception ex) {
            log.noArgConstructorMissing(type == null ? "null" : type.getName());
        }
        if (Set.class.isAssignableFrom(type)) {
            return new HashSet<>();
        }
        return new ArrayList<>();
    }
```
it will return a `new HashSet<>()` if no constructor was found.

As the `SortedSet` is an interface no constructor will be found.

Workaound:
```Java
    @Query
[...]

PR #942, opened 2021-08-02
Added tests  for SortedSet.

phillip-kruger:


phillip-kruger:
@sacnoth - I could not recreate #855. See this test case added that always return the set sorted. Can you share more details on your environment ?

sacnoth:
What I've committed at https://github.com/smallrye/smallrye-graphql/pull/942 is how I was able to reproduce the issue on my end when I created the ticket.

I hope this will help.

phillip-kruger:
Thanks, I'll have a look in the morning

Issue #521, opened 2020-11-18
Add support for federation


phillip-kruger:
Might be a lot of work, but let's consider support for federation

https://www.apollographql.com/docs/federation/managed-federation/overview/

chet20r:
Thanks Phillip, I think Federation support would be really helpful given organizations adopting microservices pattern. Especially with large organizations with varied technology stacks who want to have a single unified data graph combining multiple microservices on different technology platforms.

There is already a support for federation on the JVM via apollo [federation-jvm](https://github.com/apollographql/federation-jvm) package. Probably worth looking if that can be leveraged here?

t1:
This is about services taking part in a federation, right? Writing a federation gateway is a completely different and much bigger topic. I've read several blogs and watched several talks trying to understand esp. the difference between Schema Stitching and Federation. For me, [this](https://www.apollographql.com/blog/the-evolution-of-graphql-at-scale/) blog post explained it best.

My ultra-short summary: Schema Stitching works with custom code in the Gateway that hooks the endpoints together. While this approach starts very simple, complexity grows exponentially, as there is a single point that has to know about _all_ connections. Federation OTOH distributes this responsibility to the separate services, e.g. a user service only knows about the `User` type, while a review service _extends_ the `User` type with a `reviews: [Review!]!` field while only knowing about the id of the user. This and a few more things is done with quite flexible directives; and the gateway then does all the Federation magic on a generic level.

t1:
Maybe it would be sufficient to create a separate library like `smallrye-graphql-federation` that contains the required types and some [custom directives](https://github.com/eclipse/microprofile-graphql/issues/344)?

t1:
The correct link to start from is https://www.apollographql.com/docs/federation/federation-spec/

The managed-federation link in the description is about the advanced federation gateway in the Apollo Studio product that manages the services with a registry instead of having to list them manually before startup of the gateway.

t1:
I found that if we really want to support federation with an extension (i.e. a library you can include, not something built directly into the server implementation), we not only need custom directives (#628 for `@external`, `@key`, `@provides`, `@requires`, and `@extends`), we also need:

* Unions (#400 for `_Entity`); this even has to be dynamic, as the union is the list of all types with a `@key` directive.
[...]

PR #965, opened 2021-08-08
WIP: Integrate Federation Into Schema Builder

shmulik-klein:
Initial PR for #521

### DIFF
--- a/common/schema-builder/pom.xml
+++ b/common/schema-builder/pom.xml
@@ -48,11 +48,6 @@
             <artifactId>smallrye-graphql-api</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>io.smallrye</groupId>
-            <artifactId>smallrye-graphql-federation-api</artifactId>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 
     <build>
--- a/common/schema-builder/pom.xml
+++ b/common/schema-builder/pom.xml
@@ -48,6 +48,11 @@
             <artifactId>smallrye-graphql-api</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye</groupId>
+            <artifactId>smallrye-graphql-federation-api</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
+++ b/common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
@@ -134,7 +134,7 @@ public class SchemaBuilderTest {
 
         final Schema schema = SchemaBuilder.build(index);
 
-        Assertions.assertFalse(schema.isFederated());
+        assertFalse(schema.isFederated());
     }
 
     /**
@@ -148,7 +148,7 @@ public class SchemaBuilderTest {
 
         final Schema schema = SchemaBuilder.build(index);
 
-        Assertions.assertTrue(schema.isFederated());
+        assertTrue(schema.isFederated());
     }
 
     @Test
--- a/common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
+++ b/common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
@@ -133,6 +133,43 @@ public class SchemaBuilderTest {
         }
     }
 
+    /**
+     * Test a schema which includes a federated directives. The model should indicate that the schema isn't federated.
+     */
+    @Test
+    public void testSchemaWithoutFederatedDirectives() {
+        try {
+            Indexer indexer = new Indexer();
+            indexDirectory(indexer, "io/smallrye/graphql/index/app");
+            IndexView index = indexer.complete();
+
+            final Schema schema = SchemaBuilder.build(index);
+
+            Assertions.assertFalse(schema.isFederated());
+        } catch (SchemaBuilderException e) {
+            // ok
+        }
+    }
+
+
+    /**
+     * Test a schema which includes a federated directives. The model should indicate that the schema is federated.
+     */
+    @Test
+    public void testSchemaWithFederatedDirectives() {
+        try {
+            Indexer indexer = new Indexer();
+            indexDirectory(indexer, "io/smallrye/graphql/index/federated");
+            IndexView index = indexer.complete();
+
+            final Schema schema = SchemaBuilder.build(index);
+
+            Assertions.assertTrue(schema.isFederated());
+        } catch (SchemaBuilderException e) {
+            // ok
+        }
+    }
+
     @Test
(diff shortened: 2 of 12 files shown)


common/schema-builder/pom.xml
```diff
@@ -48,6 +48,11 @@
             <artifactId>smallrye-graphql-api</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye</groupId>
+            <artifactId>smallrye-graphql-federation-api</artifactId>
+            <scope>test</scope>
+        </dependency>
```
t1, line 55: I think we should move all the federation annotations directly into the api module, maybe in a separate package. @phillip-kruger: what do you think?

common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
```diff
@@ -133,6 +133,43 @@ public void testSchemaWithDuplicates() {
...
+            Indexer indexer = new Indexer();
+            indexDirectory(indexer, "io/smallrye/graphql/index/app");
+            IndexView index = indexer.complete();
+
+            final Schema schema = SchemaBuilder.build(index);
+
+            Assertions.assertFalse(schema.isFederated());
+        } catch (SchemaBuilderException e) {
+            // ok
+        }
```
t1, line 151: why is this okay?

common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
```diff
@@ -133,6 +133,43 @@ public void testSchemaWithDuplicates() {
...
+    @Test
+    public void testSchemaWithoutFederatedDirectives() {
+        try {
+            Indexer indexer = new Indexer();
+            indexDirectory(indexer, "io/smallrye/graphql/index/app");
+            IndexView index = indexer.complete();
+
+            final Schema schema = SchemaBuilder.build(index);
+
+            Assertions.assertFalse(schema.isFederated());
```
t1, line 148: We generally use a static import for the Assertions.

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Schema.java
```diff
@@ -33,7 +34,13 @@
...
     private Map<String, ErrorInfo> errors = new HashMap<>();
 
-    public Schema() {
+
+    /**
+     * Constructor
+     * @param isFederated indicates if the schema is a federated schema
+     */
+    public Schema(boolean isFederated) {
+        this.isFederated = isFederated;
```
t1, line 43: All the other schema details are set with setters, not with a constructor parameter. Just for uniformity, we should stick to that, or is there a good reason to change it?

common/schema-builder/pom.xml
phillip-kruger: Yes this should move to the api module

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Schema.java
phillip-kruger: Yes for Quarkus we need a public constructor and getters and setters

common/schema-builder/pom.xml
shmulik-klein: Copied into the seperated `server/api/src/main/java/io/smallrye/graphql/api/federation` package. 
Please let me know if you want it to be in the root package together with the core Graph QL annotations.

t1:
Sorry for not coming back to this earlier, I was on vacation. How far do you think you are? I'm just exploring the code a bit.

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Schema.java
```diff
@@ -33,6 +34,10 @@
 
     private Map<String, ErrorInfo> errors = new HashMap<>();
 
+    /**
+     * Constructor
+     *
+     */
```
t1, line 40: don't comment the obvious ;-)
```suggestion
```

shmulik-klein:
> Sorry for not coming back to this earlier, I was on vacation. How far do you think you are? I'm just exploring the code a bit.

Sorry for the late reply.

Not too far, I was trying to replace the event listening on `io/smallrye/graphql/federation/impl/Federation.java:111` into an explicit call to generate the federated schema before the schema is build on `io/smallrye/graphql/bootstrap/Bootstrap.java:174`.

t1:
@shmulik-klein: the federation topic is getting more traction in my current project. Can you find some time to work on this in the near future? I completely understand how limited time is.

robp94:
@t1 I added another pr at https://github.com/t1/graphql-federation-demo/pull/6.

There is still a problem left.
On start up, there is a crash at https://github.com/smallrye/smallrye-graphql/blob/12ea1257a46c68c90851b59eaa53b92ce22340b2/server/federation/runtime/src/main/java/io/smallrye/graphql/federation/impl/Federation.java#L144
The problem is the line above
https://github.com/smallrye/smallrye-graphql/blob/12ea1257a46c68c90851b59eaa53b92ce22340b2/server/federation/runtime/src/main/java/io/smallrye/graphql/federation/impl/Federation.java#L143
Without it, the war demo does work again.
Do we need this for something, or could we remove it?

For Quarkus there is a second problem (maybe more). The Federation.class needs the jandex index. The index is created in the quarkus extension
https://github.com/quarkusio/quarkus/blob/e24742a2f3da1572e7529dfd67b36f90abc7f704/extensions/smallrye-graphql/deployment/src/main/java/io/quarkus/smallrye/graphql/deployment/SmallRyeGraphQLProcessor.java#L209
 and stored in the ScanningContext  But when it is needed it is not there anymore since we have a different current. Do you maybe have an idea how to fix this, or who could have an idea?

johgoe:
@ robp94 Just an idea: 
if https://github.com/quarkusio/quarkus/issues/20802 is fixed

a configuration like 
```
quarkus.smallrye-graphql.events.enabled=true
quarkus.smallrye-graphql.schema-include-scalars=true
quarkus.index-dependency.graphql-federation-runtime.group-id=io.smallrye
quarkus.index-dependency.graphql-federation-runtime.artifact-id=smallrye-graphql-federation-api
quarkus.index-dependency.graphql-federation-api.group-id=io.smallrye
quarkus.index-dependency.graphql-federation-api.artifact-id=smallrye-graphql-federation-api
```

and maven dependencies like
```
      <dependency>
        <artifactId>smallrye-graphql-federation-api</artifactId>
        <version>1.3.4</version>
      </dependency>
      <dependency>
        <artifactId>smallrye-graphql-federation-runtime</artifactId>
        <version>1.3.4</version>
      </dependency>
```
should fix the quarkus problems on startup.

phillip-kruger:
@t1 - this looks more like what we need to do - update the model to contain info on federation, and use that info in Bootstrap

robp94:
@jmartisk @phillip-kruger since federation is now supported I think this could be closed.

Issue #994, opened 2021-08-23
Graphql with authentication


AnakinPt:
I need to connect to a service requiring authentication.
First time I connect to it, I must use the User and password to connect and I'm fine to connect. But, in the service answer, they add a cookie, jwt, to the response that I must use in the following requests.

I would like to have a way to read these cookies from the response so I can act on this.

I create the client this way:
```
        UsernamePasswordCredentials creds = new UsernamePasswordCredentials(username, password);
        Header authenticate = new BasicScheme().authenticate(creds, new HttpPost(url), null);


        DynamicGraphQLClient dynamicGraphQLClient = DynamicGraphQLClientBuilder.newBuilder()
                .url(url)
                .header(authenticate.getName(), authenticate.getValue())
                .build();
        this.dynamicGraphQLClient = dynamicGraphQLClient;
```

and I fetch the data this way:
```
Response response = getDynamicGraphQLClient().executeSync(query);
```

I would like to do some magic, after the first request, so I could recreate the client using an additional header.

phillip-kruger:
@jmartisk

AnakinPt:
I used a big hammer for my tests and seems to work, but I only applied it for the code I really need.

In class VertxDynamicGraphQLClient I rewrote the private method executeSync to this:
```
    private Response executeSync(Buffer buffer) throws ExecutionException, InterruptedException {
        HttpResponse<Buffer> result = webClient.postAbs(url)
                .putHeaders(headers)
                .sendBuffer(buffer)
                .toCompletionStage()
                .toCompletableFuture()
                .get();
        MyResponseImpl response = readFrom(result.body());
        response.setHeaders(result.headers().entries());
        return response;
    }
```

And I created a new class MyResponseImpl like this:
```
@ToString
public class MyResponseImpl  extends ResponseImpl implements Response {

    @Setter
    @Getter
    private List<Map.Entry<String, String>> headers;

    public MyResponseImpl(JsonObject data, List<Error> errors) {
        super(data, errors);
    }

}

```

This was just an hack, but seems to serve my goal. I don't know if we can apply a similar solution for all methods.

AnakinPt:
I just opened a PR with the change I propose.

PR #997, opened 2021-08-25
#994 add headers to the Response in the VertxDynamicGraphQLClient

AnakinPt:
This is a simplified way for the Vetx ResponseImpl to also have the received headers so we can use them for Authentication

### DIFF
--- a/client/api/src/main/java/io/smallrye/graphql/client/Response.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/Response.java
@@ -1,9 +1,10 @@
 package io.smallrye.graphql.client;
 
 import java.util.List;
-import java.util.Map;
 
 import javax.json.JsonObject;
+
+import io.vertx.core.MultiMap;
 
 public interface Response {
 
@@ -19,5 +20,5 @@ public interface Response {
 
     boolean hasError();
 
-    List<Map.Entry<String, String>> getHeaders();
+    MultiMap getHeaders();
 }
--- a/client/api/src/main/java/io/smallrye/graphql/client/Response.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/Response.java
@@ -1,6 +1,7 @@
 package io.smallrye.graphql.client;
 
 import java.util.List;
+import java.util.Map;
 
 import javax.json.JsonObject;
 
@@ -17,4 +18,6 @@ public interface Response {
     boolean hasData();
 
     boolean hasError();
+
+    List<Map.Entry<String, String>> getHeaders();
 }
--- a/client/api/src/main/java/io/smallrye/graphql/client/Response.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/Response.java
@@ -1,10 +1,9 @@
 package io.smallrye.graphql.client;
 
 import java.util.List;
+import java.util.Map;
 
 import javax.json.JsonObject;
-
-import io.vertx.core.MultiMap;
 
 public interface Response {
 
@@ -20,5 +19,5 @@ public interface Response {
 
     boolean hasError();
 
-    MultiMap getHeaders();
+    List<Map.Entry<String, String>> getHeaders();
 }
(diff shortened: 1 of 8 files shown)


client/api/src/main/java/io/smallrye/graphql/client/Response.java
```diff
@@ -17,4 +18,6 @@
     boolean hasData();
 
     boolean hasError();
+
+    List<Map.Entry<String, String>> getHeaders();
```
jmartisk, line 22: Why not simply `Map<String, String>`?  This is unnecessarily complicated.

client/implementation/src/main/java/io/smallrye/graphql/client/dynamic/ResponseImpl.java
```diff
@@ -15,10 +16,12 @@
 
     private final JsonObject data;
     private final List<Error> errors;
+    private List<Map.Entry<String, String>> headers;
```
jmartisk, line 19: This can be `final` (and a `Map<String, String>`)

client/implementation/src/main/java/io/smallrye/graphql/client/dynamic/ResponseImpl.java
```diff
@@ -63,4 +66,7 @@ public String toString() {
         return "GraphQLResponse{" + "data=" + data + ", errors=" + errors + '}';
     }
 
+    public List<Map.Entry<String, String>> getHeaders(){
+        return new ArrayList<>(headers);
```
jmartisk, line 70: Maybe I'd prefer wrapping the headers into an unmodifiable collection in the constructor (they're not supposed to change after the constructor finishes), and here returning that already unmodifiable collection, instead of allocating a new object.. WDYT?

jmartisk (changes requested):
Looks nice! I added comments for just a few little things before accepting.
Also CI is complaining that the source files are not properly formatted (please run a Maven build before committing), and it also needs a rebase.
Thanks!

client/api/src/main/java/io/smallrye/graphql/client/Response.java
t1: Headers can be repeated, so a `Map` won't work. In JAX-RS there's `MultivaluedMap` which extends `Map<K, List<V>>`. In vert.x there's `MultiMap` which implements `Iterable<Entry<String, String>>`.

jmartisk: Hmm okay, I didn't realise that. We certainly want to avoid depending on JAX-RS or Vert.x classes, so...  `List<Map.Entry<String, String>>` makes retrieving a header by its name somewhat complicated, so maybe we should automatically transform the headers to a `Map<K, List<V>>`, where it is more straightforward?

AnakinPt: Rebase done.
I changed to HeadersMultiMap as suggested, I'm also filling the MultiMap in the constructor.

client/implementation/src/main/java/io/smallrye/graphql/client/dynamic/ResponseImpl.java
AnakinPt: Changed

client/implementation/src/main/java/io/smallrye/graphql/client/dynamic/ResponseImpl.java
AnakinPt: Changed the constructor code as suggested.

client/api/src/main/java/io/smallrye/graphql/client/Response.java
jmartisk: @AnakinPt Ah, no no, I meant we want to avoid depending on Vert.x in the common module, you've just added that dependency. `Response` should store a `Map<String, List<String>>`, not a `MultiMap`.

AnakinPt: Sorry. I misunderstood. Later on today I'll revert and do the changes proposed.

AnakinPt: Changes done. I think now we can work with this.

jmartisk (dismissed):

jmartisk:
Thanks!

PR #1006, opened 2021-08-29
fix #1005: map JsonNumber to BigDecimal for Objects

### LINKED ISSUE #1005: Typesafe Client: exception while reading array path index
It throws, e.g., `GraphQlClientException: invalid java.lang.Object value for test.graphql.OrderIT$Api#order.items[0].product.path[2]: 0`

t1:
so error paths containing an array index can be read

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/json/JsonNumberReader.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/json/JsonNumberReader.java
@@ -41,7 +41,7 @@ class JsonNumberReader extends Reader<JsonNumber> {
             return value.doubleValue();
         if (BigInteger.class.equals(rawType))
             return value.bigIntegerValueExact();
-        if (BigDecimal.class.equals(rawType))
+        if (BigDecimal.class.equals(rawType) || Object.class.equals(rawType))
             return value.bigDecimalValue();
 
         throw new GraphQLClientValueException(location, value);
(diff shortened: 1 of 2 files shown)


client/implementation/src/main/java/io/smallrye/graphql/client/typesafe/impl/json/JsonNumberReader.java
```diff
@@ -41,7 +41,7 @@ private Object read(Location location, JsonNumber value, Class<?> rawType) {
             return value.doubleValue();
         if (BigInteger.class.equals(rawType))
             return value.bigIntegerValueExact();
-        if (BigDecimal.class.equals(rawType))
+        if (BigDecimal.class.equals(rawType) || Object.class.equals(rawType))
```
phillip-kruger, line 44: Not sure if I understand this correct, but it does not look right that Object maps to BigDecimal ??

t1: It's the other way around: the `path` can contain Strings (object field names) and Numbers (array indexes), so it's a `List<Object>`. When we actually receive a number, we could map it to an `Integer`, but as this code could be used by other mappings as well, it's safer to map it to a `BigDecimal` without loosing any precision, etc. As we don't access this number anywhere, only client code looking into the path would be able to tell the difference, and any existing code would have run into an exception before this fix. So it's all safe :-)

phillip-kruger: ok thanks

phillip-kruger (approved):

PR #1096, opened 2021-10-14
Start work on wildcard type support.

nenros:
I added simply try to handle wildcard type. But i'm not able to write unit test for that. I tried to write one but it pass without my code so probably it wasn't best.

I would like to have wildcard support to make kotlin coroutines work in the future.

So i would be gratefull for help and comments ;)


nenros:
So I need some help with working test, that would really test something.

phillip-kruger:
@nenros - Thanks for this. When do you get the error mentioned above ?

phillip-kruger:
What happened here ?

nenros:
Maybe i will try write once again, i'm not sure is that PR still valid after all changes.

Issue #1109, opened 2021-10-18
Kotlin support for "?" operator for non null and nullable types 


robp94:
In kotlin all types are non-nullable except you mark them with a `?`. So different to Java, there would be no need for the annotation `NonNull` as kotlin itself has this information.

Quick example:

```
fun nullable(): String?{}

fun nunNull(): String {}
```

Could this be supported?

robp94:
Hm, this would actually be pretty simple as soon as Jandex 3 is used. Kotlin already places Annotations for the `?`. But jandex currently can't read them. This is resolved in https://github.com/wildfly/jandex/issues/120.

So we would only need to add this annotation: `org.jetbrains.annotations.NotNull`

This propably would be a breaking change, so maybe it should be behind a property? @phillip-kruger

robp94:
Ok I will make a PR later.

PR #1113, opened 2021-10-19
add support for kotlin non-null annotations

robp94:
This will fix https://github.com/smallrye/smallrye-graphql/issues/1109 after jandex 3 is used. At the moment it should do nothing. Since the test can not work, it is not active.


phillip-kruger:
@robp94 - please build locally and commit and push again

robp94:
> 
> 
> @robp94 - please build locally and commit and push again

@phillip-kruger done

PR #1132, opened 2021-10-26
#521: add federation to core:

t1:
* move federation api into smallrye-graphql-api
* add config to enable federation
* add dependency on com.apollographql.federation:federation-graphql-java-support
* transform federation schema
* remove old federation module

- [x] implement `Bootstrap#fetchEntities` and `Bootstrap#fetchEntityType`
- [x] add tests
- [x] properly resolve the class name
- [x] update README.


t1:
@phillip-kruger & @johgoe : I think the core functionality should be working. I have some trouble to get the `ApolloGateway` to work properly. It seem that it doesn't recognize the `@key(fields : [\"id\"])` directive, but AFAIU this should be legal. IIRC I had to work around this in the former, hand-made solution. Could you help me double-check?

t1:
It works after updating Apollo to the latest version!!! So it _was_ a bug.

I shared my demo project: https://github.com/t1/smallrye-graphql-federation-demo

I'll continue writing tests now.

Comments welcome!!!!!!!1111

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -183,7 +186,30 @@ private void generateGraphQLSchema() {
...
+        }
+    }
+
+    private TypeResolver fetchEntityType() {
+        return env -> {
+            Object src = env.getObject();
+            if (src == null) {
+                return null;
+            }
+            GraphQLObjectType result = env.getSchema().getObjectType(src.getClass().getSimpleName()); // TODO respect @Name, etc.
```
johgoe, line 207: To allow the use of @Name

```
Class<?> clazz = src.getClass();
String typeName = clazz.getSimpleName();
Name annotation = clazz.getAnnotation(Name.class);
if (annotation!=null) {
  typeName = annotation.value();
}
GraphQLObjectType result = env.getSchema().getObjectType(typeName);
```

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/FederationDataFetcher.java
```diff
@@ -0,0 +1,91 @@
...
+            if (matchesReturnType(field, typename) && matchesArguments(requestedArgs, field)) {
+                return execute(field, env, requestedArgs);
+            }
+        }
+        throw new RuntimeException("no query found for " + typename + " by " + requestedArgs.keySet());
+    }
+
+    private boolean matchesReturnType(GraphQLFieldDefinition field, String typename) {
+        GraphQLOutputType returnType = field.getType();
+        return returnType instanceof GraphQLObjectType && ((GraphQLObjectType) returnType).getName().equals(typename);
```
johgoe, line 53: In our current schema we are also use GraphQLInterfaceType for Federation return.

I'm not sure but I think this could work:

```
return (returnType instanceof GraphQLObjectType && ((GraphQLObjectType) returnType).getName().equals(typename)) || 
(returnType instanceof GraphQLInterfaceType && ((GraphQLInterfaceType) returnType).getName().equals(typename))
```

phillip-kruger:
Thanks @t1. I'll have a look a.s.a.p

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
t1: I think we'll have to use Jandex to find the annotation.

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/FederationDataFetcher.java
t1: I just pushed a change using `GraphQLNamedSchemaElement` instead, which covers `GraphQLInterfaceType` and in the future Unions, etc.

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
t1: @phillip-kruger : can you give us a hint on how to resolve this best? I already searched for some helper method.

t1:
Is the documentation sufficient?

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
robp94: Shouldn't something like this work
`
ScanningContext.getIndex().getClassByName(typeName)
...
`

robp94:
Looks nice. Does it support mutiny return types? Would this already work with quarkus? Then we could test it with our services and see if we find anything.

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
t1: The `Bootstrap` class is in `server/implementation` which doesn't have a dependency on `common/schema-builder`, so I can't access the `ScanningContext`. There's also no dependency on Jandex, so I think we should go for the minimalistic solution  that @johgoe has suggested... for now. If a better solution comes up, we can do that later.

t1: Done. Thanks, all!

t1:
> Does it support mutiny return types?

It probably should, but I haven't tried.

> Would this already work with quarkus?

The quarkus tests fail with compile errors that I don't recognise immediately.

t1:
I think this is complete now and could be merged.

phillip-kruger:
Hi @t1 - Thanks for this. I'll have a look as soon as I have a gap

t1:
@jmartisk: it looks like @phillip-kruger is extremely busy with other things right now. maybe you can take give some feedback?

phillip-kruger:
I'll look at this soon. This will need a thorough review and we need to see if this will work in Quarkus

jmartisk:
This could also use writing some Quarkus-side integration test coverage as part of doing the review (well, if possible - because running a real integration test seems to require a gateway and at least two Quarkus instances, which makes it very difficult)
If you prefer I can get to it, but I'm quite unfamiliar with Federation so it will take me quite some time

t1:
You're right, writing a real integration test would mean running a node app. But there's already a simplified integration test in the tck package, which only checks the result of a single federated request. This, and the Federation specific directives, should be good enough for now. And maybe we will get back to [Feder](https://github.com/graphql-feder/feder) soon, so we could add a full integration test and still stay 100% in the Java universe.

jmartisk:
I'd suggest rebasing, as there is a conflict anyway ;) and pushing again

server/implementation/pom.xml
```diff
@@ -43,26 +46,30 @@
...
         <!-- GraphQL Java -->
         <dependency>
             <groupId>com.graphql-java</groupId>
             <artifactId>graphql-java</artifactId>
         </dependency>
-        
+        <dependency>
+            <groupId>com.apollographql.federation</groupId>
+            <artifactId>federation-graphql-java-support</artifactId>
+            <version>2.0.6</version>
```
phillip-kruger, line 58: Maybe move the version to parent with `graphql-java` ?

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -184,7 +189,32 @@ private void generateGraphQLSchema() {
...
+            Object src = env.getObject();
+            if (src == null) {
+                return null;
+            }
+            Name annotation = src.getClass().getAnnotation(Name.class);
+            String typeName = (annotation == null) ? src.getClass().getSimpleName() : annotation.value();
+            GraphQLObjectType result = env.getSchema().getObjectType(typeName);
+            if (result == null) {
+                throw new RuntimeException("can't resolve federated entity type " + src.getClass().getName());
+            }
```
phillip-kruger, line 215: This needs to happen with Jandex, when building the model, so the model might have to be enhanced. Bootstrap (and any other class used in runtime) will then use the model.

server/runner/src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -2,7 +2,7 @@ smallrye.graphql.printDataFetcherException=true
 smallrye.graphql.tracing.enabled=true
 smallrye.graphql.allowGet=true
 smallrye.graphql.logPayload=true
-
 mp.graphql.showErrorMessage=java.security.AccessControlException,io.smallrye.graphql.test.apps.exceptionlist.*
 mp.graphql.hideErrorMessage=java.io.IOException,io.smallrye.graphql.test.apps.exceptionlist.*
-smallrye.graphql.errorExtensionFields=exception,classification,code,description,validationErrorType,queryPath
\ No newline at end of file
+smallrye.graphql.errorExtensionFields=exception,classification,code,description,validationErrorType,queryPath
+smallrye.graphql.federation.enabled=true
```
phillip-kruger, line 8: Maybe we should make the default `true` ? I might have said the opposite in the past, but looking at this now, I think this can be enabled by default ? We not breaking anything with the current usage, right ?

phillip-kruger (changes requested):
LGTM ! Thanks @t1 - this is a big contribution !  If we can move that one part to the model, I think we are good to go !

server/implementation/pom.xml
t1: done

server/runner/src/main/resources/META-INF/microprofile-config.properties
t1: not that I'm aware of. will do

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
phillip-kruger: If we do the same in InterfaceResolver and UnionResolver, we need to fix those too. I'll check out your PR branch and see if this is going to be possible. We might not be able to since we are using a library. I'll get back to you a.s.a.p

t1: Do you consider this PR as incomplete or rather just imperfect? If the latter, we could, maybe, merge this as it is and fix these shortcomings in a separate PR.

phillip-kruger: I think it's complete, we can merge and then follow up, but just give me some time, I'll merge either way later this week. Is that ok ?

t1: sure!

phillip-kruger:
@jmartisk - I am merging here. Can you help with creating a new 1.9.x branch and backport (non-Jakarata) to there ? Then I cal look at the jandex in that branch and once we are happy we can get this into Quarkus. Le me know

phillip-kruger:
@t1 Thanks for this massive contribution !

jmartisk:
I'll create a 1.9.x branch with this that will target Quarkus 2.15, which is due some time in December or January, so we will have enough time to polish it. At minimum, I believe we should make the new dependency (`federation-graphql-java-support`) optional so it doesn't have to be included with every application even if it doesn't use Federation.

Issue #923, opened 2021-07-20
`@NonNull` at class/package level


ybroeker:
It would be nice, if `@NonNull` (or an alternative annotation) could be used at class or package-level to mark all fields as non null.

Most of our classes look like this: 

```java
class API {
  @Query @NonNull String someField() {...}

  @Query Optional<String> someOptionalField() {...}

  @Query @NonNull List<@NonNull String> someFields() {...}

  @Mutation @NonNull SomePayload someFields(@NonNull SomeInput input) {...}
}
```

which could look like this if `@NonNull` on classes would be possible, where anything except `someOptionalField` would be non-null:

```java
@NonNull
class API {
  @Query String someField() {...}

  @Query Optional<String> someOptionalField() {...}

  @Query List<String> someFields() {...}

  @Mutation SomePayload someFields(SomeInput input) {...}
}
```

----

I don't know if `@NonNull` is the best approach for this, maybe a new annotation to specify default nullability for fields, inputs, parameters would be better. 

Maybe someone has an idea?

PR #1133, opened 2021-10-27
Draft: Add annotations to mark classes/packages as non-null and fields as nullable

ybroeker:
Just a first draft, would like some feedback.
`@AllNonNull` on a class or package marks all fields/parameters in that class/package as non-null.
Using `Optional` or annotating the field with `@Nullable` makes this field nullable again.

Currently this would apply to all fields and parameters, which may not be appropriate.
Also, it might be handy to make exceptions for wrapper types. `Integer` would then always be nullable, `int` would be the alternative for non-null.

----

Code like this:

```java
@AllNonNull
public class NonNullClass {

    public String nonNullString = "";

    public Optional<String> optionalString = Optional.empty();

    @Nullable
    public String nullableString = null;

}
```

would then result in a schema like:

```graphql
type NonNullClass {
  nonNullString: String!
  optionalString: String
  nullableString: String
}
```

----

see #923


phillip-kruger (changes requested):
@ybroeker please rename AllNonNull to DefaultNonNull, then we merge

phillip-kruger (approved):
Thanks !

PR #1135, opened 2021-10-29
DRAFT: Additional features for graphql federation

johgoe:
This pull request should only publish some changes, which we did in our project to get some additional federation features. It's not tested yet because on quarkus we did some additional changes like the generation of an additional jandex index, which shouldn't be required for normal application servers like a wildfly.
Maybe our changes can be a inspiration for smallrye-graphql federation support.

- Add support for GraphQLInterfaces
- Allow primitive types as parameters in resolver methods
- Allow mutiny! returns


phillip-kruger:
@t1 - can you look at this ?

t1:
This is based on the old approach which was completely different to what we did in #521, which ist based on `federation-graphql-java-support`. I haven't tested it, but the requirements for this ticket may very well already have been addressed. I think we can close this. WDYT, @johgoe?

johgoe:
I also think this PR is obsolete. I think the new approach already covers the most of these changes.

PR #1148, opened 2021-11-05
Support default values for a collection/array of input types

tkareine:
Add support for specifying default values for a collection or an array of custom input types.

Especially, being able to specify a field to have the empty list/array as the default is useful, because then we can avoid null checks. For example:

``` java
@GraphQLApi
public class API {
    @Query
    public String firstNestedFieldOrNull(@NonNull Root input) {
        // no null check needed in accessing input.nestedList
        return input.nestedList.stream().findFirst().map(n -> n.field).orElse(null);
    }

    @Input
    public static class Root {
        @DefaultValue("[]")
        public List<Nested> nestedList;
    }

    @Input
    public static class Nested {
        public String field;
    }
}
```


phillip-kruger (approved):
Thanks for your contribution !

PR #1199, opened 2021-12-02
Map as operation input

phillip-kruger:
@jmartisk f.y.i


jmartisk:
Worked perfectly with my tests that I had prepared! Thanks!

PR #1222, opened 2021-12-17
add setting headers in the TypesafeGraphQLClientBuilder

t1:
and improve mixing config (MP Config + annotations) for typesafe client

### DIFF
--- a/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientBuilder.java
+++ b/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientBuilder.java
@@ -4,7 +4,8 @@ import java.lang.reflect.Proxy;
 import java.net.URI;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
 
 import io.smallrye.graphql.client.impl.ErrorMessageProvider;
 import io.smallrye.graphql.client.impl.GraphQLClientConfiguration;
@@ -23,6 +24,7 @@ public class VertxTypesafeGraphQLClientBuilder implements TypesafeGraphQLClientB
 
     private String configKey = null;
     private URI endpoint;
+    private Map<String, String> headers;
     private Vertx vertx;
     private HttpClientOptions options;
     private WebClient webClient;
@@ -44,6 +46,7 @@ public class VertxTypesafeGraphQLClientBuilder implements TypesafeGraphQLClientB
         return this;
     }
 
+    @SuppressWarnings("unused")
     public TypesafeGraphQLClientBuilder options(HttpClientOptions options) {
         this.options = options;
         return this;
@@ -56,36 +59,44 @@ public class VertxTypesafeGraphQLClientBuilder implements TypesafeGraphQLClientB
     }
 
     @Override
+    public TypesafeGraphQLClientBuilder header(String name, String value) {
+        if (this.headers == null) {
+            this.headers = new LinkedHashMap<>();
+        }
+        this.headers.put(name, value);
+        return this;
+    }
+
+    @Override
     public <T> T build(Class<T> apiClass) {
         if (configKey == null) {
             configKey = configKey(apiClass);
         }
 
-        GraphQLClientsConfiguration configs = GraphQLClientsConfiguration.getInstance();
-        GraphQLClientConfiguration persistentConfig = configs
-                .getClient(configKey);
-        if (persistentConfig == null) {
-            // in case that we're running in a plain Java SE application, it is possible that the client configuration
-            // hasn't been added yet, because there is no CDI extension or Jandex processor that scans for @GraphQLClientApi annotations
-            // at startup => try adding a configuration entry dynamically for this client interface in particular.
-            // Then try again.
-            configs.addTypesafeClientApis(Collections.singletonList(apiClass));
-            persistentConfig = configs
-                    .getClient(configKey);
-        }
-        if (persistentConfig != null) {
-            applyConfig(persistentConfig);
-        }
+        applyConfigFor(apiClass);
 
         if (endpoint == null) {
             throw ErrorMessageProvider.get().urlMissingErrorForNamedClient(configKey);
         }
 
         initClients();
-        VertxTypesafeGraphQLClientProxy graphQlClient = new VertxTypesafeGraphQLClientProxy(persistentConfig,
-                endpoint, httpClient, webClient);
+        VertxTypesafeGraphQLClientProxy graphQlClient = new VertxTypesafeGraphQLClientProxy(headers, endpoint, httpClient,
+                webClient);
         return apiClass.cast(Proxy.newProxyInstance(getClassLoader(apiClass), new Class<?>[] { apiClass },
                 (proxy, method, args) -> invoke(apiClass, graphQlClient, method, args)));
+    }
+
+    private void applyConfigFor(Class<?> apiClass) {
+        GraphQLClientsConfiguration configs = GraphQLClientsConfiguration.getInstance();
+        // In case that we're running in a plain Java SE application, it is possible that the client configuration
+        // hasn't been added yet, because there is no CDI extension or Jandex processor that scans for @GraphQLClientApi annotations
+        // at startup => try adding a configuration entry dynamically for this client interface in particular.
+        configs.initTypesafeClientApi(apiClass);
+
+        GraphQLClientConfiguration persistentConfig = configs.getClient(configKey);
+        if (persistentConfig != null) {
+            applyConfig(persistentConfig);
+        }
     }
 
     private void initClients() {
@@ -140,6 +151,9 @@ public class VertxTypesafeGraphQLClientBuilder implements TypesafeGraphQLClientB
     private void applyConfig(GraphQLClientConfiguration configuration) {
         if (this.endpoint == null && configuration.getUrl() != null) {
             this.endpoint = URI.create(configuration.getUrl());
+        }
+        if (this.headers == null && configuration.getHeaders() != null) {
+            this.headers = configuration.getHeaders();
         }
     }
 
--- a/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
+++ b/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
@@ -6,7 +6,6 @@ import java.net.URI;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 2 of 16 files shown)


client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientBuilder.java
```diff
@@ -55,39 +58,47 @@ public TypesafeGraphQLClientBuilder endpoint(URI endpoint) {
...
         return apiClass.cast(Proxy.newProxyInstance(getClassLoader(apiClass), new Class<?>[] { apiClass },
                 (proxy, method, args) -> invoke(apiClass, graphQlClient, method, args)));
     }
 
+    private void applyConfigFor(Class<?> apiClass) {
+        GraphQLClientsConfiguration configs = GraphQLClientsConfiguration.getInstance();
+        // In case that we're running in a plain Java SE application, it is possible that the client configuration
+        // hasn't been added yet, because there is no CDI extension or Jandex processor that scans for @GraphQLClientApi annotations
+        // at startup => try adding a configuration entry dynamically for this client interface in particular.
+        configs.initTypesafeClientApi(apiClass);
```
t1, line 94: Yes, we do this always, so _all_ the config is being picked up, i.e. also credentials

client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
```diff
@@ -70,11 +72,8 @@ Object invoke(Class<?> api, MethodInvocation method) {
         if (method.isDeclaredInObject())
             return method.invoke(this);
 
-        MultiMap headers = new HeaderBuilder(api,
-                method,
-                configuration != null ? configuration.getHeaders() : Collections.emptyMap())
-                        .build();
-        headers.set("Accept", APPLICATION_JSON_UTF8);
```
t1, line 77: This has moved to HeaderBuilder

client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
```diff
@@ -230,15 +187,55 @@ private JsonObject objectValue(Object object, Stream<FieldInfo> fields) {
 
     private CompletionStage<HttpResponse<Buffer>> postAsync(String request, MultiMap headers) {
         return webClient.postAbs(endpoint.toString())
-                .putHeader("Content-Type", APPLICATION_JSON_UTF8)
```
t1, line 233: This has also moved to HeaderBuilder

client/implementation-vertx/src/test/java/test/tck/DynamicTckSuite.java
```diff
@@ -1,6 +0,0 @@
-package test.tck;
-
-import tck.graphql.dynamic.core.DynamicTCK;
-
-class DynamicTckSuite extends DynamicTCK {
-}
```
t1, line 6: this class was duplicated, so the dynamic tests ran twice

client/implementation/src/main/java/io/smallrye/graphql/client/impl/GraphQLClientsConfiguration.java
```diff
@@ -55,40 +60,32 @@ public GraphQLClientsConfiguration() {
...
-     * client configuration objects for them. This needs to be called by the runtime some time during initialization,
+     * Scan the passed Java interface for `@GraphQLClientApi` annotations and create and register
+     * client configuration objects for it. This needs to be called by the runtime some time during initialization,
      * before clients are actually created.
      */
-    public void addTypesafeClientApis(List<Class<?>> apis) {
-        if (apis == null) {
-            SmallRyeGraphQLClientLogging.log.apisNotSet();
-            apis = Collections.emptyList();
-        }
```
t1, line 73: This could never have happened

client/implementation/src/main/java/io/smallrye/graphql/client/impl/GraphQLClientsConfiguration.java
```diff
@@ -55,40 +60,32 @@ public GraphQLClientsConfiguration() {
...
     }
 
     /**
-     * Scan the passed Java interfaces for `@GraphQLClientApi` annotations and create and register
-     * client configuration objects for them. This needs to be called by the runtime some time during initialization,
+     * Scan the passed Java interface for `@GraphQLClientApi` annotations and create and register
+     * client configuration objects for it. This needs to be called by the runtime some time during initialization,
      * before clients are actually created.
      */
-    public void addTypesafeClientApis(List<Class<?>> apis) {
```
jmartisk, line 69: We have to keep this method. Quarkus calls it. I can't compile Quarkus without it
```
[ERROR] /home/jmartisk/Workspace/quarkus/extensions/smallrye-graphql-client/runtime/src/main/java/io/quarkus/smallrye/graphql/client/runtime/SmallRyeGraphQLClientRecorder.java:[34,19] cannot find symbol
[ERROR]   symbol:   method addTypesafeClientApis(java.util.List<java.lang.Class<?>>)
[ERROR]   location: variable configBean of type io.smallrye.graphql.client.impl.GraphQLClientsConfiguration
```

I would be fine with having both, where `addTypesafeClientApis` calls `initTypesafeClientApi` in a loop

client/implementation/src/main/java/io/smallrye/graphql/client/impl/GraphQLClientsConfiguration.java
jmartisk: Not with pure SmallRye code, but this method is called by Quarkus so I think this should stay guarded

client/implementation/src/main/java/io/smallrye/graphql/client/impl/GraphQLClientsConfiguration.java
```diff
@@ -55,40 +60,32 @@ public GraphQLClientsConfiguration() {
...
+    public void initTypesafeClientApi(Class<?> api) {
+        TypesafeClientConfigurationReader reader = new TypesafeClientConfigurationReader(api);
+        clients.put(reader.getConfigKey(), reader.getClientConfiguration());
     }
 
     public GraphQLClientConfiguration getClient(String key) {
         return clients.get(key);
     }
 
-    public void addClient(String key, GraphQLClientConfiguration config) {
```
jmartisk, line 84: We have to keep this, this is called by Quarkus
```
[ERROR] /home/jmartisk/Workspace/quarkus/extensions/smallrye-graphql-client/runtime/src/main/java/io/quarkus/smallrye/graphql/client/runtime/GraphQLClientConfigurationMergerBean.java:[50,38] cannot find symbol
[ERROR]   symbol:   method addClient(java.lang.String,io.smallrye.graphql.client.impl.GraphQLClientConfiguration)
[ERROR]   location: variable upstreamConfiguration of type io.smallrye.graphql.client.impl.GraphQLClientsConfiguration
```

jmartisk (changes requested):

jmartisk:
Through https://github.com/smallrye/smallrye-graphql/pull/1228 I've just added some Quarkus-based tests to the CI, so once you fix the Quarkus compilation issues and rebase+push the PR, the smallrye CI should be able to verify it

t1:
@jmartisk: I reverted the API changes, but now some Quarkus tests fail. I reproduced them locally, but don't understand what's happening. Can you help?

client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientBuilder.java
jmartisk: If you do this repeatedly (without checking that the config is not initialized yet) you will overwrite the customization that Quarkus performs on the configuration class - using `io.quarkus.smallrye.graphql.client.runtime.GraphQLClientConfigurationMergerBean` which merges config added through  native `quarkus.*` properties into the configuration class, so configuration using `quarkus.*` properties will stop working - this is why you're getting errors in the Quarkus tests.

It seems fixed if I do this:
```
private void applyConfigFor(Class<?> apiClass) {
        GraphQLClientsConfiguration configs = GraphQLClientsConfiguration.getInstance();

        GraphQLClientConfiguration persistentConfig = configs.getClient(configKey);
        if (persistentConfig == null) {
            // In case that we're running in a plain Java SE application, it is possible that the client configuration
            // hasn't been added yet, because there is no CDI extension or Jandex processor that scans for @GraphQLClientApi annotations
            // at startup => try adding a configuration entry dynamically for this client interface in particular.
            configs.initTypesafeClientApi(apiClass);
            persistentConfig = configs.getClient(configKey);
        }
        if (persistentConfig != null) {
            applyConfig(persistentConfig);
        }
    }
```

I'm not sure what you mean by *all* config not getting picked up, do you see any problem with this?
SmallRye should generally call `initTypesafeClientApi` only in a plain Java SE app - in Quarkus, this is called by Quarkus, and if you call it again, you overwrite what Quarkus enhanced.

jmartisk: Oh, I think I know what you mean now - the tests in `ConfigBehavior` change the properties-based configuration over time, so you need to rescan the config for the same typesafe client class. That's a bit problematic because we generally assume the configuration to be immutable and not expected to change, at least Quarkus always treats it as such.

jmartisk: Would it be too much of a hassle for you to rewrite the tests so that you don't change the configuration for one typesafe client over time?
Especially since it's a TCK, and therefore it's meant to be reused by different implementations in the future, I think we should not assume that the implementation will consider the config mutable, especially in the era of microservice toolkits like Quarkus and Micronaut.

t1: I get your point. I don't want to mutate existing config, but merge different config "sources", i.e. annotations and config properties. I'll have to think about how to solve this properly. Thanks for investigating!

t1: now it works... WDYT?

client/implementation/src/main/java/io/smallrye/graphql/client/impl/GraphQLClientsConfiguration.java
t1: done

client/implementation/src/main/java/io/smallrye/graphql/client/impl/GraphQLClientsConfiguration.java
t1: done

client/implementation/src/main/java/io/smallrye/graphql/client/impl/GraphQLClientsConfiguration.java
t1: done

client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientBuilder.java
jmartisk: Ok this will do

jmartisk (approved):

jmartisk:
Thanks!

t1:
Cool, thank you! Are there plans for a next release?

jmartisk:
I think we can do one this week.

PR #1224, opened 2021-12-18
Add default getters for the well-known extension fields to GraphQLError

t1:
These are the fields that can be configured with `smallrye.graphql.errorExtensionFields`


phillip-kruger (approved):
LGTM

PR #1243, opened 2022-01-10
Use annotationType instead of getClass()

charlescrookes:
Access type of annotation directly, not through the getClass() that creates a proxy. This allows for the library to be used seamlessly in GraalVM Native applications.


jmartisk (approved):

jmartisk:
Thanks!

PR #1267, opened 2022-01-29
#522: client considers `@Ignore`

t1:
this is not a fix for the full issue, but a small change to even consider the `@Ignore` annotation on a field.

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/QueryBuilder.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/QueryBuilder.java
@@ -88,7 +88,6 @@ public class QueryBuilder {
             return "{ key " + keyFields + " value " + valueFields + "}";
         }
         return type.fields()
-                .filter(FieldInfo::isOutput)
                 .map(this::field)
                 .collect(joining(" ", " {", "}"));
     }
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/QueryBuilder.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/QueryBuilder.java
@@ -88,6 +88,7 @@ public class QueryBuilder {
             return "{ key " + keyFields + " value " + valueFields + "}";
         }
         return type.fields()
+                .filter(FieldInfo::isOutput)
                 .map(this::field)
                 .collect(joining(" ", " {", "}"));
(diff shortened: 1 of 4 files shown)


client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/QueryBuilder.java
```diff
@@ -88,6 +88,7 @@ private String recursionCheckedFields(TypeInfo type) {
             return "{ key " + keyFields + " value " + valueFields + "}";
         }
         return type.fields()
+                .filter(FieldInfo::isOutput)
```
jmartisk, line 91: Can't we handle this in `TypeInfo.fields()` instead, filtering out ignored fields from the collection, and thus completely ignore this field, not producing a `FieldInfo` for it at all? It seems unnecessary to create a `FieldInfo` for something that will be ignored.

t1: `TypeInfo.fields()` is used for different things, so this would have to be a new method. And as the query is being cached, the overhead shouldn't be big, should it?

jmartisk: `TypeInfo.fields` is called by the `QueryBuilder` when building a query (actually, right above the line that we're talking about), so if it didn't return ignored fields in the first place, then the QueryBuilder wouldn't have to filter here. I'm not sure what you mean it would have to be a new method.

Also AFAICS there are more places where we call `TypeInfo.fields()` this way you would have to add the same filter there too:
https://github.com/smallrye/smallrye-graphql/blob/1.4.2/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/json/JsonObjectReader.java#L33
https://github.com/smallrye/smallrye-graphql/blob/1.4.2/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/json/JsonObjectReader.java#L39
Potentially there will be more such places in the future - if we make `TypeInfo.fields()` not return the field, we won't have to keep in mind that this filtering needs to be applied.

t1: I hadn't checked the other places. But you're right: this is easier. Done.

jmartisk (approved):
Thanks!

PR #1331, opened 2022-04-01
Mulitple cleanups in prep for non blocking 

phillip-kruger:
This PR contains some cleanup on the way to get non blocking to work


phillip-kruger:
Yes, this was weird. I had to re generate the certs using java 11, then if was fine for both 11 and 17. if i generate with 17 it breaks the build

phillip-kruger:
@jmartisk - please check the "Refactor subscriptions to use async and reuse common code" commit, thet needs some eyes. The other is fairly small clean up

phillip-kruger:
Quarkus failure is expected, needs a small api fix in Quarkus. Merging for now, we can roll forward if there are any issues

PR #1346, opened 2022-04-08
Make sure to not create adapted types in the schema

### LINKED ISSUE #1342: Adapted field are still included in the scheme
This seem to only happen on input.

See https://github.com/quarkusio/quarkus/issues/24716

phillip-kruger:
Fix #1342
Signed-off-by: Phillip Kruger <phillip.kruger@gmail.com>

### DIFF
--- a/server/runner/pom.xml
+++ b/server/runner/pom.xml
@@ -129,7 +129,7 @@
             <plugin>
                 <groupId>org.wildfly.plugins</groupId>
                 <artifactId>wildfly-maven-plugin</artifactId>
-                <version>3.0.0.Final</version>    
+                <version>2.1.0.Final</version>    
                 <configuration>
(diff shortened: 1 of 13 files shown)


server/runner/pom.xml
```diff
@@ -129,7 +129,7 @@
             <plugin>
                 <groupId>org.wildfly.plugins</groupId>
                 <artifactId>wildfly-maven-plugin</artifactId>
-                <version>3.0.0.Final</version>    
+                <version>2.1.0.Final</version>
```
jmartisk, line 132: You sure?

phillip-kruger: Yes, it breaks the runner. Not too important, but dependably upgrade it a while ago, I just revert as it's not working

jmartisk (approved):

Issue #1349, opened 2022-04-09
ResponseReader locations parsing error (again)


cherrydev:
This is sort of related to #1153 where once again, I have found a graphql server (again, it's AWS AppSync) that sometimes spits back error messages that are slightly different than expected. In this case, the server is returning a message that looks like:
```json
{
  "data": null,
  "errors": [
    {
      "path": null,
      "locations": null,
      "message": "<some error message>"
    }
  ]
}
```
As you can see in `ResponseReader.java` when it is attempting to read the `"path"`, it defensively checks to make sure it's actually an array before attempting to parse it, but for `"location"` it doesn't check.

I'm wondering if it wouldn't also be a good idea to just wrap each portion of the error being decoded in a try/catch so that no matter what out-of-spec errors object the server returns, at least whatever is readable (hopefully the message!) will be returned.

PR #1350, opened 2022-04-09
Defensively decoding graphql errors

cherrydev:
This is to address #1349
While the error checking might seem excessive, because the code in question is error handling code that it should make every effort to be as defensive as is practical.


cherrydev:
Please note that the failing tests were failing before I got to it (and are unrelated)

jmartisk:
Can you please rebase and run a full Maven build to have the formatter plugin fix the formatting? And then push again.. I'd do that for you but seems I don't have a permission to write into your branch. Thanks

cherrydev:
@jmartisk Done!

jmartisk (approved):
Quarkus failures are unrelated, wildfly failures are because you introduced usage of `org.jboss.logging` in the client/implementation module, that will need a minor update on the WildFly side to fix the class loading, I'll take care of that.

Thanks!!

PR #1353, opened 2022-04-12
Refactor to make SmallRyeContext a POJO

phillip-kruger:
Still towards Non blocking... few more PR's to come...

Signed-off-by: Phillip Kruger <phillip.kruger@gmail.com>


server/implementation-cdi/src/main/java/io/smallrye/graphql/cdi/context/SmallRyeContextAccessorProxy.java
```diff
@@ -0,0 +1,302 @@
...
+import io.smallrye.graphql.execution.QueryCache;
+import io.smallrye.graphql.execution.context.DocumentSupplier;
+import io.smallrye.graphql.execution.context.SmallRyeContext;
+import io.smallrye.graphql.execution.context.SmallRyeContextManager;
+import io.smallrye.graphql.schema.model.Field;
+
+/**
+ * Class that serves as a proxy decorator for obtaining the current SmallRyeContext. It always calls
+ * SmallRyeContext.getContext() to obtain the current instance, and delegates to it.
+ * This is necessary because the SmallRyeContext is an immutable class, yet we, in some cases,
```
jmartisk, line 24: I'm not sure where this note is coming from (I think it was there before) but it's clearly NOT immutable. The part that we need to inject different instances during one request, and more importantly, even into the same `@GraphQLApi` endpoint instance, is still true. 

Could you rephrase it so that it is clear that it's needed because we need a single `@GraphQLApi` instance be able to see different instances of the context? (source method needs to see a different context than the operation root)

server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
```diff
@@ -46,407 +23,213 @@
...
-        if (dfe != null) {
-            return dfe.getGraphQlContext().get(key);
-        }
-        return null;
+    public String getOperationType() {
+        return this.operationType;
     }
 
-    @Override
-    public <K, V> void putMetaData(K key, V value) {
```
phillip-kruger, line 246: @jmartisk - see here, in the current version, the adding of metadata is not new...

server/implementation-cdi/src/main/java/io/smallrye/graphql/cdi/context/SmallRyeContextAccessorProxy.java
phillip-kruger: OK I removed all meta data. So users can still unwrap and get to dataFetcherEnv and add there, but not directly anymore.

jmartisk (approved):

PR #1397, opened 2022-05-13
fix #299: custom GraphQL error `extension` fields

### LINKED ISSUE #299: Add generic ability to add extension key-value for Errors
Relate to #282. We have code as a key supported with #295 but still need a generic key value.

Some more features:

- Move the knowledge to the model, so that we do not need to check annotations in runtime.
- Allow adding of Annotation on method that throw the exception, to get per instance usage (currently it's global)
- Add way to do all that is done for code for any other key-value

t1:
Add `ErrorExtensionProvider` via `ServiceLoader` and use it for `code` and `exception` (class name)

### DIFF
--- a/docs/custom-error-extensions.md
+++ b/docs/custom-error-extensions.md
@@ -1,0 +1,41 @@
+# GraphQL Error Extensions
+
+Exceptions are reported in GraphQL in the `errors` array, next to the `data` field, so it's possible to return partial results, e.g.:
+
+``` json
+{
+  "data": {
+    "superHero": {
+      "name": "Wolverine",
+      "location": null
+    }
+  },
+  "errors": [{
+    "message":"location unknown",
+    "path": ["superHero","location"],
+    "extensions":{"code":"location-unknown"}
+  }]
+}
+```
+
+The `location` field couldn't be returned, and in the `errors`, there's the reason with a few predefined fields,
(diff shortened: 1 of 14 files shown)


jmartisk:
Are you be willing to add a short example of how to use this into the documentation adoc files?

Also, @phillip-kruger after the experience with having to branch out 1.4.3, I think this should go into 1.6.0 rather than 1.5.1 (or maybe we won't plan a 1.6.0 at all, and just keep this in `main` for 2.0, without backporting at all) - WDYT?

phillip-kruger (approved):
LGTM, this does not look like a breaking change ? Or am I missing something ?

jmartisk:
Not breaking, but a new public API feature. With semantic versioning, new features mean a minor version bump (https://semver.org/#spec-item-7).

phillip-kruger:
a, ok gotcha. Yes then probably 2.0.0 and 1.6.0

jmartisk:
I've created a 1.6.x branch and sent a PR (https://github.com/smallrye/smallrye-graphql/pull/1400) backporting this one, so that we can properly run the CI (Quarkus and WildFly CI jobs are broken with main because main has already migrated to the jakarta namespace.... 1.6 branch should work with Quarkus and WF)

jmartisk (approved):

Issue #1423, opened 2022-05-27
Business Error Reporting


t1:
### Discussed in https://github.com/smallrye/smallrye-graphql/discussions/1410

Add a way to configure return types to be wrapped in `Result` types and map the exceptions accordingly.

```java
@Query Order order(@Id String id) throws OrderNotFoundException, OrderLockedException { ... }
```

Results in a schema:

```graphql
type OrderResult {
  order: Order
  error_orderLocked: OrderLocked
  error_orderNotFound: OrderNotFound
}

type Query {
  order(id: ID): OrderResult
}
```

t1:
In a later step, the `@GraphQLResult` annotation could take an enum parameter to define the type of wrapping applied. The above default is `ERROR_FIELDS`, a second value is `UNION` resulting in this schema:

```graphql
union OrderResult = Order | OrderLocked | OrderNotFound

type OrderLocked {
  ...
}

type OrderNotFound {
  ...
}
```

PR #1424, opened 2022-05-27
#1423: add GraphQLResult and wrap successful result

t1:
Also: rename Wrapper#isNotEmpty to Wrapper#isNonNull and fix some warnings

TODOs:

- [ ] Map all declared exceptions to fields
    - [ ] In the schema
    - [ ] When executing
- [ ] Consider `@GraphQLResult` annotations on an API class
- [ ] Consider `@GraphQLResult` annotations on a package
- [ ] Provide a config option to globally enable result-wrapping


t1:
@phillip-kruger: I have several smaller problems that you might be able to help with.

1. The `main` branch doesn't compile (`GraphQLWSSubprotocolHandler` and `GraphQLTransportWSSubprotocolHandler`)
2. In the `main` branch, the `ReferenceCreator` causes a StackOverflow
3. I've added a TCK test for `result`, but it fails because it doesn't find the `order` method, but the same request runs in the runner!

I could fix 1. and 2., but I don't understand how they got in there.

PR #1456, opened 2022-06-29
Update to Jandex 2.4.3 and avoid deprecated methods

Ladicek:
This PR also removes Power Annotations from the 1.6.x branch (already happened on `main`) and adds a note about locale-sensitive tests to the README.


Ladicek:
Failing jobs are an interesting case of dependency circularity :-) Fortunately, the entire point of Jandex 2.4.3 is that it doesn't matter in which order various projects update, so this PR staying open for a while doesn't block anything.

phillip-kruger:
Ok, I was thinking to remove power annotation only in 2.x, but we might as well remove it here too, as the version in 1.5.x is the same as here. We need a new home for power annotations. @t1 I messaged you on gitter. Lets discuss there.

phillip-kruger (approved):
LGTM ! Thanks a lot @Ladicek !

phillip-kruger:
@Ladicek So does this mean the Jandex 3.x upgrade will be easier ?

phillip-kruger:
The quarkus tests fail, as we also need the new version of Jandex in Quarkus. @Ladicek when will that happen ? We will need to do the graphql 1.6.1 upgrade with that PR

Ladicek:
Yes this is supposed to make Jandex 3.0 upgrade much easier. I described that in an email to quarkus-dev.

A PR to upgrade Jandex to 2.4.3 in Quarkus is already up: https://github.com/quarkusio/quarkus/pull/26442 -- I hope it's merged in a few days.

PR #1457, opened 2022-06-29
Update to Jandex 2.4.3 and avoid deprecated methods

Ladicek:
This PR also finishes removing Power Annotations (just some remnants) and adds a note about locale-sensitive tests to the README.


phillip-kruger (approved):
Thanks !

PR #1469, opened 2022-07-18
#400: Add support for GraphQL union type

### LINKED ISSUE #400: Union Types
We had a discussion on the weekly live stream about supporting Union Types (see GraphQL [tutorial](https://graphql.org/learn/schema/#union-types) and [spec](http://spec.graphql.org/draft/#sec-Unions)). Before we bring the discussion into MP GraphQL (see https://github.com/eclipse/microprofile-graphql/issues/32), we'd like to start an open discussion here:

# use cases

The idiomatic use-case in the tutorial is search results, i.e. I search for a string an get a list of objects that can be, e.g. a `Hero`, a `Starship`, or anything else. And the result list should render the matched items using a component with the appropriate data for the actual type.

I'm not sure, if this is possible in other GraphQL servers, but another use case could be when you evolve an API: you may start off with a simple `String` field for, e.g., an `email` address. But then you want to add more details, maybe the verification status. If it's only one field, you may add a `emailVerificationStatus`, but when there are many fields, it may be better to structure these fields into an `Email` type with `address`, `verificationStatus`, and other fields. To remain backwards compatible, clients requesting the `email` field without subfields will still get the `String`, while clients requesting sub fields will get those.

# thoughts about the API

A query could return a wrapper type with two or more type parameters (similar to the [`Either` type from vavr](https://www.vavr.io/vavr-docs/#_either)).

```java
@Query List<Union<Hero, Starship>> search(String queryString) { ... }
```

We would also need `Union3<T1, T2, T3>` up to some number of possible alternatives... maybe 10?

craig-day:
I eagerly started with [defining the spec](https://github.com/eclipse/microprofile-graphql/pull/459), but also wanted to get an implementation done. 

This adds support for creating GraphQL union types using an annotation `@Union` on a Java interface. I chose an empty Java interface so there would be an explicit layer of abstraction to represent the union itself. 

Example: 

```java
@Union 
interface MyUnion {
}

class Foo implements MyUnion {
  String getName();
}

class Bar implements MyUnion {
  int getCount();
}
```
would create
```graphql
type Foo {
  name: String
}

type Bar {
  count: Int
}

union MyUnion = Foo | Bar
```

An alternative could have been to directly annotate a type with the name of the union(s) it should be a member of, then generate the union from there.

<details>
  <summary>Expand for alternative Java example</summary>
  
```java
  @Union("MyUnion")
  class Foo {
    String getName();
  }
  
  @Union("MyUnion")
  class Bar {
    int getCount();
  }
  ```
</details>

Event thought this is a little less code, I didn't like it because the `@Union` annotation no longer describes an object directly, rather a relationship between objects. This relationship now only exists in the world of strings, and has no concrete representation in code.

For lots more thoughts, see the PR description here: https://github.com/eclipse/microprofile-graphql/pull/459

Resolves #400

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/Annotations.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/Annotations.java
@@ -207,7 +207,7 @@ public class Annotations {
         return new Annotations(annotationMap);
     }
 
-    // 
+    //
 
     /**
      * Used when we are creating operation and arguments for these operations
@@ -580,6 +580,7 @@ public class Annotations {
     public static final DotName INPUT = DotName.createSimple("org.eclipse.microprofile.graphql.Input");
     public static final DotName TYPE = DotName.createSimple("org.eclipse.microprofile.graphql.Type");
     public static final DotName INTERFACE = DotName.createSimple("org.eclipse.microprofile.graphql.Interface");
+    public static final DotName UNION = DotName.createSimple("io.smallrye.graphql.api.Union");
     public static final DotName ENUM = DotName.createSimple("org.eclipse.microprofile.graphql.Enum");
     public static final DotName ID = DotName.createSimple("org.eclipse.microprofile.graphql.Id");
     public static final DotName DESCRIPTION = DotName.createSimple("org.eclipse.microprofile.graphql.Description");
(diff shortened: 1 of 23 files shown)


phillip-kruger:
@t1 @andymc12 @jmartisk @WhiteCat22 @kaqqao any thoughts on this ?

phillip-kruger:
@craig-day the current main in SmallRye is using jakarta and are not yet release. Maybe you can also open a PR for the 1.6.x (javax) branch ?

craig-day:
@phillip-kruger done! #1470

phillip-kruger:
The failure of Quarkus build is expected with the Jakarta branch

phillip-kruger (approved):
Thanks !

PR #1470, opened 2022-07-19
#400: Add support for GraphQL unions (1.6.x)

craig-day:
This ports #1469 to be against 1.6.x (javax)

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/Annotations.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/Annotations.java
@@ -207,7 +207,7 @@ public class Annotations {
         return new Annotations(annotationMap);
     }
 
-    // 
+    //
 
     /**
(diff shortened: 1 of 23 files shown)


phillip-kruger (approved):
LGTM ! Thanks for this @craig-day - this is awesome. I think we should be able to move this to the spec soon. Next step is to get this into a runtime like Quarkus. For Quarkus we might have to do some work to get the classes registered. Are you keen to look at this ? If so we can do a SmallRye release once this is merge and you can look at the Quarkus implementation ?

phillip-kruger:
We would also need some documentation in the Quarkus Guide...

phillip-kruger:
Ok v 1.6.1 is released (https://github.com/smallrye/smallrye-graphql/releases/tag/1.6.1) @craig-day are you keen to get this into Quarkus ? I can assist if needed

craig-day:
@phillip-kruger I'll give it a shot today and see how it goes

PR #1476, opened 2022-07-20
default constructor for UnionType

craig-day:
When trying to pull the union type changes into quarkus, I ran into a build error

```
java.lang.RuntimeException: java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
	[error]: Build step io.quarkus.deployment.steps.MainClassBuildStep#build threw an exception: java.lang.RuntimeException: Unable to serialize objects of type class io.smallrye.graphql.schema.model.UnionType to bytecode as it has no default constructor
```

@phillip-kruger do you know if there is anything I could have done in this repo to prevent that from happening? Was a test missed somewhere?

### DIFF
--- a/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/UnionType.java
+++ b/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/UnionType.java
@@ -4,6 +4,9 @@ public final class UnionType extends Reference {
 
     private String description;
 
+    public UnionType() {
+    }
+
     public UnionType(String className, String name, String description) {
         super(className, name, ReferenceType.UNION);
         this.description = description;


phillip-kruger:
No not really. Typically once it's in Quarkus we have tests, but this is a bit of a chicken-egg problem

phillip-kruger:
Maybe we can release 1.6.2 so you can try an get this into Quarkus ?

craig-day:
I was able to use local maven with quarkus and got the tests running and passing. I'll link open a draft PR against quarkus to see if I'm missing anything.

craig-day:
The quarkus work is here: https://github.com/quarkusio/quarkus/pull/26865

The tests all work locally now. I was looking for places where there are docs about GraphQL Interfaces and was going to add a section next to it about unions, but couldn't find any. Should I add docs about both unions and interfaces somewhere?

PR #1559, opened 2022-09-16
Support directives on enums and enum values

acanda:
Adds support for directives on enums and enum values.

The following API results in the schema below:
```java
@GraphQLApi
public class EnumTestApi {

    @Query
    public int query(EnumWithDirectives enumWithDirectives) {
        return 0;
    }

    @EnumDirective
    public enum EnumWithDirectives {
        @EnumDirective
        A,
        B
    }

}

```

```
enum EnumWithDirectives @enumDirective {
  A @enumDirective
  B
}
```

### DIFF
--- a/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/EnumValue.java
+++ b/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/EnumValue.java
@@ -1,5 +1,8 @@
 package io.smallrye.graphql.schema.model;
 
+import java.util.Collection;
+import java.util.List;
+
 /**
  * Represents one of an enum's values. Is part of {@link EnumType}.
  *
@@ -9,13 +12,15 @@ public final class EnumValue {
 
     private String description;
     private String value;
+    private List<DirectiveInstance> directiveInstances;
 
     public EnumValue() {
     }
 
-    public EnumValue(String description, String value) {
+    public EnumValue(String description, String value, List<DirectiveInstance> directiveInstances) {
         this.description = description;
         this.value = value;
+        this.directiveInstances = directiveInstances;
     }
 
     public String getDescription() {
@@ -36,6 +41,19 @@ public final class EnumValue {
 
     @Override
     public String toString() {
-        return "EnumValue{" + "description=" + description + ", value=" + value + '}';
+        return "EnumValue{" +
+                "description=" + description +
+                ", value=" + value +
+                ", directiveInstances=" + directiveInstances +
+                '}';
     }
+
+    public boolean hasDirectiveInstances() {
+        return directiveInstances != null && !directiveInstances.isEmpty();
+    }
+
+    public Collection<DirectiveInstance> getDirectiveInstances() {
+        return directiveInstances;
+    }
+
 }
--- a/server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
@@ -32,6 +32,7 @@ import graphql.schema.GraphQLArgument;
 import graphql.schema.GraphQLCodeRegistry;
 import graphql.schema.GraphQLDirective;
 import graphql.schema.GraphQLEnumType;
+import graphql.schema.GraphQLEnumValueDefinition;
 import graphql.schema.GraphQLFieldDefinition;
 import graphql.schema.GraphQLInputObjectField;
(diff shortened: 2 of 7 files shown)


common/schema-model/src/main/java/io/smallrye/graphql/schema/model/EnumValue.java
```diff
@@ -36,6 +42,20 @@ public void setValue(String value) {
 
     @Override
     public String toString() {
-        return "EnumValue{" + "description=" + description + ", value=" + value + '}';
+        return "EnumValue{" +
+                "description=" + description +
+                ", value=" + value +
+                ", directiveInstances=" +
+                (directiveInstances == null ? "null" : Arrays.toString(directiveInstances.toArray())) +
```
t1, line 49:
```suggestion
                ", directiveInstances=" + directiveInstances +
```

Should look exactly the same: if it's null, it renders "null", if not, it renders the list in square brackets.

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -339,9 +340,20 @@ private void createGraphQLEnumType(EnumType enumType) {
...
                 .description(enumType.getDescription());
+        // Directives
+        if (enumType.hasDirectiveInstances()) {
+            enumBuilder = enumBuilder.withDirectives(createGraphQLDirectives(enumType.getDirectiveInstances()));
+        }
         // Values
         for (EnumValue value : enumType.getValues()) {
-            enumBuilder = enumBuilder.value(value.getValue(), value.getValue(), value.getDescription());
+            GraphQLEnumValueDefinition.Builder definitionBuilder = GraphQLEnumValueDefinition.newEnumValueDefinition()
+                    .name(value.getValue())
```
t1, line 350: It looks strange that the `name` and `value` both are set to `getValue()`, but it was like this before, so it's not getting worse. I don't understand what graphql-java is doing here.

t1 (dismissed):
LGTM (only minor questions)

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/EnumValue.java
acanda: You're right, I changed it.

Technically it depends on the implementation of List, but everything inheriting from AbstractCollection inherits the square brackets implementation.

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
acanda: I though exactly the same: looks suspicous but I won't change it in this PR.

acanda: As far as I can tell `value` should hold the "Java runtime value" for the respective GraphQL enum value, i.e. it is used to map the GraphQL enum value to the Java enum value when executing a query. See the [comment in EnumValuesProvider.java](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/idl/EnumValuesProvider.java#L6). It would probably be safe to not set it here since we're not executing any queries. At least the SchemaPrinter uses only the name, not the value.

t1: 👍thanks for the explanation. maybe a comment would be helpful?

acanda: The above explanation is my educated guess after diving into the code for about 10 minutes. Someone set the `value` for a reason, maybe it needs to be set to a value to omit an NullPointerException somewhere. I'd have to check if we could remove setting `value`, but that would be a different PR. I don't want to break things in this PR that are outside of the scope of supporting directives on enums.

t1: yeah, sure. I just thought that a comment could help the next person reading the line to understand what's happening.

phillip-kruger (approved):
LGTM

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
acanda: I looked into the issue and set `value`to null and had a couple of tests failing, so it looks like we need it like it was. I also added some comments.

PR #1561, opened 2022-09-16
Fix randomly failing SubscriptionHasToBeCancelledOnWebSocketCloseTestCase

acanda:
SubscriptionHasToBeCancelledOnWebSocketCloseTestCase was randomly failing on my machine because the hard-coded sleeps were not long enough. I replaced the sleeps with awaits that wait until the things the test expects actually happen or fail when it takes too long.


phillip-kruger (approved):
LGTM, Thanks !

phillip-kruger:
@jmartisk the WF build fails with noclassdeffound. Can you check ?

jmartisk:
@phillip-kruger that's expected because wildfly hasn't upgraded to jandex 3 yet

phillip-kruger:
Ok cool, thanks

PR #1564, opened 2022-09-21
Access Context in EventingService.errorDataFetch and .errorExecute

computerlove:
This is a suggestion adding new versions of EventingService.errorDataFetch and .errorExecute making `Context` available for the implementer.

`errorDataFetch` is called when an error occurs when fetching a field in the query.
It's signature is 
```java
public void errorDataFetch(String executionId, Throwable t)
```
Is there any reason for not exposing the `Context`/`DataFetchingEnvironment` of the execution?

For instance a query like 
```
query twoQueries {
   getA { a }
   getB { b }
} 
```
Only having access to executionId and exception makes it impossible to know whether data fetching operation getA or getB triggered `errorDataFetch`.


phillip-kruger (approved):
LGTM. We might need some code changes in the Quarkus extension for this.

phillip-kruger:
Thanks @computerlove !

jmartisk (approved):
Nice!

PR #1575, opened 2022-09-26
#1570: add executionResult field to SmallRyeContext

### LINKED ISSUE #1570: Provide a way to access an ExecutionResult instance
Currently there seems no way to access an `ExecutionResult` instance, because it is directly consumed by `ExecutionResponse` and `ExecutionResponseWriter`.
https://github.com/smallrye/smallrye-graphql/blob/main/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionService.java#L210

If we have the way to access that result instance, it's convenient to execute some result-dependent functionality e.g. custom logging.

How about to emit `@AfterExecute` event with not only `Context` instance but also `ExecutionResult` instance?
As a precedent, `@ErrorExecute` event is emitted with context and throwable instance.
https://github.com/smallrye/smallrye-graphql/blob/d15bbd47085f2dceddf4d5e655d3cc459fb008b2/server/implementation-cdi/src/main/java/io/smallrye/graphql/cdi/event/EventsService.java#L76

ikenox:
Fixes #1570

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
@@ -205,13 +205,6 @@ public class SmallRyeContext implements Context {
         this.executionResult = executionResult;
     }
 
-    public Optional<ExecutionResult> getExecutionResult() {
-        if (this.executionResult != null) {
-            return Optional.of(this.executionResult);
-        }
-        return Optional.empty();
-    }
-
     @Override
     public <T> T unwrap(Class<T> wrappedType) {
         // We only support DataFetchingEnvironment, ExecutionInput and Document at this point
@@ -225,6 +218,12 @@ public class SmallRyeContext implements Context {
                 return (T) documentSupplier.get();
             }
             return null;
+        } else if (wrappedType.equals(ExecutionResult.class)) {
+            if (executionResult != null) {
+                return (T) executionResult;
+            } else {
+                return null;
+            }
         }
         throw msg.unsupportedWrappedClass(wrappedType.getName());
     }
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import graphql.ExecutionResult;
 import jakarta.json.JsonArray;
 import jakarta.json.JsonObject;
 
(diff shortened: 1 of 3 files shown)


server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
```diff
@@ -199,6 +201,17 @@ public void setDocumentSupplier(DocumentSupplier documentSupplier) {
...
+    public void setExecutionResult(ExecutionResult executionResult) {
+        this.executionResult = executionResult;
+    }
+
+    public Optional<ExecutionResult> getExecutionResult() {
+        if (this.executionResult != null) {
+            return Optional.of(this.executionResult);
+        }
+        return Optional.empty();
+    }
```
phillip-kruger, line 213: We do not want to give direct access to graphql-java with a getter (the getters is also in the interface , Context, that will move to MicroProfile spec, and hence can not have any implementation specific things in there. So the way we have done that, is with the unwrap method. So you can remove this, and add a `} else if (wrappedType.equals(ExecutionResult.class)) {` in the unwrap that contains this.

phillip-kruger (changes requested):
LGTM, one change request

server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
ikenox: Thanks for your review. I could have fixed by dfe5f12

phillip-kruger (dismissed):
LGTM ! Thanks

server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java
phillip-kruger: Please run a full build locally and commit (see error in CI : `Please format file and commit before running validation!`)

phillip-kruger:
Something still off. Do a full build locally, then merge: `Error:  Failed to execute goal net.revelc.code:impsort-maven-plugin:1.6.2:check (default-cli) on project smallrye-graphql: Imports are not sorted in /home/runner/work/smallrye-graphql/smallrye-graphql/server/implementation/src/main/java/io/smallrye/graphql/execution/context/SmallRyeContext.java -> [Help 1]`

ikenox:
@phillip-kruger 
Sorry for failing CI repeatedly.
I've successfully done a build locally, and CI seems to be passed on my forked repository on e15deca ([link](https://github.com/ikenox/smallrye-graphql/pull/2)).

phillip-kruger:
Thanks !
Let see what CI does now.

phillip-kruger:
@jmartisk this can be backported to 1.8.x

jmartisk:
@phillip-kruger, sent a PR for that: https://github.com/smallrye/smallrye-graphql/pull/1576

Issue #1587, opened 2022-10-09
Support subprotocol graphql-transport-ws


gabrielsson:
As described in #811 and outlined really nicely by GavinRay97 in the comments beeing able to consume graphql over an authenticated websocket with subprotocol graphql-transport-ws is not working with the GraphQLTransportWSSubprotocolHandler as it lacks the means of payload in the connection init message:

the connection message cannot have a payload  GraphQLTransportWSSubprotocolHandler.java#L78
 ```
    this.connectionInitMessage = Json.createObjectBuilder().add("type", "connection_init").build();
 ```

while in the protocol it states the message is
```
interface ConnectionInitMessage {
  type: 'connection_init';
  payload?: Record<string, unknown>;
}
```
typically used for auth-tokens etc. 

edit: Better defining the problem

phillip-kruger:
Hi @gabrielsson . Are you keen to do a PR to add this ? I assume you are looking for a way to add the payload to the init message ?

gabrielsson:
@phillip-kruger Yes, I will take a look during this week!

jmartisk:
The `payload` field is optional, hence the question mark behind it in the protocol specification. Therefore saying that "graphql-transport-ws is not working" isn't correct - we only have a limitation in that we don't offer an API for applications to specify that they want to pass some payload to clients this way. Is this what you need? It could be achievable, we just didn't yet see any compelling use case for supporting it.

jmartisk:
Do you have any particular idea what the API should look like? How will the application define the payload items?

gabrielsson:
@jmartisk I would say that in the best of worlds it should be possible alongside where we set headers on the connection
```
CLIENT_NAME/mp-graphql/header/KEY
```
something like
```
CLIENT_NAME/mp-graphql/initpayload/KEY
```

But that might be a stretch goal. As I am using the DynamicGraphQLClientBuilder.java I would start there.

jmartisk:
It might need a way to specify the values dynamically because they might change over time, in which case a simple configuration property wouldn't work, but rather an SPI that the application will implement

gabrielsson:
I would look at the payload as something which is negotiating the connection. I agree that it might be useful to have a dynamical approach to set the initial payload, but as only one connection_init message can be sent during a connection, I believe it is a good idea to start treating the init payload as we treat the header property.

jmartisk:
There's one connection_init per connection, but over the runtime of a server, there will be tons of connections :)
[...]

PR #1590, opened 2022-10-12
Adding payload to `connection_init` message

gabrielsson:
Adding optional payload to the `connection_init` message, only on `graphql-transport-ws` sub protocol. Solving #1587. 

Usage: 

```
DynamicGraphQLClientBuilder.newBuilder()
        .websocketInitializationTimeout(5000)
        .initPayload(Map.of("token", "secret"))
        .url("wss://some.url")
        .build();
```

Tested in quarkus main branch with settings after adding the initPayload to the `GraphQLClientConfig` and necessary changes 
```
quarkus.smallrye-graphql-client."my-client".url=${url}
quarkus.smallrye-graphql-client."my-client".initPayload.token=${token}
```
Signed-off-by: Max Gabrielsson <max.gabrielsson@gmail.com>

### DIFF
--- a/client/api/src/main/java/io/smallrye/graphql/client/dynamic/api/DynamicGraphQLClientBuilder.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/dynamic/api/DynamicGraphQLClientBuilder.java
@@ -1,6 +1,7 @@
 package io.smallrye.graphql.client.dynamic.api;
 
 import java.util.Iterator;
+import java.util.Map;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 
@@ -17,6 +18,8 @@ public interface DynamicGraphQLClientBuilder {
     DynamicGraphQLClientBuilder configKey(String configKey);
 
     DynamicGraphQLClientBuilder header(String key, String value);
+
+    DynamicGraphQLClientBuilder initPayload(Map<String, Object> payload);
 
     DynamicGraphQLClientBuilder subprotocols(WebsocketSubprotocol... subprotocols);
 
--- a/client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/TypesafeGraphQLClientBuilder.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/TypesafeGraphQLClientBuilder.java
@@ -74,6 +74,11 @@ public interface TypesafeGraphQLClientBuilder {
      */
     TypesafeGraphQLClientBuilder header(String name, String value);
 
+    /**
+     * Static payload to send with initialization method on subscription.
+     */
+    TypesafeGraphQLClientBuilder initPayload(Map<String, Object> initPayload);
+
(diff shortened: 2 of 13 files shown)


jmartisk:
We talked about this together on Zulip, so just briefly the next steps:
- Support the `graphql-ws` protocol too
- Support configuring via `CLIENT_NAME/mp-graphql/init-payload/KEY=VALUE` properties

Also please update the relevant documentation (`docs/client_configuration.md`)

Developing tests probably doesn't make a lot of sense now until we have a server-side API able to expose the payload to applications...

phillip-kruger:
@jmartisk please have a look

jmartisk (commented):
Really nice! Just suggested one minor update to the documentation, please

gabrielsson:
Added the null check for the payload fromt the builder, missed that so the integration tests are now succeeding.

gabrielsson:
Since there are no converters for `Object` when adding the configuration to quarkus the `GraphQLClientConfig.initPayload` configuration must be of type `Map<String, String>` 

Converted in the `toSmallRyeNativeConfiguration` with something like
```
transformed.setInitPayload(Optional.ofNullable(quarkusConfig.initPayload)
                .map(m -> new HashMap<String, Object>(m)).orElse(null));
```

jmartisk (approved):
Quarkus 2.14 is coming soon, so I'll cut a 1.8.2 release (today or Monday).

Then I will need to upgrade Quarkus to this new version, so if you have a Quarkus branch with the Quarkus-side configuration enhancement, I can add that along with the upgrade - I just need a link to the branch and then I can handle it from there. Or, if you prefer, you can add that yourself later as a separate PR - we just need to make sure it is all merged to Quarkus by Wednesday 26th, I think.

Great work, thanks!

t1:
@jmartisk: it would be really nice, if you could also release a WildFly Feature Pack for 1.8.2.

jmartisk:
@t1, I'm afraid that supporting 1.8.x by the feature pack isn't possible because of a compatibility mismatch: 

- WildFly 23-26 has Jandex 2 and javax namespace 
- WildFly 27 will have Jandex 3 and Jakarta namespace
- SmallRye GraphQL 1.7.x has Jandex 2 and javax namespace
- SmallRye GraphQL 1.8.x has Jandex 3 and javax namespace
- SmallRye GraphQL 2.0.x will have Jandex 3 and Jakarta namespace

You see that SR-GQL 1.8.x can work neither with WildFly 23-26 (because of Jandex) nor with 27 (because of javax/jakarta). That's why I'm planning to skip supporting SR-GQL 1.8.x and 1.9.x, and jump straight to 2.0.x after WF 27 is released. See https://github.com/wildfly-extras/wildfly-graphql-feature-pack/wiki/Supported-versions-and-future-versions-roadmap

If this connection_init feature is important to you, the solution would be to backport it to 1.7.x and upgrade the feature pack to use that, I suppose that would work.

t1:
I guess I'll have to continue testing with WF27 and the RCs for now... going live will have to wait.

PR #1606, opened 2022-11-04
add the cdi extension service file also with the `jakarta` package name

t1:
This is necessary for the client CDI extension to work with Jakarta 9+


jmartisk (approved):
Good catch, I completely missed that. Thanks!

Issue #1616, opened 2022-11-16
Make ErrorExtensionProvider a singleton


pipinet:
<img width="1766" alt="image" src="https://user-images.githubusercontent.com/6096680/202229443-10cedfe9-e46b-45aa-9191-7586a80d7122.png">

it will create new instance every times error.
```java
ServiceLoader.load(ErrorExtensionProvider.class)
```

phillip-kruger:
Would you be keen to provide an pull request with the fix ?

jmartisk:
This will be resolved in 2.0.0, 1.9.0 and 1.8.3

PR #1617, opened 2022-11-17
#1616: make ErrorExtensionProvider a singleton

pipinet:


server/implementation/src/main/java/io/smallrye/graphql/execution/error/ExecutionErrorsService.java
```diff
@@ -37,6 +35,7 @@ public class ExecutionErrorsService {
     private static final Jsonb JSONB = JsonbBuilder.create(new JsonbConfig()
             .withNullValues(Boolean.TRUE)
             .withFormatting(Boolean.TRUE));
+    private static final ErrorExtensionProviders ERROR_EXTENSION_PROVIDERS = new ErrorExtensionProviders();
```
jmartisk, line 38: Can we avoid this being static? That would cause issues in scenarios where multiple GraphQL apps run in the same JVM (in WildFly, for example) because all apps would share the same value here, and therefore I envision that

- we would lose the ability for each application to have different providers
- if a provider is provided in the JAR of one application, attempt of another application to use it would lead to class loading problems

I think this field should just work if you remove `static`, no?

server/implementation/src/test/java/io/smallrye/graphql/execution/error/ErrorExtensionProvidersTest.java
```diff
@@ -0,0 +1,28 @@
...
+    @Test
+    void should_get_same_instance() {
+        List<ErrorExtensionProvider> result1 = providers.get().collect(Collectors.toList());
+        List<ErrorExtensionProvider> result2 = providers.get().collect(Collectors.toList());
+
+        Assertions.assertThat(result1.size()).isEqualTo(3);
+        assertEquals(result1.size(), result2.size());
+        assertTrue(result1.get(0) == result2.get(0));
+        assertTrue(result1.get(1) == result2.get(1));
+        assertTrue(result1.get(2) == result2.get(2));
```
t1, line 26: This would be more assertj-ish:

```suggestion
        assertThat(result1).hasSize(3);
        assertThat(result2).hasSize(result1.size());
        assertThat(result1.get(0)).isSameAs(result2.get(0));
        assertThat(result1.get(1)).isSameAs(result2.get(1));
        assertThat(result1.get(2)).isSameAs(result2.get(2));
```

t1 (approved):

server/implementation/src/main/java/io/smallrye/graphql/execution/error/ExecutionErrorsService.java
t1: +1

pipinet: agree， i didn't pay attention that ExecutionErrorsService is static in ExecutionResponse

server/implementation/src/test/java/io/smallrye/graphql/execution/error/ErrorExtensionProvidersTest.java
pipinet: fixed.

jmartisk:
Thanks!
I'll take care of backporting to 1.8 and 1.9 branches

PR #1658, opened 2022-12-29
Graphiql2 update 

debu999:
update the platform to graphiql 2 from previous graphiql version.
Changes added:
react.version updated from 17.0.2 to 18.2.0
graphiql.version updated from 1.9.13 to 2.2.0

### DIFF
--- a/client/api/pom.xml
+++ b/client/api/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.smallrye</groupId>
         <artifactId>smallrye-graphql-client-parent</artifactId>
-        <version>1.9.0</version>
+        <version>1.9.1-SNAPSHOT</version>
     </parent>
     
     <artifactId>smallrye-graphql-client-api</artifactId>
--- a/client/generator-test/pom.xml
+++ b/client/generator-test/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>io.smallrye</groupId>
         <artifactId>smallrye-graphql-client-parent</artifactId>
-        <version>1.9.0</version>
+        <version>1.9.1-SNAPSHOT</version>
     </parent>
 
     <artifactId>smallrye-graphql-client-generator-test</artifactId>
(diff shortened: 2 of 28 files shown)


jmartisk:
Thanks for your effort on this!

- I believe this is quite a big update and I'm not sure we want to do this within the 1.9.x stream, perhaps it would be a nice new feature to add in Quarkus 3.0, which would mean we do this in SmallRye GraphQL 2.0.x, so only in the `main` branch? @phillip-kruger WDYT?
- There seem to be some bogus commits included in your PR. It should be only one commit.
- I believe this will most likely need some changes in the Quarkus and WildFly integration (a few JavaScript code substitutions that we apply) and a bit of manual verification to make sure nothing breaks, I can look into that once we've decided on the branch to use

t1:
As it's not breaking any existing API, we don't _have_ to wait for 2.0, do we? It's nonetheless reasonable to consider the additional effort to backport.

t1:
Hmmmm.... it's only updates to the dependencies, so that's not a big deal, is it?

debu999:
> Thanks for your effort on this!
> 
> * I believe this is quite a big update and I'm not sure we want to do this within the 1.9.x stream, perhaps it would be a nice new feature to add in Quarkus 3.0, which would mean we do this in SmallRye GraphQL 2.0.x, so only in the `main` branch? @phillip-kruger WDYT?
> * There seem to be some bogus commits included in your PR. It should be only one commit.
> * I believe this will most likely need some changes in the Quarkus and WildFly integration (a few JavaScript code substitutions that we apply) and a bit of manual verification to make sure nothing breaks, I can look into that once we've decided on the branch to use

I believe the issue with commits were because i tried to build on top of 1.9.0 tag instead of 1.9.x branch. But the difference is just the one file . Once i commited just rebased on top  of 1.9.x branch. I did for 2.0.7 version of graphiql and later once 2.2.0 version was uploaded was able to do another commit. 

I feel the branding work might be needed on the UI. 
Rest all is exactly the same the look and feel is with graphiql 2.x version.

debu999:
I did raise another pr from main branch for review and merge. Since we use quarkus 2.x the jakarta change is causing some issue to test.  Feel free to review the other PR as well.

PR #1659, opened 2022-12-29
update the platform to graphiql 2 from previous graphiql version.

debu999:
Changes added:
react.version updated from 18.1.0 to 18.2.0
graphiql.version updated from 1.9.13 to 2.2.0


jmartisk:
Yes, most likely some changes will be needed there.

PR #1668, opened 2023-01-04
add IntelliJ plugin

### LINKED ISSUE #1667: Client IntelliJ Plugin
IntelliJ plugin to auto-complete Typesafe Client API interfaces and types/inputs from a `schema.graphql` file in your project directory.

t1:
fix #1667


jmartisk:
Wow.

I do use IntelliJ, so I shall give it a try.

How will the plugin publication work? Do you want to have some GH action, just like we have for the Gradle plugin, that publishes it?

I'm also thinking - shouldn't we have a separate repository for this? I would say that IDE plugins generally have versioning independent of the project they serve for. This way we would have too many releases of the plugin, many of which would contain no changes compared to the previous one. 

We can't assume that IntelliJ users will keep their plugin version aligned with the SmallRye GraphQL version that they are using anyway. It seems the plugin only has a compile-time dependency on graphql-java, not on SmallRye GraphQL stuff, so it shouldn't be a big problem to make it compatible with many SmallRye versions.

t1:
I just have started to read the docs about publication. At least the first one has to happen manually. And the GH Actions they use in the [template](https://github.com/JetBrains/intellij-platform-plugin-template) is quite complex; as I'm hardly using Gradle, I didn't fully understand it, yet.

I totally agree on your point on versioning and it's already decoupled (1.0-SNAPSHOT). We just would need to make sure that the normal release process doesn't release the plugin and vice versa... if we _do_ automate it... maybe it's not worth the effort... I don't know.

jmartisk:
Ah ok I didn't notice it has decoupled versions already - but it probably should still have a separate repo.
Having two projects with separate versioning in one git repo is something we should avoid (how would we create tags, for example?).
I think the plugin won't have that many releases, so perhaps the release can stay manual.. At least for now. We should have some instructions written for it, possibly in the wiki pages of the separate repository.
I also assume releasing requires some secret keys, we can talk about that privately and come up with a way of storing them.
So, shall I create a repo, will you move your work there?

t1:
I wouldn't say that it's really necessary to have a separate repo (we could use, e.g., tags like `intellij-plugin-1.0`), but I can absolutely live just fine with your decision.

You can go ahead and create a new repo, maybe even from the official template? There's just a lot of stuff in it.

Then I'd create a PR there. Thanks!

PR #1669, opened 2023-01-04
maven-cache the github setup-java action

t1:
There's a newer way to properly cache the maven repository to speed up build times

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,7 @@ jobs:
         with:
           java-version: ${{matrix.java}}
           distribution: 'temurin'
+          cache: 'maven'
 
       - name: Install graphviz
         run: sudo apt install graphviz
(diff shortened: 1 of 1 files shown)


t1:
Build times actually go down from ~14min to ~8min!

jmartisk:
Where did you see the base build take 14 minutes? It's always been around 8 (with a cache hit) :) I don't think this is meant to speed up the build compared to the old caching approach, it is just easier to configure. It still uses the  original [actions/cache](https://github.com/actions/cache) action under the hood, per the [documentation](https://github.com/actions/setup-java#caching-packages-dependencies)

The Quarkus job unfortunately takes 30+ minutes anyway, because we include native-mode tests, and we can't do much about it, so the time needed to fully test a PR is still the same.

But of course, it is easier to configure, so I guess we can do this.

t1:
Oh, I probably didn't look properly, as there was a change to the pom, so a cache-miss. Maybe it's still worth making it simpler, and it's in more places.

PR #1672, opened 2023-01-04
Throw exception on unknown input to prevent invalid json

### LINKED ISSUE #1671: Invalid json when using DynamicGraphQLClient DSL
It possible to create invalid request with the `DynamicGraphQLClient` DSL by adding an unsupported object type as arg to a field

```groovy
@Inject
@GraphQLClient
DynamicGraphQLClient client;
...
var response = client.executeSync(
  document(
    operation(
      field("test",
          List.of(arg("uuid", UUID.fromString("a0a0683e-d773-411b-a382-0591a09d9e0f"))), // Adding a List,UUID object instead of a string causes the invalid query json
          field("uuid")
      )
     )
  )
)
```

Problem is the fallback to `Object.toString` when writing the value to the request via `ValueFormatter`  

https://github.com/smallrye/smallrye-graphql/blob/9cefecb754da67af0da8d2ed714bb7e88878baf1/client/implementation/src/main/java/io/smallrye/graphql/client/impl/core/utils/ValueFormatter.java#L13-L36

I think format should also support `Iterables`, `? extends Number`, `UUID`, `Map<String,?>` and otherwise throw a exception

roggenbrot:
fixes #1671

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/impl/core/utils/ValueFormatter.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/impl/core/utils/ValueFormatter.java
@@ -2,6 +2,7 @@ package io.smallrye.graphql.client.impl.core.utils;
 
 import java.lang.reflect.Array;
 import java.time.LocalDate;
+import java.util.UUID;
 
 import io.smallrye.graphql.client.core.exceptions.BuildException;
 import io.smallrye.graphql.client.impl.core.EnumImpl;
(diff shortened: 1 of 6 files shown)


jmartisk:
From looking at the code, it looks great!
The build is complaining that your code isn't formatted, please run a Maven build again to have the maven-formatter-plugin do its job and then amend your commit... Thanks!

roggenbrot:
I formatted the code, but I'm running into problems during local build (Some Server TCK Test Case are failing). Hope this is only a local problem in my environment

roggenbrot:
Is there some extra setting required for the formatter plugin to work? It seems that the formatter plugin acts differently in my environment then on CI. In my environment the result seems to differ from run to run (Sometimes its valid, sometime it reports 2 other files then on CI)

jmartisk:
It still has unformatted code in `test` directories, maybe you excluded tests from the run completely.
I tried running a build on my local machine and it passed, let me push a commit to your branch that resolves the formatting.

roggenbrot:
Really strange, the formatter did no apply the same formatting in my environment (Sorting the imports ....)

jmartisk:
Maybe you have some specific configuration in your local settings, or it could simply be that you somehow excluded test code from the build.. Don't know, in any case I think it's working now, let's wait for WildFly and Quarkus tests to finish

jmartisk (approved):

PR #1675, opened 2023-01-05
1670 repeatable directives

t1:


t1:
Strange: `DynamicClientSSLTest.serverAuthentication_correctTruststore` fails with a timeout. @jmartisk: any idea?

t1:
Now the `FederationTestCase.testFederationMetadataInSchema` fails, because the directive changed. I tried to fix it, but a local build fails, so I couldn't test it. Line 60 should read: `.body(Matchers.containsString("type Product @extends @key(fields : \"id\")"));`.

jmartisk:
Yeah I can update that when upgrading the feature pack's codebase to a version that contains this change, thanks.
One of your commit messages says that you're updating the documentation, but I don't see that, could you add that? 
See https://smallrye.io/smallrye-graphql/2.1.0.Beta1/federation/ - we need to, at least, update the schema snippet at the end

t1:
The documentation update was only the descriptions of the federation annotations. I'll update the actual docs now; thanks reminding me.

docs/federation.md
```diff
@@ -1,6 +1,6 @@
 # Federation
 
-To enable support for [GraphQL Federation](https://www.apollographql.com/docs/federation), simply set the `smallrye.graphql.federation.enabled` config key to `true`.
+Support for [GraphQL Federation](https://www.apollographql.com/docs/federation) is enabled by default. If you add one of the federation annotations, the corresponding directives will be declared to your schema and the additional Federation queries will be added automatically. You can also disable Federation completely by setting the `smallrye.graphql.federation.enabled` config key to `false`.
```
jmartisk, line 3: Enabled by default is bad, IMO.
On the Quarkus side, I've already had a user complaining that they suddenly started seeing
```
type _Service {
  sdl: String!
}
```

in their application after upgrading the Quarkus version, even though they don't have anything related to Federation in it, and it broke some of their tooling.
My plan is to make Quarkus detect automatically whether federation should be enabled  by scanning through the application for any annotations from `io.smallrye.graphql.api.federation`. Of course, you will be able to override this automatic enabling using configuration.

This behavior is Quarkus specific though. WildFly can probably implement something similar if we want. So I'm not sure what to write in the general SmallRye documentation, because it potentially depends on the runtime that you're using. If we make both Quarkus and WildFly behave this way, I would say we can describe this  automatic enabling in the SmallRye documentation. WDYT?

jmartisk: So your current text is kinda correct in the sense that if you add a Federation annotation, it will appear in the schema, but it's wrong to say that "Federation is enabled by default", because we don't want to include the `type _Service` stuff by default. And we won't include directive type declarations in the schema either-  I have submitted https://github.com/smallrye/smallrye-graphql/pull/1678 to address this.

t1: The original suggestion to "enable" it by default was from @phillip-kruger, actually... where "enable" means automatic scanning, so the `_service` and `_entities` queries only appear when necessary. I thought this was already so... also in JEE containers.

jmartisk (approved):

PR #1701, opened 2023-01-22
activate test

robp94:
When we merged the change for kotlin notnull, jandex did not support it. Now it is supported and we can activate the test.


jmartisk (approved):

jmartisk:
Thanks!

PR #1710, opened 2023-02-01
graphql/client: Fix variable JSON serialization for bool/null

### LINKED ISSUE #1704: Boolean and null Request variables filtered before sent via GraphQL client
Hello! First off, thank you for all your work on this library.

I wanted to report a small bug I noticed in the GraphQL client. Specifically, variables of native type "boolean" or "null" get filtered out of the payload. It's possible "null" is intentional, but booleans feel like they should be included!

Here's a small repro in Kotlin:

```kotlin
    val r = RequestImpl("foo")
    r.setVariables(mapOf("int" to 1, "str" to "bar", "bool" to true, "null" to null))
    println(r.variables)
    println(r.toJsonObject())
```

Prints:

```
{int=1, str=bar, bool=true, null=null}
{"query":"foo","variables":{"int":1,"str":"bar"}}
```

I've tracked the bug down to the `RequestImpl`'s `toJsonObject()` call on the `variables` property. Specifically, in `_formatJsonVariables` there doesn't seem to be a branch for these types:

```java
    private JsonObject _formatJsonVariables() {
        JsonObjectBuilder varBuilder = Json.createObjectBuilder();

        variables.forEach((k, v) -> {
            // Other types to process here
            if (v instanceof String) {
                varBuilder.add(k, (String) v);
            } else if (v instanceof Integer) {
                varBuilder.add(k, (Integer) v);
            } else if (v instanceof JsonValue) {
                varBuilder.add(k, (JsonValue) v);
            } else {
                try (Jsonb jsonb = JsonbBuilder.create()) {
                    JsonStructure struct = ((JsonBinding) jsonb).toJsonStructure(v);
                    varBuilder.add(k, struct);
                } catch (Exception ignore) {
                }
            }
        });

        return varBuilder.build();
    }
```

I haven't tested it, but it might be as simple as adding a couple new branches. Here's the link to the code: https://github.com/smallrye/smallrye-graphql/blob/d274e03530f060820a1dddbc46ca5f166324be7c/client/implementation/src/main/java/io/smallrye/graphql/client/impl/RequestImpl.java#L54

Thanks in advance for your time and help!

akarle:
As discussed in #1704, the GraphQL client currently drops boolean and null variable values due to a bug in the JSON serialization of the variables:

```kotlin
    val r = RequestImpl("foo")
    r.setVariables(mapOf("int" to 1, "str" to "bar", "bool" to true, "null" to null))
    println(r.variables)
    println(r.toJsonObject())
```

Prints:

```
{int=1, str=bar, bool=true, null=null}
{"query":"foo","variables":{"int":1,"str":"bar"}}
```

This patch should fix the issue by updating the JSON serialization routine to add boolean and null types. It also adds a unit test to verify it works as expected.

Closes smallrye/smallrye-graphql#1704

---

Note: I'm pretty new to the java landscape, so the formatting might be a bit off (default IntelliJ) and/or the test might not follow best practices--all feedback welcome!

Thanks again!

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/impl/RequestImpl.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/impl/RequestImpl.java
@@ -62,6 +62,10 @@ public class RequestImpl implements Request {
                 varBuilder.add(k, (Integer) v);
             } else if (v instanceof JsonValue) {
                 varBuilder.add(k, (JsonValue) v);
+            } else if (v instanceof Boolean) {
+                varBuilder.add(k, (Boolean) v);
+            } else if (v == null) {
+                varBuilder.addNull(k);
             } else {
                 try (Jsonb jsonb = JsonbBuilder.create()) {
                     JsonStructure struct = ((JsonBinding) jsonb).toJsonStructure(v);
--- a/client/implementation/src/test/java/io/smallrye/graphql/client/RequestImplTest.java
+++ b/client/implementation/src/test/java/io/smallrye/graphql/client/RequestImplTest.java
@@ -1,0 +1,47 @@
+package io.smallrye.graphql.client;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.HashMap;
+
+import org.junit.jupiter.api.Test;
+
+import io.smallrye.graphql.client.impl.RequestImpl;
+
+/**
+ * Validates the JSON serialization of the query and variables in a RequestImpl
+ * <p>
+ * See https://github.com/smallrye/smallrye-graphql/issues/1704
+ */
+public class RequestImplTest {
+    @Test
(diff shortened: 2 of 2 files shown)


jmartisk:
It looks great! The CI pipeline is complaining that the `RequestImplTest.java` file is not properly formatted - we use the Maven formatter plugin to take care of that, so please just run another full Maven build, and the plugin should reformat it automatically, and then please amend and force-push your commit. Thanks!

jmartisk:
By the way no need for such lengthy commit messages, we don't usually do that here :) something like 
"Fix variable JSON serialization for bool/null + a test" would be sufficient

akarle:
Thanks for the review! Formatted via `mvn clean install` and force-pushed with a smaller commit message (feel free to trim fully on merge if still too long :))

jmartisk (approved):

jmartisk:
Thanks!

PR #1711, opened 2023-02-02
Add multi-module support to Maven plugin

eamelink:
The `smallrye-graphql-maven-plugin` breaks on multi-module projects:

* Without `includeDependencies` it won't find parts of the schema in other modules
* With `includeDependencies` it crashes, because it tries to index the module artifact as a `jar` (but it is a directory)

This PR adds support for multiple modules through the followin additions:

* If a dependency Artifact is not a jar but a dir, it's indexed it as a dir
* A new config setting `includeDependenciesGroupIds` allows filtering for group ids, to allow avoiding indexing all external dependencies.
* A test case


jmartisk (approved):
Awesome! Thanks a lot

PR #1752, opened 2023-02-21
Added feature to add extensions to SmallRyeContext object, added tests

### LINKED ISSUE #1750: Extensions Support
Is there support for returning an extensions object? I see in the ExecutionResponse that there is already code to add errors, data, and extensions to the response but I haven't found a way to add extensions to the response. I have tried accessing the context object in the @query annotated method to add extensions through the ExecutionInput object but it is too far down the chain of events at that point. Having a way for developers to add to the extensions object is useful for returning metadata to the client about the data and fields that are returned. In my case, a warning that a field is deprecated is very useful. 

Graphql-java already has support for instrumentation so I think being able to hook into that functionality is a good route.

mskacelik:
… tests
cc
@jmartisk
fixes: #1750

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -2,6 +2,12 @@
 
 Documentation is built using the `mkdocs` tool.
 See link:https://www.mkdocs.org/user-guide/installation/[Installation guide] on how to install it.
+On top of that install these 3 plugins:
+```
+pip install mkdocs-material
+pip install mkdocs-build-plantuml-plugin
+pip install mkdocs-macros-plugin
+```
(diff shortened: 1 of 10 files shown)


jmartisk:
CC @michaelbird1155 please have a look if this suits your needs

michaelbird1155:
@jmartisk This looks perfect! I was trying to set from within the context already so the implementation is right on!

jmartisk:
Ok we're just waiting for @mskacelik to add a little documentation enhancement related to this, and then we should be good to go

jmartisk (approved):

PR #1765, opened 2023-03-01
Change bean discovery mode to annotated in all beans.xml

manovotn:
Follow up on the discussion in Quarkus issue https://github.com/quarkusio/quarkus/issues/31378#issuecomment-1446381954

In short, a Quarkus app using graphql now logs warnings because graphql artifacts contains `beans.xml` with discovery mode `all` which is unsupported in CDI Lite and in Quarkus/Arc.

I quick scanned the sub modules containing `beans.xml` and it seems that all classes meant to be beans already have bean defining annotations (although keep in mind I am very much unfamiliar with this project).
The change seems to pass all tests in this repo and shouldn't affect Quarkus (since even up until now, Arc treated these JARs as `annotated` anyway).
It would be nice to test this with some other container such as WFLY to see if we're good. @jmartisk what's the easiest way to verify that? Are there some WFLY tests we could run? If so, where can I find them?

### DIFF
--- a/client/implementation/src/main/resources/META-INF/beans.xml
+++ b/client/implementation/src/main/resources/META-INF/beans.xml
@@ -1,5 +1,5 @@
 <beans xmlns="https://jakarta.ee/xml/ns/jakartaee"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/beans_3_0.xsd"
-       bean-discovery-mode="all">
+       bean-discovery-mode="annotated">
 </beans>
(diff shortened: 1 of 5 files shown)


jmartisk:
Let me run the tests for the wildfly graphql feature pack, that should be pretty quick (I'll just need to cherry-pick it into the 2.0.x branch, because we don't have a WildFly version that is compatible with 2.1 right now)

jmartisk:
Cool, it worked! 
Let's now wait for the Quarkus tests and then merge it

jmartisk (approved):

manovotn:
Glad to hear that, thanks for checking!

jmartisk:
Thanks! We just had a release yesterday, but we'll make another one ~ next week or so

Issue #1760, opened 2023-02-28
Support adding operation directives with dynamic client's DSL


jmartisk:
Something like

```
Document query = document(
            operation(
                field("customers",  directive("blabla", Argument.arg("param", 3)),
                    field("name"),
                    field("id")
                )
            )
        );
```
this would generate this query 
```
customers @blabla(param: 3) {
  name
  id
}
```

PR #1771, opened 2023-03-06
Added client-side implementation for adding directives into queries (…

mskacelik:
…dynamic client), added test
cc @jmartisk 
fixes: #1760

### DIFF
--- a/client/tck/src/main/java/tck/graphql/dynamic/core/BuildDirectiveTest.java
+++ b/client/tck/src/main/java/tck/graphql/dynamic/core/BuildDirectiveTest.java
@@ -1,0 +1,360 @@
+package tck.graphql.dynamic.core;
+
+import static io.smallrye.graphql.client.core.Argument.arg;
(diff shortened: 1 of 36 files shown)


client/tck/src/main/java/tck/graphql/dynamic/core/BuildDirectiveTest.java
```diff
@@ -0,0 +1,342 @@
...
+    @Test
+    public void buildInlineFragmentWithDirectiveTest() {
+        String expectedRequest = Utils.getResourceFileContent("core/directivesInlineFragment.graphql");
+        Document document = document(
+                operation(QUERY,
+                        "SimpleQuery",
+                        field("customers",
+                                field("name"),
+                                on("Customer",
+                                        List.of(
```
jmartisk, line 45: use `Directive.directives()` instead of `List.of` to create a list, it's the recommended way to use the DSL

jmartisk (approved):

Issue #1770, opened 2023-03-06
Dynamic client: more validations when building documents


jmartisk:
As an example, check that a field's or directive's name complies with the grammar as described here: https://spec.graphql.org/draft/#Name

PR #1779, opened 2023-03-14
More validation for building documents and added feature for creating Inline Fragments without TypeCond.

mskacelik:
cc @jmartisk 
fixes: #1770

### DIFF
--- a/client/api/src/main/java/io/smallrye/graphql/client/core/utils/validation/NameValidation.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/core/utils/validation/NameValidation.java
@@ -1,0 +1,94 @@
+package io.smallrye.graphql.client.core.utils.validation;
+
+import java.util.regex.Pattern;
+
+/**
+ * This class provides utility methods for validating names according to the GraphQL specification.
+ * 
+ * @see <a href="https://spec.graphql.org/draft/#Name">https://spec.graphql.org/draft/#Name</a>
+ */
+public class NameValidation {
+
+    /**
+     * The regular expression patterns for a valid GraphQL names.
+     */
+    private static final String _NAME_REGEX = "[a-zA-Z_][a-zA-Z0-9_]*";
+    private static final String _FIELD_NAME_REGEX = "^" + _NAME_REGEX + "(:" + _NAME_REGEX + ")?$";
+    private static final Pattern NAME_PATTERN = Pattern.compile(_NAME_REGEX);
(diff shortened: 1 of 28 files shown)


client/api/src/main/java/io/smallrye/graphql/client/core/utils/validation/NameValidation.java
```diff
@@ -0,0 +1,86 @@
...
+     * for alias.
+     *
+     * @param fieldName the field name to validate
+     * @return the validated field name
+     * @throws IllegalArgumentException if the field name is null or invalid
+     * @see <a href="https://spec.graphql.org/draft/#sec-Language.Fields">
+     *      https://spec.graphql.org/draft/#sec-Language.Fields
+     *      </a>
+     */
+    public static String setAndValidateFieldName(String fieldName) {
```
jmartisk, line 79: why `setAndValidateFieldName`? You're not setting it anywhere here (the same applies to other methods in this class)

client/api/src/main/java/io/smallrye/graphql/client/core/utils/validation/NameValidation.java
```diff
@@ -0,0 +1,86 @@
...
+ * This class provides utility methods for validating names according to the GraphQL specification.
+ * 
+ * @see <a href="https://spec.graphql.org/draft/#Name">https://spec.graphql.org/draft/#Name</a>
+ */
+public class NameValidation {
+
+    /**
+     * The regular expression patterns for a valid GraphQL names.
+     */
+    private static final String NAME_PATTERN = "[a-zA-Z_][a-zA-Z0-9_]*";
```
jmartisk, line 13: For performance purposes, can we compile these patterns into a `java.util.regex.Pattern` (using `Pattern.compile` and just once) and then use `pattern.matcher(actualString).matches()`? It will be faster than using `string.matches(pattern_as_string)` because that would mean recompiling the pattern every time again and again.

client/implementation/src/main/java/io/smallrye/graphql/client/impl/core/AbstractInputObjectField.java
```diff
@@ -20,7 +22,8 @@ public String getName() {
     }
 
     public void setName(String name) {
-        this.name = name;
+        this.name = setAndValidateName(name);
+        ;
```
jmartisk, line 26: Superfluous semicolon :)

jmartisk (approved):

Issue #1809, opened 2023-04-19
Generics on Mutations produce random / wrong Types in Schema


tacu90:
Hi everyone,

I'm trying to create an Interface that describes some queries / mutations that all my GraphQL Api Resources should implement. I'm using Generics to define the methods return and parameter types.

It seems that it is working as expected for Queries. The Return Type is always correctly set how it is defined in the implementing ressource class.
But it seems to be random if mutations are getting the correct / concrete Type or the abstract type.

Example:

```java
public abstract class Character {
  public String name;
  public String surname;
}
```
```java
@Input("HeroInput")
public class Hero extends Character {

  public String name;
  public String surname;
  public Double height;
  public Integer mass;
  public Boolean darkSide;
}
```

```java
public interface CharacterResource<T extends Character> {
  List<T> getAll();

  T add(T character);

  T remove(T character);

  T update(T character);

  T doSomething();
}
```

```java
@GraphQLApi
public class HeroResource implements CharacterResource<Hero> {

  @Override
  @Query("heroes")
  public List<Hero> getAll() {
    return null;
  }

  @Override
  @Mutation("addHero")
  public Hero add(@Name("hero") Hero character) {
    return null;
  }

  @Override
  @Mutation("removeHero")
  public Hero remove(@Name("hero") Hero character) {
    return null;
  }

  @Override
  @Mutation("updateHero")
  public Hero update(@Name("hero") Hero character) {
    return null;
  }

  @Override
  @Mutation("doSomething")
  public Hero doSomething() {
    return null;
  }

}
```

This is resulting in the following schema. Note that only updateHero has the correct concrete Types "HeroInput" and "Hero". The other three Mutations are using "CharacterInput" and "Character". This is changing everytime i restart the project.
I would expect that all Mutations are using "HeroInput" and "Hero" as their types.
```
type Character {
  name: String
  surname: String
}

type Hero {
  darkSide: Boolean
  height: Float
  mass: Int
  name: String
  surname: String
}

"Mutation root"
type Mutation {
  addHero(hero: CharacterInput): Character
  doSomething: Character
  removeHero(hero: CharacterInput): Character
  updateHero(hero: HeroInput): Hero
}

"Query root"
type Query {
  heroes: [Hero]
}

input CharacterInput {
  name: String
  surname: String
}

input HeroInput {
  darkSide: Boolean
  height: Float
  mass: Int
  name: String
  surname: String
}
```

Reproducer:
[...]

PR #1813, opened 2023-04-25
Filtered out synthetic methods in the schema

mskacelik:
fixes: #1809 
/cc @jmartisk @t1

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
@@ -9,6 +9,7 @@ import java.util.Optional;
 import java.util.Queue;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationTarget;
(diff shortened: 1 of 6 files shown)


common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
```diff
@@ -183,6 +183,23 @@ public void testSchemaWithDirectives() throws IOException {
...
+    public void testGenericSchemaBuilding() {
+        Indexer indexer = new Indexer();
+        indexDirectory(indexer, "io/smallrye/graphql/index/generic");
+        IndexView index = indexer.complete();
+
+        Schema schema = SchemaBuilder.build(index);
+
+        assertNotNull(schema);
+
+        String schemaString = JSONB.toJson(schema);
```
jmartisk, line 196: Instead of making very non-specific assertions about the schema string, couldn't we look at the operation arguments and return types?
Call `schema.getQueries()` or `schema.getMutations()`, retrieve operations from that set and then use `operation.getArguments()` for arguments and `operation.getReference()` for the return type?

jmartisk (changes requested):

jmartisk (approved):

common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
```diff
@@ -147,7 +148,8 @@ private List<MethodInfo> getAllMethodsIncludingFromSuperClasses(ClassInfo classI
         IndexView index = ScanningContext.getIndex();
         List<MethodInfo> methods = new ArrayList<>();
         while (current != null) {
-            methods.addAll(current.methods());
+            methods.addAll(
+                    current.methods().stream().filter(methodInfo -> !methodInfo.isSynthetic()).collect(Collectors.toList()));
```
Ladicek, line 152: Just for the record, this would be simpler and probably more efficient:

```java
current.methods().stream().filter(methodInfo -> !methodInfo.isSynthetic()).forEach(methods::add);
```

(Of course avoiding streams would be even more efficient, but that wouldn't be a one-liner anymore :-) )

Issue #1831, opened 2023-05-10
Feature for adding directives to whole GraphQL schema


mskacelik:
While keyword is reserved for schema via:
https://github.com/smallrye/smallrye-graphql/blob/2.1.3/server/api/src/main/java/io/smallrye/graphql/api/DirectiveLocation.java#L36
there seems to be missing an implementation, that handles adding directives to the GraphQL schema.

PR #1857, opened 2023-06-14
schema directives

mskacelik:
/cc @jmartisk 
fixes: #1831


jmartisk (changes requested):
Please don't use federation-related annotations in tests, this would tangle two unrelated concepts together. The tests should declare their own schema-applied directives as necessary

jmartisk (approved):

Issue #960, opened 2021-08-06
Show the `@RolesAllowed` as directives in the schema


t1:
If a user has access to a field (see #958 and #959), the schema also contains a corresponding directive indicating the roles that are allowed to see that field. E.g. a field annotated as `@RolesAllowed("admin")` has a directive `@rolesAllowed("admin")`.

t1:
veeery cool! big thanks to @mskacelik!!!

PR #1858, opened 2023-06-23
Added feature that creates RolesAllowed (security) directives into the graphql.schema fields

mskacelik:
/cc @jmartisk 
fixes: #960
Please check if the directive type is properly defined in the `RolesAllowedDirectivesHelper` class.


common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/OperationCreator.java
```diff
@@ -164,4 +171,16 @@ private Execute getExecution(Annotations annotationsForMethod, Annotations annot
         return Execute.DEFAULT;
     }
 
+    private void addDirectivesForRolesAllowed(Annotations annotationsForPojo, Annotations classAnnotations, Operation operation,
+            Reference parentObjectReference) {
+        if (rolesAllowedHelper != null) {
```
t1, line 176: this can never be null!

common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/RolesAllowedDirectivesHelper.java
```diff
@@ -0,0 +1,78 @@
...
+public class RolesAllowedDirectivesHelper {
+
+    public final static DirectiveType ROLES_ALLOWED_DIRECTIVE_TYPE;
+
+    static {
+        ROLES_ALLOWED_DIRECTIVE_TYPE = new DirectiveType();
+        ROLES_ALLOWED_DIRECTIVE_TYPE.setName("rolesAllowed");
+        Set<String> locations = new LinkedHashSet<>();
+        locations.add("FIELD_DEFINITION");
+        ROLES_ALLOWED_DIRECTIVE_TYPE.setLocations(locations);
```
t1, line 28:
```suggestion
        ROLES_ALLOWED_DIRECTIVE_TYPE.setLocations(Set.of("FIELD_DEFINITION"));
```

common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/RolesAllowedDirectivesHelper.java
```diff
@@ -0,0 +1,78 @@
...
+
+        addArgument("value", Scalars.getStringScalar());
+    }
+
+    private static void addArgument(String name, Reference reference) {
+        DirectiveArgument arg = new DirectiveArgument();
+        arg.setName(name);
+        arg.setReference(reference);
+        ROLES_ALLOWED_DIRECTIVE_TYPE.addArgumentType(arg);
+    }
```
t1, line 40:
```suggestion

        ROLES_ALLOWED_DIRECTIVE_TYPE.addArgumentType(createArgument("value", Scalars.getStringScalar());
    }

    private static void createArgument(String name, Reference reference) {
        DirectiveArgument arg = new DirectiveArgument();
        arg.setName(name);
        arg.setReference(reference);
        return arg;
    }
```

Makes easier to follow the flow, because all writing to the constant happens in a single method.

common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/RolesAllowedDirectivesHelper.java
```diff
@@ -0,0 +1,78 @@
...
+            if (annotationName.equals(DotName.createSimple("jakarta.annotation.security.RolesAllowed"))) {
+                DirectiveInstance directive = new DirectiveInstance();
+                directive.setType(ROLES_ALLOWED_DIRECTIVE_TYPE);
+                String value = getStringValue(methodAnnotations, annotationName, "value");
+                if (value == null) {
+                    value = getStringValue(classAnnotations, annotationName, "value");
+                }
+                directive.setValue("value", value);
+                result.add(directive);
+                break;
```
t1, line 62: really only the first one?

t1 (approved):
LGTM! Only a few minor suggestions.

common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/RolesAllowedDirectivesHelper.java
mskacelik: > really only the first one?
Since the `@RolesAllowed` annotation is non-repeatable (afaik), I thought it would make sense to stop the iteration.

t1: But then it doesn't make much sense to return a List, does it?

mskacelik: Hm, yea that's true. In the `Field` class there was no method for adding single `DirectiveInstance` (only for `List<DirectiveInstance>`). So I will just add it and refactor the code.

jmartisk (approved):

PR #1875, opened 2023-08-01
Reworking MetricServices

mskacelik:
# Reworking Metric Services.
/cc @jmartisk 

In this pull request, I have made significant changes to how metric services are handled. I introduced a new standalone class called `MetricsEmitter`, which is responsible for generating unique metric IDs. The previous approach relied on `EventEmitter` and context, which occasionally caused issues with IDs, particularly when dealing with source fields.

Tests are in the quarkus branch (commit)

quarkus branch: https://github.com/mskacelik/quarkus/tree/graphql-ui-update
> exact commit: https://github.com/quarkusio/quarkus/commit/7b20c2389c582f9a7329a9b3eb738d6e439d2530

Contains the fix for  https://github.com/quarkusio/quarkus/issues/34397


server/implementation/src/main/java/io/smallrye/graphql/execution/metrics/MetricsEmitter.java
```diff
@@ -0,0 +1,57 @@
...
+        Iterator<MetricsService> it = metricService.iterator();
+        List<MetricsService> enabledServices = new ArrayList<>();
+
+        while (enabled && it.hasNext()) {
+            try {
+                enabledServices.add(it.next());
+            } catch (Throwable t) {
+                // Ignore that service...
+                Throwable cause = t.getCause();
+                LOG.debug("Failed to register " + t.getMessage()
```
jmartisk, line 40: Can we remove this? Given that a `MetricsEmitter` is created multiple times (at least once for every operation) this can pollute the log unnecessarily

mskacelik: Yes, of course. Should I also delete the same log which is in the `EventEmitter` class?

jmartisk: I'd keep that one, because in this case it's not quite "expected" that the constructor throws an exception. In the metrics case, it will be quite usual and it won't necessarily be a sign of an actual error.

jmartisk (approved):
Awesome work, thanks!

phillip-kruger:
Nice work ! Thanks @mskacelik !

Issue #1896, opened 2023-08-22
@java.lang.Deprecated does not transform to Deprecated directive


mskacelik:
When adding `@java.lang.Deprecated` annotation to Java "Objects" that are represented in GraphQL as on of these locations: `FIELD_DEFINITION, ARGUMENT_DEFINITION, INPUT_FIELD_DEFINITION, ENUM_VALUE`.

The annotation does not transform itself to `Deprecated` directive even tho `DeprecatedDirectivesHelper.class` exists via (https://github.com/smallrye/smallrye-graphql/commit/492810f6db75d8c287ac0161ce23ca38589ca954). 

On top of that, having `@java.lang.Deprecated` on argument throws a NPE.
And having annotation `@java.lang.Deprecated` for high level fields (operations), `ENUM_VALUES` and `INPUT_FIELD_DEFINITION` seems to be missing in each *creator class.

Gonna have a look at this, since 492810f is my work...

PR #1897, opened 2023-08-31
Fixing some bugs regarding transformation of annotation 'java.lang.Deprecated' to graphQL's 'Deprecated'. directive.

mskacelik:
fixes:#1896
/cc @jmartisk 
## Changes:
Changes:
- [x] Refactored the code to ensure that the transformation methods return the `Optional<DirectiveInstance>` type instead of the raw type. This change eliminates situations where the method could return `null` or a list, as was the case with `DeprecatedDirectivesHelper`. Since the `java.lang.Deprecated` is non repeatable annotation.
- [x] Resolved a bug causing a NullPointerException when using `java.lang.Deprecated` during schema generation via the `smallrye-graphql-maven-plugin` goal: `generating-schema`. The issue stemmed from the fact that `GenerateSchemaMojo` did not scan the `smallrye-graphql-api`, where the GraphQL Deprecated annotation was defined.
- [x] Implemented a feature enabling the transformation of `java.lang.Deprecated` from operations (high-level fields), enum values, and field definitions of an output type. This addition complements the existing implementation for attribute definitions and input type field definitions. 

tests: https://github.com/quarkusio/quarkus/commit/204c115d9f23c22df54bdc9bd92284a87ddd99d8

### DIFF
--- a/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/EnumValue.java
+++ b/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/EnumValue.java
@@ -59,4 +59,7 @@ public final class EnumValue {
         this.directiveInstances = directiveInstances;
     }
 
+    public void addDirectiveInstance(DirectiveInstance directiveInstance) {
+        this.directiveInstances.add(directiveInstance);
+    }
 }
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/ArgumentCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/ArgumentCreator.java
@@ -103,14 +103,16 @@ public class ArgumentCreator extends ModelCreator {
             }
         }
         if (deprecatedHelper != null && directives != null) {
-            List<DirectiveInstance> deprecatedDirectives = deprecatedHelper
-                    .transformDeprecatedToDirectives(annotationsForThisArgument,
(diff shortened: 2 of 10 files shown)


common/schema-model/src/main/java/io/smallrye/graphql/schema/model/EnumValue.java
```diff
@@ -59,4 +59,9 @@ public void setDirectiveInstances(List<DirectiveInstance> directiveInstances) {
         this.directiveInstances = directiveInstances;
     }
 
+    public void addDirectiveInstance(DirectiveInstance directiveInstance) {
+        if (directiveInstances != null) {
```
jmartisk, line 63: Probably not a real issue, but generally I'm not quite happy that this method will basically do **nothing** in case that `directiveInstances == null`. I'd rather see a NPE thrown than silently ignoring the call.

jmartisk (changes requested):

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/EnumValue.java
mskacelik: Sure, I just went by the method `Field#addDirectiveInstances` https://github.com/smallrye/smallrye-graphql/blob/74107939776694ddc866b843d45f7a061f3cb3d8/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Field.java#L202
Should I also change the method `Field#addDirectiveInstance` which is right under it?

jmartisk:
Thanks!

PR #1898, opened 2023-09-01
Try to make federation work (with reactive)

robp94:
Since there are still problem with using federation like https://github.com/quarkusio/quarkus/issues/30180, and there seems to be no progress for quite some time, I tried to get it work for me.

Now there is no need for `quarkus.smallrye-graphql.nonblocking.enabled=false`. Although unis seem to work.
I am absolutely not sure if reactive stuff is handled correctly, but maybe someone can take a look? @t1 @jmartisk @phillip-kruger 

I tested with this quarkus app, query and variables to test via quarkus dev ui are included.
[code-with-quarkus-graphqll-federation.zip](https://github.com/smallrye/smallrye-graphql/files/12497533/code-with-quarkus-graphqll-federation.zip) (Although it is based on 2.3 since I use quarkus for testing)


phillip-kruger:
@jmartisk are you looking at this ?

PR #1907, opened 2023-09-06
Replace webjars with mvnpm

phillip-kruger:


jmartisk:
What exactly is the benefit from this? I'm worried about adding more problematic work to the productization team because they will have to productize these new libraries for RHBQ

phillip-kruger:
Do they currently do it for the web jars? We have much more control with mvnpm. It's smaller, and it auto update. It's replacing all web jars in Quarkus. So the new dev ui is using it, swagger ui now use it. All that is left is graphql

jmartisk:
Ah, ok if everything else is using it, then we probably want GraphQL too. Just a question, what do you mean by "auto update"? How does that work? I hope you don't mean that it will download a different artifact without us changing the version in the pom....

phillip-kruger:
No, but we should get dependabot alerts when a new version is released. So the sync between npm and maven central is automatically done. With web jars someone have to go to the website and do the sync.

Issue #1912, opened 2023-09-10
DynamicGraphQLClient subscriptions don't let you know if they fail


rubik-cube-man:
Given the following code:
```java
client.subscription(command, variables)
    .onFailure().invoke(throwable -> log.error("Error with subscription", throwable))
```
If you try and execute this against some random IP address that isn't a GraphQL server, the connection obviously fails. However, despite this, the API doesn't let you know, so there's no way to know if your subscription is active. The `onFailure` is only triggered if the connection WebSocket initially succeeds and fails whilst the subscription is active.

If the connection fails, you see something along these lines printed in the console:
```log
[-- Mutiny had to drop the following exception --]
Exception received by: io.smallrye.mutiny.infrastructure.Infrastructure.handleDroppedException(Infrastructure.java:295)
io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: no further information: localhost/127.0.0.1:8080
Caused by: java.net.ConnectException: Connection refused: no further information
	at java.base/sun.nio.ch.Net.pollConnect(Native Method)
...
[------------------------------------------------]
```

Also, if you attempt to connect to something and that connection fails, it permanently stores the broken `webSocketHandler`. This means all future subscriptions fail to connect as well, and you are forced to remake the client.

PR #1913, opened 2023-09-10
WebSocket connection failure handling

rubik-cube-man:
- Attempt to reconnect when the connection to the WebSocket fails.
- Pass down connection exceptions to created Multi.

Fixes https://github.com/smallrye/smallrye-graphql/issues/1912


jmartisk (approved):
Good catch! And thanks for the fix!

PR #1945, opened 2023-10-24
Fix field index computation for Source parameters

dpolysiou:
See https://github.com/quarkusio/quarkus/issues/36658

smallrye/smallrye-graphql#1314 also had a test, but it looks like it was removed as part of smallrye/smallrye-graphql#1825

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/validation/BeanValidationUtil.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/validation/BeanValidationUtil.java
@@ -11,6 +11,8 @@ import java.util.Set;
 import jakarta.validation.ConstraintViolation;
 import jakarta.validation.Path;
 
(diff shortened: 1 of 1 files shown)


jmartisk:
We've moved the tests to the Quarkus repo, because it now requires a full runtime with CDI and Hibernate Validator integrated, which was hard to achieve with just Jetty (which is used in the smallrye-graphql test suite), as far as I remember. See https://github.com/quarkusio/quarkus/blob/3.5.0/extensions/smallrye-graphql/deployment/src/test/java/io/quarkus/smallrye/graphql/deployment/GraphQLBeanValidationTest.java

So after we release a smallrye-graphql version with this fix included and upgrade Quarkus to use it, we can add a test there.

Thanks for your contribution!

jmartisk (approved):

phillip-kruger (approved):

Issue #1950, opened 2023-11-01
Issues in handling Mongo ObjectId


jmartisk:
See https://github.com/smallrye/smallrye-graphql/issues/1753#issuecomment-1469897009

When using ObjectId as input, the application receives a random (different) ObjectId than what was passed by the client

PR #1951, opened 2023-11-06
added fix for non-necessary json transformation for ObjectId argument value

mskacelik:
fixes: #1950
/cc @jmartisk


server/implementation/src/main/java/io/smallrye/graphql/execution/datafetcher/helper/ArgumentHelper.java
```diff
@@ -304,6 +304,9 @@ private Object correctObjectClass(Object argumentValue, Field field, DataFetchin
         } else if (receivedClassName.equals(String.class.getName())) {
             // We got a String, but not expecting one. Lets bind to Pojo with JsonB
             // This happens with @DefaultValue and Transformable (Passthrough) Scalars
+            if (field.getReference().getClassName().equals("org.bson.types.ObjectId")) { // edge case for ObjectId
```
jmartisk, line 307: Could you add a little  bit more of an explanation into the comment?
Something along the lines that we need to avoid letting JSON-B call the no-arg constructor of ObjectId, because then a new, random ID would get generated.

jmartisk (changes requested):
Nice! I've just asked to add a little more explanation why we're doing this,  because it's an ugly hack...

jmartisk (approved):

chokme:
Thanks for this fix :D

Issue #1963, opened 2023-11-13
GraphQL schema is not generated when mutation input has Long generic type


vtcosta:
I've been trying to run a similar solution as the one described in the [issue 1928](https://github.com/smallrye/smallrye-graphql/issues/1928), but including mutations. Here is a similar example as the other issue, including a mutation, with a parameterized type of String:

```
public interface Attribute<T> {
    T getValue();
    void setValue(T value);
}

public class Greet implements Attribute<String> {
    private String value;

    public Greet() {}

    public Greet(String value) {
        this.value = value;
    }

    @Override
    public String getValue() {
        return this.value;
    }

    @Override
    public void setValue(String value) {
        this.value = value;
    }
}

    @Query
    public Greet sayHello(String value) {
        return new Greet(value);
    }

    @Mutation
    public boolean updateGreet(Greet greet) {
        return true;
    }
```

The example above seems to be working just fine. But if I change the parameterized type to Long, it will throw the exception `graphql.AssertException: type Object not found in schema`

Here is the example:

```
public class Greet implements Attribute<Long> {
    private Long value;

    public Greet() {}

    public Greet(Long value) {
        this.value = value;
    }

    @Override
    public Long getValue() {
        return this.value;
    }

    @Override
    public void setValue(Long value) {
        this.value = value;
    }
}

    @Query
    public Greet sayHello(Long value) {
        return new Greet(value);
    }

    @Mutation
    public boolean updateGreet(Greet greet) {
        return true;
    }
```

It seems this Long problem is specific for the **mutation**, because if I leave just the **query**, it works.

PR #1966, opened 2023-11-14
Fixed a problem where during the scanning of InputTypes it included synthetic methods, added a filter + test

mskacelik:
//cc @jmartisk 
fixes: #1963

The problem was caused by not filtering bridge (synthetic) methods in  `InputTypeCreator` (INPUT). The previous issue only handled the `TypeCreator` (OUTPUT). That is why if the mutation method is not present in the issue's reproducer, it does not throw the Error. Since the given query has only Output type `Greet`.(it is not a MUTATION/QUERY problem but an Input/Output problem).

I am not sure why the reproducer with Type Variable `String` worked and wrapper classes (Long, Integer, etc.) did not—probably some Jandex magic.

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/InputTypeCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/InputTypeCreator.java
@@ -153,7 +153,7 @@ public class InputTypeCreator implements Creator<InputType> {
         // Find all methods and properties up the tree
         for (ClassInfo c = classInfo; c != null; c = ScanningContext.getIndex().getClassByName(c.superName())) {
             if (!c.toString().startsWith(JAVA_DOT)) { // Not java objects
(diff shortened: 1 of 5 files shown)


jmartisk (approved):

alaa-az:
This is breaking Quarkus @VirtualThread support , who should handle this yo or Quarkus team?

jmartisk:
@alaa-az hmmm, could you please file a new issue for it (it can be in this repo if you really believe that the problem is in this codebase), preferably with an executable reproducer to demonstrate the issue?

alaa-az:
> @alaa-az hmmm, could you please file a new issue for it (it can be in this repo if you really believe that the problem is in this codebase), preferably with an executable reproducer to demonstrate the issue?

Sorry for disturbing you. It turned out I am using the wrong annotation

jmartisk:
No problem @alaa-az, glad it's working

Issue #1965, opened 2023-11-14
Make GraphQLClientException(String, List<GraphQLError>) more NPE secure


api-from-the-ion:
In our test, we could get an NPE from GraphQLClientException. We just created an ArrayList<GraphQLError> and add null value to the list. Throwing this exception or using it as cause for another exception causing the NPE by printing stack trace, because then this exception is involved.

The toString() method of exception doesn't check if some elements of error list are null and causing an NPE by calling toString() of element.

I think that a small filter in the constructor could be a good solution. And the filter here, running once on the object creation, is better than the same filter in the to String() method, which would run over and over again.

I'll create a patch and PR for this.

PR #1967, opened 2023-11-14
[#1965] Filter out the null elements from the list of errors in Graph…

api-from-the-ion:
…QLClientException

### DIFF
--- a/client/api/src/main/java/io/smallrye/graphql/client/GraphQLClientException.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/GraphQLClientException.java
@@ -1,10 +1,12 @@
 package io.smallrye.graphql.client;
 
+import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.joining;
 
-import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
 /**
  * Represents a response that contained application-level errors and thus can't be turned into a domain object.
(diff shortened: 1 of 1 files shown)


jmartisk:
I agree this kinda makes sense, but I'm not completely sure I understand the point - is there a real possibility that, with a regular usage of the GraphQL client, the library tries to create a `GraphQLClientException` where one of the errors is null? Has that happened to you (except cases where you construct the exception directly)?

jmartisk:
And for consistency, if the variant that takes a single `GraphQLError` throws an exception on a null error, shouldn't the other one (that takes a list of errors) also throw an error instead of ignoring the null value?

api-from-the-ion:
For explanation, how we get here. We are writing some test for our service, that use other services (noting strange here). And we try to test the reaction of our service in the case that foreign service call have no result but an error. And here we're creating our own exception's instance of this class for the mock, for example.

Now I see our test is failing, but not the way that we like. It fails not because of reaction of our code on the error, but because logging or printing of the exception or something with toString() of the exception, caused another exception.

So, back to your questions: I think that this is possible in some business case, but would be unusual. But in any case, especially in testing, I would like to have an error (an exception in our case) as early as possible. Not in toString() call, but in the call of the constructor. Then I would immediately know that something went wrong.

To your other question: you are completely right. If we have an exception in one case, we should have an exception in other case also. Or no exception at all in any of these cases. Should I change nonNull() to requireNonNull()?

jmartisk:
Yeah I think that we should throw an exception if any of the errors is null. If there's a null error, something probably went really wrong, because that's normal at all, and we shouldn't silently ignore it.

api-from-the-ion:
So, I made the change and it now in the PR. I had the failure in the Quarkus tests, but I re-run them and now it's over. Looked like a temporary trouble with Quarkus image itself.

I thought once again about the whole situation. I already described what does it mean in the test environment. How can we get here in the live environment? We have a response where: error is null, or errors are null or some of the errors in the list are null. The attribute isn't missing, but null. Is this normal? I don't think so. The information should be here, but absent. So it would be OK even in a live system to have an exception because something is wrong with an error's structure. And this directly on the JSON mapping of the error.

jmartisk (approved):

jmartisk:
Thanks for your contribution!

PR #1968, opened 2023-11-14
Expose close status codes for WebSockets

rubik-cube-man:
Make WebSocket close exceptions expose the status code given by the server.

As it stands, there is no nice way to check the reason the WebSocket was closed. At the moment, in my code, I'm just checking if the message contains a String. This change would allow you to check the close status code for the WebSockets termination and handle it accordingly.

### DIFF
--- a/client/api/src/main/java/io/smallrye/graphql/client/UnexpectedCloseException.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/UnexpectedCloseException.java
@@ -1,0 +1,26 @@
+package io.smallrye.graphql.client;
+
+/**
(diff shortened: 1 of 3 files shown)


rubik-cube-man:
I pushed a fix for the formatting that caused the pipeline to fail!
I've just opened this as well: https://github.com/smallrye/smallrye-graphql/pull/1971 - I originally didn't have it formatted correctly because I noticed that after I ran a maven task, it was converting all my files to CRLF. After doing some research into it, I found out more about the formatter plugin, and though it'd be useful to define the line endings in the pom.

jmartisk (approved):

Issue #1978, opened 2023-11-22
Enum type is not supported in directive arguments


MrHekky:
Hello,
when I try to use an enum type in my directive declaration like this:
```@Target(ElementType.FIELD)
@Retention(RetentionPolicy.RUNTIME)
@Directive(on = DirectiveLocation.FIELD_DEFINITION)
@Name("MyDirective")
public @interface MyDirective {

    @NonNull
    MyEnum value();

}
```
And I try to start the Quarkus application I get the error:
> 2023-11-22 19:29:10,075 ERROR [io.qua.run.boo.StartupActionImpl] (Quarkus Main Thread) Error running Quarkus: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at io.quarkus.runner.bootstrap.StartupActionImpl$1.run(StartupActionImpl.java:113)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.ExceptionInInitializerError
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unsafe.java:1160)
	at java.base/jdk.internal.reflect.MethodHandleAccessorFactory.ensureClassInitialized(MethodHandleAccessorFactory.java:300)
	at java.base/jdk.internal.reflect.MethodHandleAccessorFactory.newConstructorAccessor(MethodHandleAccessorFactory.java:103)
	at java.base/jdk.internal.reflect.ReflectionFactory.newConstructorAccessor(ReflectionFactory.java:200)
	at java.base/java.lang.reflect.Constructor.acquireConstructorAccessor(Constructor.java:549)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:70)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:44)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:124)
	at io.quarkus.runner.GeneratedMain.main(Unknown Source)
[...]

PR #1979, opened 2023-11-24
Switching createGraphQLEnumTypes and createGraphQLDirectiveTypes to allow EnumTypes directive arguments

mskacelik:
/cc @jmartisk 
fixes: #1978 

Possible two edge case problems:
- if the EnumType has another directive whose arguments have a different EnumType that has yet to be added to the enumMap.
- Circular dependency: for example, EnumType has a directive whose argument is the same EnumType

For the second case, I think it's not doable even using the raw `java-graphql` API since the `GraphQLEnumType` would need a reference to itself in its definition.


jmartisk (approved):
Nice!

PR #1988, opened 2023-12-13
WIP Custom Scalar Proof of Concept

shirskor5:
What: This PR allows users to define custom scalars based on GraphQL's `String`, `Int`, and `Float` literal primitives. This design might also be extended to support custom scalars based on JSON literals in the future.

Why: Current functionality supports mapping object fields to the scalars that are supported by smallrye-grapqhl, however a user cannot define their own scalars in the resulting GraphQL schema.

How: This PR is suggesting a new annotation of `@CustomScalar("ScalarName")` and a limited set of interfaces representing each GraphQL literal primitive type.

The following is an example of an alternative BigDecimal scalar that serializes to a GraphQL String instead of a Float.

```
@CustomScalar("BigDecimalString")
public class BigDecimalString implements CustomStringScalar {
    public BigDecimalString(String stringValue) {
        ...
    }
    @Override
    public String stringValue() {
        ...
    }
}
```

Disclaimer: This is just a suggestion for how this might be done and we are very open to design feedback.  Would also like to note that the exception handling was purposefully left undone as we did not want to infer how the exception handling should be done.  As we move towards a final design we would appreciate any input on exception handling best practices for this project as well.

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/CustomScalarCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/CustomScalarCreator.java
@@ -1,0 +1,56 @@
+package io.smallrye.graphql.schema.creator.type;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.logging.Logger;
+
+import io.smallrye.graphql.schema.Annotations;
+import io.smallrye.graphql.schema.creator.ModelCreator;
+import io.smallrye.graphql.schema.creator.ReferenceCreator;
+import io.smallrye.graphql.schema.helper.DescriptionHelper;
+import io.smallrye.graphql.schema.helper.Directives;
+import io.smallrye.graphql.schema.model.CustomScalarType;
+import io.smallrye.graphql.schema.model.DirectiveInstance;
+
+public class CustomScalarCreator extends ModelCreator {
+
+    private static final Logger LOG = Logger.getLogger(CustomScalarCreator.class.getName());
+
+    private Directives directives;
+
+    public CustomScalarCreator(ReferenceCreator referenceCreator) {
+        super(referenceCreator);
+    }
+
+    public CustomScalarType create(
+            ClassInfo classInfo,
+            String scalarName) {
+        LOG.debug("Creating custom scalar from " + classInfo.name().toString());
+
+        Annotations annotations = Annotations.getAnnotationsForClass(classInfo);
+
+        return new CustomScalarType(
+                classInfo.name().toString(),
+                scalarName,
+                DescriptionHelper.getDescriptionForType(annotations).orElse(null),
+                classInfo.interfaceNames().stream().map(DotName::toString).collect(Collectors.toSet()));
+
+    }
+
+    @Override
+    public String getDirectiveLocation() {
+        return "SCALAR";
+    }
+
+    private List<DirectiveInstance> getDirectiveInstances(Annotations annotations,
+            String referenceName) {
+        return directives.buildDirectiveInstances(annotations, getDirectiveLocation(), referenceName);
+    }
+
+    public void setDirectives(Directives directives) {
+        this.directives = directives;
+    }
+}
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/CustomScalarCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/CustomScalarCreator.java
@@ -1,9 +1,17 @@
 package io.smallrye.graphql.schema.creator.type;
 
+import static io.smallrye.graphql.schema.model.CustomScalarType.CustomScalarPrimitiveType.FLOAT_TYPE;
+import static io.smallrye.graphql.schema.model.CustomScalarType.CustomScalarPrimitiveType.INT_TYPE;
+import static io.smallrye.graphql.schema.model.CustomScalarType.CustomScalarPrimitiveType.STRING_TYPE;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.ClassType;
 import org.jboss.jandex.DotName;
 import org.jboss.logging.Logger;
 
@@ -13,6 +21,7 @@ import io.smallrye.graphql.schema.creator.ReferenceCreator;
 import io.smallrye.graphql.schema.helper.DescriptionHelper;
 import io.smallrye.graphql.schema.helper.Directives;
 import io.smallrye.graphql.schema.model.CustomScalarType;
+import io.smallrye.graphql.schema.model.CustomScalarType.CustomScalarPrimitiveType;
 import io.smallrye.graphql.schema.model.DirectiveInstance;
 
 public class CustomScalarCreator extends ModelCreator {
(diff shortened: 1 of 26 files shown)


jmartisk:
Thanks, I'm gonna be playing around with this and throwing some notes at you.. :)
First off,
I'm getting this exception whenever I try to run it with Quarkus:

```
java.lang.NullPointerException: Cannot invoke "io.smallrye.graphql.schema.model.CustomScalarType$CustomScalarPrimitiveType.ordinal()" because "primitiveType" is null
	at io.smallrye.graphql.bootstrap.Bootstrap.getCoercing(Bootstrap.java:263)
	at io.smallrye.graphql.bootstrap.Bootstrap.createGraphQLCustomScalarType(Bootstrap.java:245)
	at io.smallrye.graphql.bootstrap.Bootstrap.createGraphQLCustomScalarTypes(Bootstrap.java:237)
	at io.smallrye.graphql.bootstrap.Bootstrap.generateGraphQLSchema(Bootstrap.java:170)
	at io.smallrye.graphql.bootstrap.Bootstrap.bootstrap(Bootstrap.java:126)
	at io.smallrye.graphql.cdi.producer.GraphQLProducer.initialize(GraphQLProducer.java:52)
	at io.smallrye.graphql.cdi.producer.GraphQLProducer.initialize(GraphQLProducer.java:42)
	at io.smallrye.graphql.cdi.producer.GraphQLProducer.initialize(GraphQLProducer.java:32)
	at io.smallrye.graphql.cdi.producer.GraphQLProducer.initialize(GraphQLProducer.java:27)
	at io.smallrye.graphql.cdi.producer.GraphQLProducer_ClientProxy.initialize(Unknown Source)
	at io.quarkus.smallrye.graphql.runtime.SmallRyeGraphQLRecorder.createExecutionService(SmallRyeGraphQLRecorder.java:36)
	at io.quarkus.deployment.steps.SmallRyeGraphQLProcessor$buildExecutionService1691419614.deploy_1(Unknown Source)
	at io.quarkus.deployment.steps.SmallRyeGraphQLProcessor$buildExecutionService1691419614.deploy(Unknown Source)
	at io.quarkus.runner.ApplicationImpl.<clinit>(Unknown Source)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:70)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:44)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:124)
	at io.quarkus.runner.GeneratedMain.main(Unknown Source)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at io.quarkus.runner.bootstrap.StartupActionImpl$1.run(StartupActionImpl.java:113)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

That is because of the `CustomScalarType` that is not compatible with Quarkus' requirements for bytecode recording. All fields need to have a properly named getter and setter, otherwise the fields will be null at runtime.
So you need to add a `setCustomScalarPrimitiveType` method and rename `customScalarPrimitiveType` to `getCustomScalarPrimitiveType`.

server/implementation/src/main/java/io/smallrye/graphql/scalar/custom/IntCoercing.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public IntCoercing(String customScalarClass) {
+        ClassloadingService classloadingService = ClassloadingService.get();
+        this.customScalarClass = (Class<? extends CustomIntScalar>) classloadingService.loadClass(customScalarClass);
+    }
+
+    private CustomIntScalar newInstance(BigInteger graphqlPrimitiveValue) {
+        try {
+            return graphqlPrimitiveValue == null ? null
+                    : customScalarClass.getConstructor(BigInteger.class).newInstance(graphqlPrimitiveValue);
```
jmartisk, line 25: AFAIK this requires that the user always creates a public constructor with this specific signature. That's not the the nicest requirement, but yeah I don't have a better idea that's not unnecessarily complicated.
 If we really do it, I think it would be nice to scan during build-time to verify that this constructor exists and throw an error right away, instead of letting it go all the way here (until a request at runtime). 
That could happen inside `CustomScalarCreator.create()` before you actually create the `CustomScalarType`instance - here you have the `ClassInfo` so you can check that the class has the appropriate constructor.

server/implementation/src/main/java/io/smallrye/graphql/scalar/custom/CustomFloatScalar.java
```diff
@@ -0,0 +1,48 @@
...
+                            .getConstructor(BigDecimal.class)
+                            .newInstance(jsonParser.getBigDecimal());
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
+    };
+
+    BigDecimal floatValue();
```
jmartisk, line 47: Hmmm... Would it be more understandable to call this method `serialize`, or `valueForSerialization`, or something, because it is used for serializing?

jmartisk:
Also, could you add a short story about how to use it in a new file in the `docs` directory? Plus some Javadocs for `CustomStringScalar`, `CustomIntScalar`, `CustomFloatScalar` and `@CustomScalar`

server/implementation/src/main/java/io/smallrye/graphql/scalar/custom/IntCoercing.java
shirskor5: Good idea! I also moved the definition of the primitive type into this class as well.

server/implementation/src/main/java/io/smallrye/graphql/scalar/custom/CustomFloatScalar.java
shirskor5: I think thats a reasonable name for this accessor method.

shirskor5:
Thank you for the tip on Quarkus' requirements for bytecode recording.  We have also written a test app for this and it should be working as expected now.  I have also added a feature doc along with some java docs where requested.  Happy to add more where it is appropriate to do so.

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/CustomScalarCreator.java
```diff
@@ -0,0 +1,92 @@
...
+            customScalarPrimitiveType = INT_TYPE;
+        } else if (interfaces.contains("io.smallrye.graphql.scalar.custom.CustomFloatScalar")) {
+            checkForOneArgConstructor(classInfo, BigDecimal.class);
+            customScalarPrimitiveType = FLOAT_TYPE;
+        } else if (interfaces.contains("io.smallrye.graphql.scalar.custom.CustomStringScalar")) {
+            checkForOneArgConstructor(classInfo, String.class);
+            customScalarPrimitiveType = STRING_TYPE;
+        } else {
+            //TODO error handle in the expected way
+            throw new RuntimeException("Required to implement a known CustomScalar primitive type. "
```
jmartisk, line 58: Please add the name of the class (`classInfo.name().toString`) into the message, otherwise it's not clear which class this is about

server/implementation/src/main/java/io/smallrye/graphql/scalar/custom/CustomStringScalar.java
```diff
@@ -0,0 +1,58 @@
...
+import jakarta.json.bind.serializer.SerializationContext;
+import jakarta.json.stream.JsonGenerator;
+import jakarta.json.stream.JsonParser;
+
+import io.smallrye.graphql.spi.ClassloadingService;
+
+/**
+ * A base class for all CustomScalars that are based on GraphQL's String.
+ */
+public interface CustomStringScalar {
```
jmartisk, line 18: I've been thinking that this is actually public API, so maybe this (and the other two) interfaces should be moved to the  `io.smallrye.graphql.api` package. The API module doesn't depend on JSON-B, so you'll also have to move out the `SERIALIZER` and `DESERIALIZER` fields somewhere else, I guess they could be directly in `io.smallrye.graphql.json.JsonBCreator` for instance?

docs/custom-scalar.md
```diff
@@ -0,0 +1,28 @@
...
+    public String stringValueForSerialization() {
+        ...
+  }
+}
+```
+In this example, `BigDecimalString` implements the `CustomStringScalar` which is used to identify the 
+proper (de) serialization for BigDecimalString.  `BigDecimalString` also provides a single argument
+constructor which takes a String.  Finally, `BigDecimalString` implements 
+`stringValueForSerialization()` which provides the String representation to be used during 
+serialization.
```
jmartisk, line 23: Please also add a note about the requirement to have a public constructor that takes `String`/`BigDecimal`/`BigInteger` depending on the scalar type

jmartisk:
Thanks!

PR #1990, opened 2023-12-18
typesafe client implementation with Jandex scanning

mskacelik:
/cc @t1 @jmartisk

WIP version of type-safe scanning and query building for Quarkus.
I also added a static directive handling

Would benefit by having tests...

This is a very early version, so stuff like proper names, packages, and overall refactoring will be done later.

I used help classes from the schema-builder, and left some of its methods just in case they will be needed later (later in my development).

Quarkus-side: https://github.com/mskacelik/quarkus/tree/typesafe-client-rework

A relevant piece of code on Quarkus-side: https://github.com/mskacelik/quarkus/blob/ecdbf30475b00277aa7ecf75790f8b6ded6f9bc2/extensions/smallrye-graphql-client/deployment/src/main/java/io/quarkus/smallrye/graphql/client/deployment/SmallRyeGraphQLClientProcessor.java#L96

### DIFF
--- a/client/model-builder/src/main/java/io/smallrye/graphql/client/model/ClientModelBuilder.java
+++ b/client/model-builder/src/main/java/io/smallrye/graphql/client/model/ClientModelBuilder.java
@@ -1,0 +1,93 @@
+package io.smallrye.graphql.client.model;
+
+import static io.smallrye.graphql.client.model.Annotations.GRAPHQL_CLIENT_API;
+import static io.smallrye.graphql.client.model.ScanningContext.getIndex;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
+import org.jboss.jandex.MethodInfo;
+import org.jboss.logging.Logger;
+
+import io.smallrye.graphql.client.model.helper.OperationModel;
+
+/**
+ * Builder class for generating client models from Jandex index.
+ * It scans for classes annotated with {@link Annotations#GRAPHQL_CLIENT_API} and generates client models based on the annotated
+ * methods.
+ */
+
+public class ClientModelBuilder {
+    private static final Logger LOG = Logger.getLogger(ClientModelBuilder.class.getName());
+
+    /**
+     * Builds the client models from the given Jandex index.
+     *
+     * @param index The Jandex index containing class information.
+     * @return The generated client models.
+     */
+    public static ClientModels build(IndexView index) {
+        ScanningContext.register(index);
+        return new ClientModelBuilder().generateClientModels();
+    }
+
+    private ClientModelBuilder() {
+    }
+
+    /**
+     * Generates the client models by scanning classes annotated with {@link Annotations#GRAPHQL_CLIENT_API}.
+     *
+     * @return The generated client models.
+     */
+    private ClientModels generateClientModels() {
+        ClientModels clientModels = new ClientModels();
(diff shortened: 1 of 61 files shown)


jmartisk:
@t1 just for context, the main goal is to have the queries scanned and computed at build time from Jandex, instead of at runtime using reflection. Configured clients in Quarkus (meaning, not dynamically created by builder) will always try to use this "ClientModel" which is a pre-computed representation of the operations that the client is aware of. We keep the original runtime-reflection-based approach in there too - dynamically created clients will use it.

client/api/src/main/java/io/smallrye/graphql/client/typesafe/api/TypesafeGraphQLClientBuilder.java
```diff
@@ -81,6 +82,8 @@ default TypesafeGraphQLClientBuilder headers(Map<String, String> headers) {
 
     TypesafeGraphQLClientBuilder subprotocols(WebsocketSubprotocol... subprotocols);
 
+    TypesafeGraphQLClientBuilder clientModel(ClientModels.ClientModel clientModel);
```
jmartisk, line 85: Only add this to our implementation of the builder (`VertxTypesafeGraphQLClientBuilder`) please, in the future we want to propose the general client API to the MP specification, and this method surely won't be part of the proposal. 

(Yeah, so to use it, it will be necessary to directly use `VertxTypesafeGraphQLClientBuilder`, or cast to it)

mskacelik:
Bump...
compatible Quarkus branch: https://github.com/mskacelik/quarkus/tree/typesafe-client-rework

# Tests 
There are three types of added tests:
- tck unit tests (using the `TypesafeTckClientModelSuite`)
These tests are executed with the client model scanning implementation.
I also fixed a bug where `java.lang.Date` acted like a `Date` (but it is a `DateTime` scalar) and made `java.sql.Date` a pure `Date.`
- Quarkus-sided tests (tests in Quarkus repository)
- Integration tests (vert.x module)
I "duplicated" the necessary tests and set them up to have the new model scanning implemented.
# Documentation
I'll do it later...

# Some other info:
I also added the temporary Quarkus-specific property `quarkus.smallrye-graphql-client.enable-build-time-scanning` for switching to the old implementation using the `java.lang.reflect` runtime scanning.

I also added static **directive** handling for the client. But only for high-level field (query) and variable definitions: `FIELD`, `VARIABLE_DEFINITION`.

`QUERY,` `MUTATION,` and `SUBSCRIPTION` locations could also be implemented, but I think it's a good topic for discussion. *How to differentiate between operation and high-level field directive...*

client/model-builder/src/main/java/io/smallrye/graphql/client/model/ClientModelBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+        Collection<AnnotationInstance> graphQLApiAnnotations = getIndex()
+                .getAnnotations(GRAPHQL_CLIENT_API);
+
+        graphQLApiAnnotations.forEach(graphQLApiAnnotation -> {
+            ClientModel operationMap = new ClientModel();
+            ClassInfo apiClass = graphQLApiAnnotation.target().asClass();
+            List<MethodInfo> methods = getAllMethodsIncludingFromSuperClasses(apiClass);
+            methods.stream().forEach(method -> {
+                String query = new QueryBuilder(method).build();
+                LOG.infof("[%s] - Query created: %s", apiClass.name().toString(), query);
```
jmartisk, line 63: I think this should rather be DEBUG than INFO

mskacelik:
new Quarkus relevant branch: https://github.com/mskacelik/quarkus/tree/srgql2.8.0
exact commit: https://github.com/quarkusio/quarkus/commit/34d794a6dafbe0026688e452898014261f8ffc2b

jmartisk (approved):

t1:
LGTM!

PR #1999, opened 2024-01-08
Ensure that longs are passed through to the variables json object

rubik-cube-man:
This fixes longs not working as a variable type.

If you execute the following code:
```java
String command =
"""
mutation executeSomethingWithLong($timestamp: BigInteger!) {
    doSomething(timestamp: $timestamp) {
        response
    }
}
""";
getClient().executeSync(command, Map.of("timestamp", System.currentTimeMillis()));
```

At the moment, the long gets ignored, and the value is then treated as null. The GraphQL server will then fail the request. This ensures that the long is correctly passed through.


jmartisk (approved):
Thanks!

PR #2004, opened 2024-01-13
fix #1996: add missing federation directives to index

### LINKED ISSUE #1996: Query "query getService { _service { sdl } }" ignores the directive @interfaceObject
I use the graphql federation 2 directive `@interfaceObject` in a subgraph:

```
type HumanResourceOrError @interfaceObject {
  id: ID!
  type: ResourceType!
}
```

The endpoint http://localhost:8080/graphql/schema.graphql returns a schema with the above type definition, but if I call the entdpoint http://localhost:8080/graphql with the following body, the directive `@interfaceObject` is missing in the response. Why?

```
query getService { _service { sdl } }
```

t1:


jmartisk (approved):
Thanks!

Issue #1753, opened 2023-02-22
AdaptWith/AdaptToScalar - ObjectID MongoDB input 


debu999:
If we have a graphql schema like below
```
{
  id: ObjectId
  details:[{id: Objectid, description: String}]
}
```

if we use @AdaptToScalar(Scalar.String.class) or @AdaptWith and provide a adaptor to convert from ObjectID to String
The system throws error complaining the transformation.

Ideally the mapping to any nested object must work with @AdaptToScalar or @AdaptWith annotations

Imagine if we have a nested JSON
```
{
  id: ObjectId
  details:[{id: Objectid, description: String,
                 definitions: [{id: Objectid, definition: String}]}]
}
```

in this scenario adding @AdaptsToScalar for String class in all nested object does not seem to do the right transformation. Can this be looked at. @AdaptToScalar only works in parent or root level conversion not inside objects which are nested.

phillip-kruger:
Are you talking client or server ? Do you have a resproducer ?

debu999:
Hi @phillip-kruger please find below the details.
It happens only for inputs. I believe default conversion to string is implemented. But we cant have the same in mutation a.k.a inputs

https://github.com/debu999/smallrye-graphql-objectid
if id is not in input works like a charm

![Screenshot 2023-03-04 at 7 10 33 PM](https://user-images.githubusercontent.com/1910577/222905631-cc239a27-e86b-4b7b-82a4-6cbab234d638.png)

with or without adapt to scalar it is able to translate it to string.

debu999:
```
{
  "data": {
    "__schema": {
      "queryType": {
        "name": "Query"
      },
      "mutationType": {
        "name": "Mutation"
      },
      "subscriptionType": null,
      "types": [
        {
          "kind": "ENUM",
          "name": "Action",
          "description": null,
          "fields": null,
          "inputFields": null,
          "interfaces": null,
          "enumValues": [
            {
              "name": "CREATE",
              "description": null,
              "isDeprecated": false,
              "deprecationReason": null
            },
            {
              "name": "DELETE",
              "description": null,
              "isDeprecated": false,
              "deprecationReason": null
            },
            {
              "name": "UPDATE",
              "description": null,
              "isDeprecated": false,
              "deprecationReason": null
            }
          ],
          "possibleTypes": null
        },
        {
          "kind": "SCALAR",
          "name": "BigDecimal",
          "description": "Scalar for BigDecimal",
          "fields": null,
[...]

PR #2005, opened 2024-01-14
Issue 1753 and Issue 1950

debu999:
Fix to provide custom serializer and deserialzer for Smallrye-graphql jsonb config.

Issues #1753  and Issues #1950 

@jmartisk @phillip-kruger 

This is my first contribution, so need your advise on the change. I have added few changes to make the JSONB Config exposed out. 

Can you have a look at it and share your feedback. Moreover need some help in testcases if those have to be added.

```
# Usage for custom changes will be exclusive to smallrye-graphql viz. no ObjectID dependency.
JsonBCreator.setJsonB(Collections.singletonList(OBJECTID_SERIALIZER),
        Collections.singletonList(OBJECTID_DESERIALIZER));
```

### DIFF
--- a/docs/custom-json-serializers-deserializers.md
+++ b/docs/custom-json-serializers-deserializers.md
@@ -1,6 +1,6 @@
-# Customizing JSON-B deserializers
+# Customizing JSON-B serializers/deserializers
 
-If your application needs finer-grained control over JSON deserialization than you can achieve via
+If your application needs finer-grained control over JSON serialization/deserialization than you can achieve via
 formatting annotations like `@JsonbDateFormat`, you may plug in your own custom instances of the `Jsonb` class
 for each input type that your GraphQL application exposes.
 
@@ -9,6 +9,8 @@ and implement its `overrideJsonbConfig` method.
 An example that plugs in a custom date format for a particular class that is used as input:
 
 ```
+package org.acme.custom.json.config;
+
 public class CustomJsonbService implements EventingService {
 
     @Override
@@ -19,7 +21,7 @@ public class CustomJsonbService implements EventingService {
     @Override
     public Map<String, Jsonb> overrideJsonbConfig() {
         JsonbConfig config = new JsonbConfig().withDateFormat("MM dd yyyy HH:mm Z", null);
-        return Collections.singletonMap("org.example.model.MyModelClass", JsonbBuilder.create(config));
+        return Collections.singletonMap("MyModelClass.class.getName()", JsonbBuilder.create(config));
     }
 }
 ```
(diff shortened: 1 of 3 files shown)


debu999:
Screenshot 2024-01-14 at 3 09 12 PM
Screenshot 2024-01-14 at 3 09 20 PM

jmartisk:
So, if I understand this correctly, the goal is to allow users to call `JsonBCreator.setJsonB(...)` to register their own serializers and this should happen at application startup?

I would be super happy if you could also

- Add a few lines about this into the documentation (a page similar to https://github.com/smallrye/smallrye-graphql/blob/main/docs/custom-json-deserializers.md, but for serializers)
- Add a test that demonstrates this (under https://github.com/smallrye/smallrye-graphql/tree/main/server/integration-tests/src/test/java/io/smallrye/graphql/tests)

jmartisk:
CI is complaining that the code is not formatted, please run a full maven build and commit+push afterwards, the formatter plugin will take care of it

jmartisk:
By the way, is there some reason why we shouldn't make it the same as for deserializers, that is in a method of `EventingService`, like it's described in https://github.com/smallrye/smallrye-graphql/blob/main/docs/custom-json-deserializers.md? With the current implementation, it's not completely clear when/how one should call `JsonBCreator.setJsonB`, so I think it would be nicer. `JsonBCreator` class wasn't meant to be public API.

debu999:
@jmartisk I was just trying to make it work  let me check the eventing service

jmartisk:
Thanks! :)

debu999:
@jmartisk I did validate seems only one liner change was needed in ArgumentHelper class. I have updated the documentation as well to reflect the same. Do you think we are good to merge now. Or we need to add documentation like 
```
@Override
    public Map<String, Jsonb> overrideJsonbConfig() {
        JsonbConfig config = new JsonbConfig().withDeserializers(OBJECTID_DESERIALIZER).withSerializers(
            OBJECTID_SERIALIZER);
        return Collections.singletonMap(ReactivePerson.class.getName(), JsonbBuilder.create(config));
    }
```

debu999:
Only change needed was `    String jsonString = JsonBCreator.getJsonB(className).toJson(m);`

jmartisk:
@chokme Can you confirm this solves your issue?

jmartisk:
@debu999, sure, we can have DEBUG-level logs for that!

debu999:
Will commit and push the changes in sometime. @jmartisk

jmartisk:
I've reformatted the code and changed the debug log a bit, let's see

jmartisk (approved):
Thanks!

PR #2006, opened 2024-01-16
fix value in JsonObjects name/value pair cannot be null

Hbllol:


Hbllol:
fix Value in JsonObjects name/value pair cannot be null

server/implementation/src/main/java/io/smallrye/graphql/execution/error/ExecutionErrorsService.java
```diff
@@ -132,15 +132,16 @@ private void addKeyValue(JsonObjectBuilder objectBuilder, String key, String val
     }
 
     private void addKeyValue(JsonObjectBuilder objectBuilder, String key, JsonValue value) {
+        if (value == null) {
```
jmartisk, line 135: Actually, I'm thinking - why should a null be ignored? Shouldn't we instead, if `value==null`, change it into `value=JsonValue.NULL` and simply go ahead with that?

jmartisk (approved):
Nice, thanks

Issue #1992, opened 2023-12-22
GraphQL & Kotlin: Schema Creator for Queries with Collections does not consider the NotNull/Nullable


vladi-bnv:
The SchemaCreator seems to ignore the Optional/NotOptional specifications for Kotlin's query methods 
Here is an example:
```
data class Output(val someNullableData: List<String?>?,
                  val someData: List<String>)

data class Input @JsonbCreator constructor(
    val someNullableData: List<String?>?,
    val someData: List<String>,
)


@ApplicationScoped
@GraphQLApi
class KotlinGraphQLQueries {
    // Input and output are nullable
    @Query
    fun queryOne(input: List<Input?>?): List<Output?>? {
        return listOf()
    }
    // List elements are nullable, lists are not nullable
    @Query
    fun queryTwo(input: List<Input?>): List<Output> {
        return listOf()
    }
    @Query
    fun queryThree(input: List<Input>): List<Output?> {
        return listOf()
    }
    // Nothing is nullable
    @Query
    fun queryFour(input: List<Input>): List<Output> {
        return listOf()
    }
}
```

Schema Creator produces:
```
type Output {
  someData: [String]! # Wrong: list elem is nonnull
  someNullableData: [String]
}
input InputInput {
  someData: [String]! # Wrong: list elem is nonnull
  someNullableData: [String]
}


type Query {  
  queryOne(input: [InputInput]): [Output]
  queryTwo(input: [InputInput]!): [Output]! # Wrong: output list elem is nonnull
  queryThree(input: [InputInput]!): [Output]! # Wrong: input list elem is nonnull
  queryFour(input: [InputInput]!): [Output]! # Wrong: input/output list elem are nonnull
}
```

The comments for the list elements don't seem to be working. At first glance, I also couldn't find them in the Jandex index.

jmartisk:
Yeah using Jandex for deducing the nullability of Kotlin types is quite tricky because of how the Kotlin compiler stores the metadata all over the place. We've only recently added some basic implementation of it (see https://github.com/smallrye/smallrye-graphql/blob/2.6.1/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/OperationCreator.java#L118) and there are definitely cases where this doesn't work properly. If you want, I invite you to dig into it (looking at the code I linked might be a good start), or I can try it, but not earlier than in January. Happy holidays :)

vladi-bnv:
Thank you, you are welcome to invite me!

[...]

PR #2008, opened 2024-01-18
Kotlin nullability for collection-wrapped types

vladi-bnv:
Fix for #1992

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/OperationCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/OperationCreator.java
@@ -1,16 +1,20 @@
 package io.smallrye.graphql.schema.creator;
 
 import java.lang.reflect.Modifier;
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.ParameterizedType;
 import org.jboss.jandex.Type;
 import org.jboss.logging.Logger;
 
 import io.smallrye.graphql.schema.Annotations;
+import io.smallrye.graphql.schema.Classes;
 import io.smallrye.graphql.schema.SchemaBuilderException;
 import io.smallrye.graphql.schema.helper.DeprecatedDirectivesHelper;
 import io.smallrye.graphql.schema.helper.Direction;
@@ -105,9 +109,7 @@ public class OperationCreator extends ModelCreator {
         addDirectiveForDeprecated(annotationsForMethod, operation);
         populateField(Direction.OUT, operation, fieldType, annotationsForMethod);
 
-        if (operation.hasWrapper()) {
-            checkWrappedTypeKotlinNullability(methodInfo, annotationsForClass, operation);
-        }
+        checkWrappedTypeKotlinNullability(methodInfo, annotationsForClass, operation);
         return operation;
     }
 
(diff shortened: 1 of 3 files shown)


common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/OperationCreator.java
```diff
@@ -130,36 +132,116 @@ private void checkWrappedTypeKotlinNullability(MethodInfo methodInfo,
...
-                KmTypeProjection arg = returnType.getArguments().get(0);
-                int flags = arg.getType().getFlags();
-                boolean nullable = Flag.Type.IS_NULLABLE.invoke(flags);
-                if (nullable) {
-                    operation.setNotNull(false);
+                // handle return type
+                if (operation.hasWrapper()) {
+                    KmType returnType = function.get().getReturnType();
+                    var nullable = isKotlinWrappedTypeNullable(returnType);
+                    // TODO: only for CollectionOrArrayOrMap applicable? What about other Wrapper?
```
jmartisk, line 139: I think the only other wrappers are reactive types, and that seems to be handled by the `else` branch, so I think it's ok

common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
```diff
@@ -324,6 +324,46 @@ public void testKotlinTypeNullability() {
...
+        assertTrue(arguments.get(2).isNotNull()); // i2
+        assertTrue(arguments.get(2).getWrapper().isNotEmpty()); // i2
+
+        assertTrue(arguments.get(3).isNotNull()); // i3
+        assertFalse(arguments.get(3).getWrapper().isNotEmpty()); // i3
+
+        assertFalse(arguments.get(4).isNotNull()); // i4
+        assertFalse(arguments.get(4).getWrapper().isNotEmpty()); // i4
+
+        // TODO: actually it is not consistent with java: if in java the wrapped list element is not null, the list will be marked as non-null in any case. Is this a problem?
```
jmartisk, line 362: Is this how Kotlin works? If yes, then I guess we shouldn't work around that

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/OperationCreator.java
```diff
@@ -130,36 +132,116 @@ private void checkWrappedTypeKotlinNullability(MethodInfo methodInfo,
...
+            signatureIsEqual = compareKotlinCollectionType(javaType, kotlinType);
+        }
+        // TODO: the matching of parameter types could use some improvements
+        // TODO: only some cases are handled, there are much more: see https://kotlinlang.org/docs/java-interop.html#mapped-types
+
+        boolean parametersAreEqual = true;
+        if (javaType instanceof ParameterizedType) {
+            List<Type> arguments = javaType.asParameterizedType().arguments();
+            for (int i = 0; i < arguments.size(); i++) {
+                Type javaArg = arguments.get(0);
```
jmartisk, line 212: Shouldn't this be `arguments.get(i)`?

vladi-bnv: sure!

common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
vladi-bnv: I also think that Kotlin and Java should behave in the same way, although it doesn't seem formally correct to me from the GraphQL point of view. But that's another topic

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/OperationCreator.java
```diff
@@ -130,36 +132,116 @@ private void checkWrappedTypeKotlinNullability(MethodInfo methodInfo,
...
+            }
+        }
+
+        return signatureIsEqual && parametersAreEqual;
+    }
+
+    private boolean compareKotlinCollectionType(Type javaType, KmType kotlinType) {
+        String javaCollectionType = javaType.name().toString();
+        String kotlinCollectionType = ((KmClassifier.Class) kotlinType.classifier).getName().replace("/", ".");
+        // TODO: naive implementation. Mutable Types are relevant?
```
jmartisk, line 229: Yeah I think the equals checks should also accept the mutable collection types, otherwise in `MutableList<Foo>` the `Foo` is detected as nullable

jmartisk (approved):
This is definitely a great improvement. Thanks!

Issue #2009, opened 2024-01-20
Updating Federation directives


RoKKim:
Since the last update to the Federation directives in SmallRye GraphQL, I have identified some changes that are not yet implemented (based on [definitions_fed2_5.graphqls](https://github.com/apollographql/federation-jvm/blob/main/graphql-java-support/src/main/resources/definitions_fed2_5.graphqls)):

1. The `@shareable` directive is now repeatable:
    ```graphql
    directive @shareable repeatable on FIELD_DEFINITION | OBJECT
    ```
2. The `@key` directive now has a `resolvable` argument, with a default value of true:
    ```graphql
    directive @key(fields: FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
    ```
3. New directives `@authenticated` and `@requiresScopes` have been added:
    ```graphql
    directive @authenticated on
        ENUM
      | FIELD_DEFINITION
      | INTERFACE
      | OBJECT
      | SCALAR
    
    directive @requiresScopes(scopes: [[Scope!]!]!) on
        ENUM
      | FIELD_DEFINITION
      | INTERFACE
      | OBJECT
      | SCALAR
    
    scalar Scope
    ```

I am also wondering if there is a specific reason for keeping the `federation-graphql-java-support` library at `2.1.1` version. I upgraded it to `4.3.0`, as the support for the above changes was added between these versions.

I have opened up a pull request #2010 with the changes so far. Here are my notes and questions:
1. `@authenticated`:
I have implemented the directive and tested it; the schema successfully generates:
    ```java
    @Authenticated
    public record Product(
    ...
    )
    ```
    ```graphql
    type Product @authenticated {
    ...
    }
    ```
2. `@key`:
I have added the `resolvable` argument to the directive.
Since the default value of argument `resolvable` is true, I modified the `io.smallrye.graphql.bootstrap.Bootstrap#createGraphQLDirectiveType` and `io.smallrye.graphql.bootstrap.Bootstrap#createGraphQLDirectiveFrom` methods to call `graphql.schema.GraphQLArgument.Builder#defaultValueProgrammatic` for each argument with the `@DefaultValue` annotation.
Currenty, the `createGraphQLDirectiveFrom` method will only adds arguments that are directly applied. For example, if we have:
    ```java
    @Key(fields = "id")
    public record Inventory(
    ```
[...]

PR #2010, opened 2024-01-20
Updating Federation directives

RoKKim:
#2009

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/DirectiveTypeCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/DirectiveTypeCreator.java
@@ -1,17 +1,24 @@
 package io.smallrye.graphql.schema.creator;
 
 import static io.smallrye.graphql.schema.Annotations.DIRECTIVE;
+import static io.smallrye.graphql.schema.Annotations.NON_NULL;
 import static java.util.stream.Collectors.toSet;
 
+import java.util.Collections;
 import java.util.Set;
 import java.util.stream.Stream;
 
 import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.ArrayType;
 import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.ClassType;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.Type;
 import org.jboss.logging.Logger;
 
+import io.smallrye.graphql.api.federation.policy.Policy;
+import io.smallrye.graphql.api.federation.requiresscopes.RequiresScopes;
 import io.smallrye.graphql.schema.Annotations;
 import io.smallrye.graphql.schema.helper.DescriptionHelper;
 import io.smallrye.graphql.schema.helper.Direction;
@@ -20,6 +27,10 @@ import io.smallrye.graphql.schema.model.DirectiveArgument;
 import io.smallrye.graphql.schema.model.DirectiveType;
 
 public class DirectiveTypeCreator extends ModelCreator {
+    private static final DotName POLICY = DotName.createSimple(Policy.class.getName());
+    private static final DotName REQUIRES_SCOPES = DotName.createSimple(RequiresScopes.class.getName());
+    private static final DotName STRING = DotName.createSimple(String.class.getName());
+
     private static final Logger LOG = Logger.getLogger(DirectiveTypeCreator.class.getName());
 
     public DirectiveTypeCreator(ReferenceCreator referenceCreator) {
@@ -31,8 +42,6 @@ public class DirectiveTypeCreator extends ModelCreator {
         throw new IllegalArgumentException(
                 "This method should never be called since 'DirectiveType' cannot have another directives");
     }
-
-    private static DotName NON_NULL = DotName.createSimple("org.eclipse.microprofile.graphql.NonNull");
 
     public DirectiveType create(ClassInfo classInfo) {
         LOG.debug("Creating directive from " + classInfo.name().toString());
@@ -48,10 +57,22 @@ public class DirectiveTypeCreator extends ModelCreator {
 
         for (MethodInfo method : classInfo.methods()) {
             DirectiveArgument argument = new DirectiveArgument();
-            argument.setReference(referenceCreator.createReferenceForOperationArgument(method.returnType(), null));
+            Type argumentType;
+            if (classInfo.name().equals(POLICY) || classInfo.name().equals(REQUIRES_SCOPES)) {
+                // For both of these directives, we need to override the argument type to be an array of nested arrays
+                // of strings, where none of the nested elements can be null
+                AnnotationInstance nonNullAnnotation = AnnotationInstance.create(NON_NULL, null,
+                        Collections.emptyList());
+                Type stringType = ClassType.createWithAnnotations(STRING, Type.Kind.CLASS,
+                        new AnnotationInstance[] { nonNullAnnotation });
+                argumentType = buildArrayType(stringType, 2, nonNullAnnotation);
+            } else {
+                argumentType = method.returnType();
+            }
+            argument.setReference(referenceCreator.createReferenceForOperationArgument(argumentType, null));
             argument.setName(method.name());
             Annotations annotationsForMethod = Annotations.getAnnotationsForInterfaceField(method);
-            populateField(Direction.IN, argument, method.returnType(), annotationsForMethod);
+            populateField(Direction.IN, argument, argumentType, annotationsForMethod);
             if (annotationsForMethod.containsOneOfTheseAnnotations(NON_NULL)) {
                 argument.setNotNull(true);
             }
@@ -71,5 +92,15 @@ public class DirectiveTypeCreator extends ModelCreator {
     private Set<String> getLocations(AnnotationInstance directiveAnnotation) {
         return Stream.of(directiveAnnotation.value("on").asEnumArray())
                 .collect(toSet());
+    }
+
+    private static Type buildArrayType(Type baseType, int dimensions, AnnotationInstance annotation) {
+        Type currentType = baseType;
+        for (int i = 0; i < dimensions; i++) {
+            ArrayType.Builder builder = ArrayType.builder(currentType, 1);
+            builder.addAnnotation(annotation);
+            currentType = builder.build();
+        }
+        return currentType;
     }
 }
--- a/server/implementation-servlet/src/main/java/io/smallrye/graphql/entry/http/IndexInitializer.java
+++ b/server/implementation-servlet/src/main/java/io/smallrye/graphql/entry/http/IndexInitializer.java
@@ -30,6 +30,7 @@ import org.jboss.jandex.Indexer;
 
 import io.smallrye.graphql.api.Deprecated;
 import io.smallrye.graphql.api.Entry;
+import io.smallrye.graphql.api.federation.Authenticated;
 import io.smallrye.graphql.api.federation.ComposeDirective;
 import io.smallrye.graphql.api.federation.Extends;
 import io.smallrye.graphql.api.federation.External;
@@ -41,6 +42,10 @@ import io.smallrye.graphql.api.federation.Provides;
 import io.smallrye.graphql.api.federation.Requires;
 import io.smallrye.graphql.api.federation.Shareable;
 import io.smallrye.graphql.api.federation.Tag;
+import io.smallrye.graphql.api.federation.policy.Policy;
+import io.smallrye.graphql.api.federation.policy.PolicyGroup;
+import io.smallrye.graphql.api.federation.requiresscopes.RequiresScopes;
+import io.smallrye.graphql.api.federation.requiresscopes.ScopeGroup;
 
 /**
  * This creates an index from the classpath.
@@ -89,6 +94,7 @@ public class IndexInitializer {
             indexer.index(convertClassToInputStream(Repeatable.class));
 
             // directives from the API module
+            indexer.index(convertClassToInputStream(Authenticated.class));
             indexer.index(convertClassToInputStream(ComposeDirective.class));
             indexer.index(convertClassToInputStream(Deprecated.class));
             indexer.index(convertClassToInputStream(Extends.class));
@@ -97,8 +103,12 @@ public class IndexInitializer {
             indexer.index(convertClassToInputStream(InterfaceObject.class));
             indexer.index(convertClassToInputStream(Key.class));
             indexer.index(convertClassToInputStream(Override.class));
+            indexer.index(convertClassToInputStream(Policy.class));
+            indexer.index(convertClassToInputStream(PolicyGroup.class));
             indexer.index(convertClassToInputStream(Provides.class));
             indexer.index(convertClassToInputStream(Requires.class));
+            indexer.index(convertClassToInputStream(RequiresScopes.class));
+            indexer.index(convertClassToInputStream(ScopeGroup.class));
             indexer.index(convertClassToInputStream(Shareable.class));
             indexer.index(convertClassToInputStream(Tag.class));
         } catch (IOException ex) {
--- a/server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
@@ -290,23 +290,22 @@ public class Bootstrap {
             directiveBuilder.validLocation(DirectiveLocation.valueOf(location));
         }
         for (String argumentName : directiveType.argumentNames()) {
-            GraphQLInputType argumentType = argumentType(directiveType.argumentType(argumentName));
-            if (directiveType.argumentType(argumentName).isNotNull()) {
-                argumentType = GraphQLNonNull.nonNull(argumentType);
+            DirectiveArgument argumentType = directiveType.argumentType(argumentName);
+            GraphQLInputType inputType = createGraphQLInputType(argumentType);
+
+            GraphQLArgument.Builder argumentBuilder = GraphQLArgument.newArgument()
+                    .name(argumentName)
+                    .type(inputType);
+
+            if (argumentType.hasDefaultValue()) {
+                argumentBuilder = argumentBuilder.defaultValueProgrammatic(
+                        sanitizeDefaultValue(argumentType));
             }
-            directiveBuilder = directiveBuilder
-                    .argument(GraphQLArgument.newArgument().type(argumentType).name(argumentName).build());
+
+            directiveBuilder = directiveBuilder.argument(argumentBuilder.build());
         }
         directiveBuilder.repeatable(directiveType.isRepeatable());
         directiveTypes.add(directiveBuilder.build());
-    }
-
-    private GraphQLInputType argumentType(DirectiveArgument argumentType) {
-        GraphQLInputType inputType = getGraphQLInputType(argumentType.getReference());
-        if (argumentType.hasWrapper() && argumentType.getWrapper().isCollectionOrArrayOrMap()) {
-            inputType = list(inputType);
-        }
-        return inputType;
     }
 
     private void addQueries(GraphQLSchema.Builder schemaBuilder) {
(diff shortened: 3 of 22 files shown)


RoKKim:
Will check the test results.

RoKKim:
I had to implement the following methods due to the upgrade of `federation-graphql-java-support`:

- `io.smallrye.graphql.scalar.number.NumberCoercing#valueToLiteral`
- `io.smallrye.graphql.scalar.time.DateCoercing#valueToLiteral`

Without these implementations, errors were produced during the execution of `CdiExecutionTest`:
```
2024-01-20T18:12:06.1113496Z [ERROR] io.smallrye.graphql.execution.SchemaTest.testSchemaModelCreation -- Time elapsed: 0.102 s <<< ERROR!
2024-01-20T18:12:06.1114956Z java.lang.UnsupportedOperationException: The non deprecated version of valueToLiteral has not been implemented by this scalar : class io.smallrye.graphql.scalar.number.NumberCoercing
2024-01-20T18:12:06.1116133Z 	at graphql.schema.Coercing.valueToLiteral(Coercing.java:228)
2024-01-20T18:12:06.1116678Z 	at graphql.schema.Coercing.valueToLiteral(Coercing.java:246)
2024-01-20T18:12:06.1117585Z 	at graphql.execution.ValuesResolverConversion.externalValueToLiteralForScalar(ValuesResolverConversion.java:225)
2024-01-20T18:12:06.1118641Z 	at graphql.execution.ValuesResolverConversion.externalValueToLiteral(ValuesResolverConversion.java:184)
2024-01-20T18:12:06.1119641Z 	at graphql.execution.ValuesResolverConversion.externalValueToLiteral(ValuesResolverConversion.java:175)
2024-01-20T18:12:06.1120684Z 	at graphql.execution.ValuesResolverConversion.externalValueToLiteralForObject(ValuesResolverConversion.java:318)
2024-01-20T18:12:06.1121731Z 	at graphql.execution.ValuesResolverConversion.externalValueToLiteral(ValuesResolverConversion.java:204)
2024-01-20T18:12:06.1122713Z 	at graphql.execution.ValuesResolverConversion.valueToLiteralImpl(ValuesResolverConversion.java:88)
2024-01-20T18:12:06.1123546Z 	at graphql.execution.ValuesResolver.valueToLiteral(ValuesResolver.java:248)
2024-01-20T18:12:06.1124207Z 	at graphql.schema.idl.SchemaPrinter.printAst(SchemaPrinter.java:718)
2024-01-20T18:12:06.1124844Z 	at graphql.schema.idl.SchemaPrinter.argsString(SchemaPrinter.java:811)
2024-01-20T18:12:06.1125570Z 	at graphql.schema.idl.SchemaPrinter.lambda$printFieldDefinitions$4(SchemaPrinter.java:536)
2024-01-20T18:12:06.1126405Z 	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
2024-01-20T18:12:06.1127084Z 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
2024-01-20T18:12:06.1127697Z 	at java.base/java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:395)
2024-01-20T18:12:06.1128348Z 	at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:258)
2024-01-20T18:12:06.1129501Z 	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:485)
2024-01-20T18:12:06.1130305Z 	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
2024-01-20T18:12:06.1131129Z 	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
2024-01-20T18:12:06.1132104Z 	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
2024-01-20T18:12:06.1132904Z 	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
2024-01-20T18:12:06.1133663Z 	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
2024-01-20T18:12:06.1134438Z 	at graphql.schema.idl.SchemaPrinter.printFieldDefinitions(SchemaPrinter.java:532)
2024-01-20T18:12:06.1135189Z 	at graphql.schema.idl.SchemaPrinter.lambda$objectPrinter$8(SchemaPrinter.java:642)
2024-01-20T18:12:06.1135928Z 	at graphql.schema.idl.SchemaPrinter.printSchemaElement(SchemaPrinter.java:1018)
2024-01-20T18:12:06.1136611Z 	at graphql.schema.idl.SchemaPrinter.print(SchemaPrinter.java:449)
2024-01-20T18:12:06.1137521Z 	at com.apollographql.federation.graphqljava.printer.ServiceSDLPrinter.generateServiceSDLV2(ServiceSDLPrinter.java:140)
2024-01-20T18:12:06.1138585Z 	at com.apollographql.federation.graphqljava.SchemaTransformer.build(SchemaTransformer.java:110)
2024-01-20T18:12:06.1139451Z 	at io.smallrye.graphql.bootstrap.Bootstrap.generateGraphQLSchema(Bootstrap.java:212)
2024-01-20T18:12:06.1140195Z 	at io.smallrye.graphql.bootstrap.Bootstrap.bootstrap(Bootstrap.java:126)
2024-01-20T18:12:06.1140948Z 	at io.smallrye.graphql.execution.SchemaTest.testSchemaModelCreation(SchemaTest.java:39)
2024-01-20T18:12:06.1141647Z 	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
2024-01-20T18:12:06.1142180Z 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
2024-01-20T18:12:06.1142701Z 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
```

I believe these `valueToLiteral` methods should cover most of the cases.

jmartisk:
Hmm, the `@requiresScopes` - and also `@policy` - directives will be quite tricky due to the 2D array. 
Also I think another problem with them is that the values of that array need to be a special scalar type named `federation__Scope` and `federation__Policy` [(see spec)](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#requiresscopes) and we currently have no straightforward way to have custom scalars as arguments of directives.

A rather crazy idea that I have is that

- on the Java declaration side, we let the `scopes` (and maybe also `policies`) arguments of type `String[]`, and specify (also in Javadoc etc.) that each of the strings in the array should be a comma-separated list of values
- programmatically unroll the `String[]` argument into `String[][]` by adding some special handling for these two directive types in https://github.com/smallrye/smallrye-graphql/blob/2.7.0/common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/Directives.java#L69
- make sure that the declaration refers to `federation__Scope` and `federation__Policy` scalars instead of `String` by adding some special handling for these directives in the `DirectiveTypeCreator`: https://github.com/smallrye/smallrye-graphql/blob/2.7.0/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/DirectiveTypeCreator.java#L50

But if it turns out to be too hacky and unwieldy, I'd say we shouldn't force-fit it... I'm not sure how much these things are in use, and whether it's absolutely necessary to support them at the cost of introducing ugly hacks.

jmartisk:
As for the change you've introduced that now shows the default values of directive arguments if they're not explicitly specified, I think that's a good idea. While AFAIU it's not strictly necessary, I guess it will help clear out some confusion.

server/implementation-servlet/src/main/java/io/smallrye/graphql/entry/http/IndexInitializer.java
```diff
@@ -87,6 +89,7 @@ private IndexView createCustomIndex() {
             indexer.index(convertClassToInputStream(Repeatable.class));
 
             // directives from the API module
+            indexer.index(convertClassToInputStream(Authenticated.class));
```
jmartisk, line 92: Note mostly to myself: the same change will also be needed in Quarkus (https://github.com/quarkusio/quarkus/blob/3.7.0.CR1/extensions/smallrye-graphql/deployment/src/main/java/io/quarkus/smallrye/graphql/deployment/SmallRyeGraphQLProcessor.java#L275)

RoKKim:
> Hmm, the `@requiresScopes` - and also `@policy` - directives will be quite tricky due to the 2D array. Also I think another problem with them is that the values of that array need to be a special scalar type named `federation__Scope` and `federation__Policy` [(see spec)](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#requiresscopes) and we currently have no straightforward way to have custom scalars as arguments of directives.
> 
> A rather crazy idea that I have is that
> 
> * on the Java declaration side, we let the `scopes` (and maybe also `policies`) arguments of type `String[]`, and specify (also in Javadoc etc.) that each of the strings in the array should be a comma-separated list of values
> * programmatically unroll the `String[]` argument into `String[][]` by adding some special handling for these two directive types in https://github.com/smallrye/smallrye-graphql/blob/2.7.0/common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/Directives.java#L69
> * make sure that the declaration refers to `federation__Scope` and `federation__Policy` scalars instead of `String` by adding some special handling for these directives in the `DirectiveTypeCreator`: https://github.com/smallrye/smallrye-graphql/blob/2.7.0/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/DirectiveTypeCreator.java#L50
> 
> But if it turns out to be too hacky and unwieldy, I'd say we shouldn't force-fit it... I'm not sure how much these things are in use, and whether it's absolutely necessary to support them at the cost of introducing ugly hacks.

Sounds like this is doable. I will give it a go as soon as I can.

RoKKim:
I managed to get this working with the following type of directive (example for `@Policy`):
```java
@Policy(policies = {@PolicyGroup({"policy1", "policy2"}), @PolicyGroup({"policy3"})})
```
I believe this approach is better than allowing users to define an array of strings separated by commas:
```java
@Policy(policies = {"policy1, policy2", "policy3"})
```
Now, achieving all of this isn't actually that hacky:

1. In `io.smallrye.graphql.schema.creator.DirectiveTypeCreator#create`, I overwrite the `argumentType` for both `@Policy` and `@RequiresScopes`, and add `NON_NULL` annotation to all levels, since this is the directive declaration:
    ```graphql
    directive @policy(policies: [[Policy!]!]!)
    scalar Policy
    ```
    Here is the current state of `io.smallrye.graphql.schema.model.DirectiveArgument`:
    image
    For now we still use `String` instead of `Policy` (which is just scalar), as we do for `@Key`, `@Requires` and `@Provides`:
    ```graphql
    directive @key(fields: FieldSet!, resolvable: Boolean = true)
    directive @requires(fields: FieldSet!)
    directive @provides(fields: FieldSet!)
    scalar FieldSet
    ```
    So if I understand the `io.smallrye.graphql.schema.model.DirectiveArgument` class correctly, we have:
  
    - a base type of `String`, for which the property `notNull` must be set to `true` beacuse of `Policy!`
    - a first wrapper around it, for which the property `notEmpty` must be set to `true`
    - a second wrapper around it, for which the property `notEmpty` must be set to `true`

2. In `io.smallrye.graphql.schema.helper.Directives#toDirectiveInstance` I parse the values of `@PolicyGroup` and `@ScopeGroup` annotations and save them into a nested list `List<List<String>>`.
3. To simplify the process, I used method `io.smallrye.graphql.bootstrap.Bootstrap#createGraphQLInputType` inside `io.smallrye.graphql.bootstrap.Bootstrap#createGraphQLDirectiveType`, as it already handles conversion of wrappers and I didn't want to duplicate the code.
Here is where I discovered a potential bug: the method in its current implementation does not correctly handle the `nonNull` attribute. For our case, which involves the type `[[String!]!]!` (since we currently use `String` instead of `Policy`), the method incorrectly generated `[[String!]]!`. I suspect this problem hasn't been identified because we haven't had to deal with such cases before.

Now my schema is successfuly generated:
```graphql
directive @policy(policies: [[String!]!]!) on SCALAR | OBJECT | FIELD_DEFINITION | INTERFACE | ENUM
type Product @authenticated @key(fields : "id")@policy(policies : [["policy1", "policy2"], ["policy3"]]) @requiresScopes(scopes : [["scope1", "scope2"], ["scope3"]]) {
  ...
}
```

However, a problem arose during testing. When running the test `io.smallrye.graphql.execution.CdiExecutionTest#testInputWithDifferentNameOnInputAndType`, I encountered the following error:
```
INFO: {"errors":[{"message":"Validation error (WrongType@[createNewHero]) : argument 'hero' with value 'ObjectValue{objectFields=[ObjectField{name='realName', value=StringValue{value='Steven Rogers'}}, ObjectField{name='name', value=StringValue{value='Captain America'}}, ObjectField{name='dateOfLastCheckin', value=StringValue{value='09/25/2019'}}, ObjectField{name='patrolStartTime', value=StringValue{value='13:00'}}, ObjectField{name='timeOfLastBattle', value=StringValue{value='09:43:23 21-08-2019'}}, ObjectField{name='tshirtSize', value=EnumValue{name='XL'}}]}' is missing required fields '[equipment, superPowers]'","locations":[{"line":2,"column":19}],
```

This is the object `org.eclipse.microprofile.graphql.tck.apps.superhero.model.SuperHero` we use for testing:
```java
public class SuperHero implements Character {
    private List<Team> teamAffiliations;
    private List<@NonNull String> superPowers;
```
Therefore, the GraphQL type should be `[String!]`.
Considering the first point of how `notNull` and `notEmpty` should reflect in GraphQL types and the fixes I have implemented, the type I get with my changes is  `[String]!`, since this is the state of `superPowers` field:
image

So my question is, in the context of handling `notNull` and `notEmpty` properties, should the `io.smallrye.graphql.schema.creator.WrapperCreator` class return `true` for `notNull` on the `String` reference and `false` for `true` on the `List` wrapper?

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -846,70 +881,67 @@ private GraphQLInputObjectField createGraphQLInputObjectFieldFromField(Field fie
...
-                graphQLInputType = GraphQLNonNull.nonNull(graphQLInputType);
-            }
-            // Collection depth
+            // Loop as long as there is a wrapper
             do {
                 if (wrapper.isCollectionOrArrayOrMap()) {
                     graphQLInputType = list(graphQLInputType);
+                    // Wrapper itself can also be mandatory
+                    if (wrapper.isNotEmpty()) {
+                        graphQLInputType = GraphQLNonNull.nonNull(graphQLInputType);
```
jmartisk, line 907: Actually I think the `wrapper.isNotEmpty()` method has a rather confusing name and it should instead be `isWrappedTypeNotNull()` - it refers to non-nullability of the wrapped type, not the wrapper itself. Therefore moving this `graphQLInputType = GraphQLNonNull.nonNull(graphQLInputType);` after `graphQLInputType = list(graphQLInputType)` is wrong because then you mark the OUTER type (the wrapper) as non-null based on the INNER type being non-null. 

I think we should rename `isNotEmpty` to `isWrappedTypeNotNull` and then here we would have this method as follows:
```
private GraphQLInputType createGraphQLInputType(Field field) {
        GraphQLInputType graphQLInputType = referenceGraphQLInputType(field);

        Wrapper wrapper = dataFetcherFactory.unwrap(field, false);
        // Field can have a wrapper, like List<String>
        if (wrapper != null && wrapper.isCollectionOrArrayOrMap()) {
            // Loop as long as there is a wrapper
            do {
                if (wrapper.isCollectionOrArrayOrMap()) {
                    if (wrapper.isWrappedTypeNotNull()) {
                        graphQLInputType = GraphQLNonNull.nonNull(graphQLInputType);
                    }
                    graphQLInputType = list(graphQLInputType);
                    wrapper = wrapper.getWrapper();
                } else {
                    wrapper = null;
                }
            } while (wrapper != null);
        }

        // Check if field is mandatory
        if (field.isNotNull()) {
            graphQLInputType = GraphQLNonNull.nonNull(graphQLInputType);
        }

        return graphQLInputType;
    }
``` 
??? The tests pass for me then. But I may not properly understand the reason for your changes here.

AFAIU there's no concept of not-empty collections in GraphQL. If the inner type is marked as not-null, it means that the collection must not contain null entries, NOT that it must not be empty.

jmartisk (changes requested):

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
jmartisk: Maybe @phillip-kruger remembers details about the `wrapper.isNotEmpty` method and can correct me, but I think it means that the wrapped type is not-null

jmartisk:
Kudos for the idea with `ScopeGroup` and `PolicyGroup`, that's definitely nicer than my original idea with parsing arrays from strings

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/DirectiveTypeCreator.java
```diff
@@ -46,12 +45,36 @@ public DirectiveType create(ClassInfo classInfo) {
         directiveType.setLocations(getLocations(classInfo.declaredAnnotation(DIRECTIVE)));
         directiveType.setRepeatable(classInfo.hasAnnotation(Annotations.REPEATABLE));
 
+        Class<?> directiveClass = null;
+        try {
+            ClassLoader loader = Thread.currentThread().getContextClassLoader();
```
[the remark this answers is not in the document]
RoKKim: Thanks for the suggestion, fixed with static final variables:
https://github.com/smallrye/smallrye-graphql/commit/a55b15213e72d1e3bce0abe652f981bc5cebfd4c#diff-5270b78e6efcb5bcd4deb9ab5d9532e4e3e8fed016364e918a7bd48accbda320R30

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/DirectiveTypeCreator.java
```diff
@@ -1,17 +1,18 @@
...
 
+import java.util.Collections;
 import java.util.Set;
 import java.util.stream.Stream;
 
-import org.jboss.jandex.AnnotationInstance;
-import org.jboss.jandex.ClassInfo;
-import org.jboss.jandex.DotName;
-import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.*;
```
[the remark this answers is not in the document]
RoKKim: Fixed with:
https://github.com/smallrye/smallrye-graphql/pull/2010/commits/a55b15213e72d1e3bce0abe652f981bc5cebfd4c#diff-5270b78e6efcb5bcd4deb9ab5d9532e4e3e8fed016364e918a7bd48accbda320R11

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/DirectiveTypeCreator.java
```diff
@@ -46,12 +45,36 @@ public DirectiveType create(ClassInfo classInfo) {
...
             DirectiveArgument argument = new DirectiveArgument();
-            argument.setReference(referenceCreator.createReferenceForOperationArgument(method.returnType(), null));
+            Type argumentType;
+            if (RequiresScopes.class.isAssignableFrom(directiveClass) || Policy.class.isAssignableFrom(
+                    directiveClass)) {
+                // For both of these directives, we need to override the argument type to be an array of nested arrays
+                // of strings, where none of the nested elements can be null
+                AnnotationInstance nonNullAnnotation = AnnotationInstance.create(NON_NULL, null,
+                        Collections.emptyList());
+                DotName stringDotName = DotName.createSimple(String.class.getName());
```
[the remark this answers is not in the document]
RoKKim: Fixed with:
https://github.com/smallrye/smallrye-graphql/pull/2010/files#diff-5270b78e6efcb5bcd4deb9ab5d9532e4e3e8fed016364e918a7bd48accbda320R32

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
RoKKim: That makes a lot more sense, thank you for the explanation. I implemented the fixes you mentioned and now the tests pass successfully.
https://github.com/smallrye/smallrye-graphql/pull/2010/commits/f0bd7447f402cc9eb0d8696ffbfad7c4422db107

jmartisk:
I've rebase, squashed the commits  and added one more little fix, hopefully it's all good
After merging, I'd create an issue for fixing the minor issue where the directive arguments are declared to be String:
```
directive @requiresScopes(scopes: [[String!]!]!)
```
This might become a more general fix because we don't have straightforward support for custom scalars in directive arguments

jmartisk (approved):

RoKKim:
> I've rebase, squashed the commits and added one more little fix, hopefully it's all good After merging, I'd create an issue for fixing the minor issue where the directive arguments are declared to be String:
> 
> ```
> directive @requiresScopes(scopes: [[String!]!]!)
> ```
> 
> This might become a more general fix because we don't have straightforward support for custom scalars in directive arguments

I started the implementation of supporting directive [@link](https://specs.apollo.dev/link/v1.0/#@link) as well, which uses [scalar Import](https://specs.apollo.dev/link/v1.0/#Import). It can be either a string or an object so I will have to check the posibility of implementing this either way.

jmartisk:
> I started the implementation of supporting directive [@link](https://specs.apollo.dev/link/v1.0/#@link) as well, which uses [scalar Import](https://specs.apollo.dev/link/v1.0/#Import). It can be either a string or an object so I will have to check the posibility of implementing this either way.

Nice! Let's do it in a separate PR please. Together with some other PRs that are going in right now, it's becoming a bit too much to handle for me.

jmartisk:
Thanks for your contribution, this is awesome

Issue #2011, opened 2024-01-20
Support for Federated tracing


RoKKim:
As per the [documentation](https://www.apollographql.com/docs/federation/metrics/), Federated tracing should be enabled if the request header contains `'apollo-federation-include-trace': 'ftv1'`. The `federation-graphql-java-support` library supports this feature out of the box in the  `com.apollographql.federation.graphqljava.tracing.FederatedTracingInstrumentation` class, which I have added to the list of instruments used.
The primary issue arises with the method [com.apollographql.federation.graphqljava.tracing.FederatedTracingInstrumentation.Options#shouldTrace](https://github.com/apollographql/federation-jvm/blob/745884f8be0042613eb3129d297493cf34a7e171/graphql-java-support/src/main/java/com/apollographql/federation/graphqljava/tracing/FederatedTracingInstrumentation.java#L492). By default, it returns `true` if the `apollo-federation-include-trace` is not defined in `executionInput.getGraphQLContext()`. This means tracing would be enabled even if the client did not specify it. To address this, I have implemented a custom  `com.apollographql.federation.graphqljava.tracing.FederatedTracingInstrumentation.Options#shouldTracePredicate` method, which uses similar logic but returns `false` by default.

Example with the correct `apollo-federation-include-trace` header:

- Query:
  ```graphql
  query Test1 {
     __typename
  }
  ```
- Header:
  ```json
  {
    "apollo-federation-include-trace": "ftv1"
  }
  ```
- Result:
  ```json
  {
    "data": {
      "__typename": "Query"
    },
    "extensions": {
      "ftv1": "GgwIjPewrQYQ8PmrvAMiDAiM97CtBhDw+au8A1jAkCByJmIkCgpfX3R5cGVuYW1lGgdTdHJpbmchQNDvEkj0xBRqBVF1ZXJ5"
    }
  }
  ```

Example with the wrong `apollo-federation-include-trace` header:

- Query:
  ```graphql
  query Test1 {
     __typename
  }
  ```
- Header:
  ```json
  {
    "apollo-federation-include-trace": "ftv2"
  }
  ```
- Result:
  ```json
  {
    "data": {
      "__typename": "Query"
    }
  }
  ```

Not providing the header will also result in tracing not being used.

Similar to #2009, I upgraded the `federation-graphql-java-support` library from version 2.1.1 to 4.3.0, as the tracing logic has undergone some changes.

PR #2012, opened 2024-01-20
Issue federation tracing

RoKKim:
#2011

### DIFF
--- a/server/implementation-servlet/src/main/java/io/smallrye/graphql/entry/http/ExecutionServlet.java
+++ b/server/implementation-servlet/src/main/java/io/smallrye/graphql/entry/http/ExecutionServlet.java
@@ -5,8 +5,7 @@ import java.io.IOException;
 import java.io.StringReader;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
--- a/server/implementation-servlet/src/main/java/io/smallrye/graphql/entry/http/ExecutionServlet.java
+++ b/server/implementation-servlet/src/main/java/io/smallrye/graphql/entry/http/ExecutionServlet.java
@@ -1,10 +1,14 @@
 package io.smallrye.graphql.entry.http;
 
+import static com.apollographql.federation.graphqljava.tracing.FederatedTracingInstrumentation.FEDERATED_TRACING_HEADER_NAME;
+
 import java.io.BufferedReader;
 import java.io.IOException;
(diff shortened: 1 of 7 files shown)


RoKKim:
Commit 955252255e3603327b284c8dd071c957ae123ec4 taken from branch #2010, since we also required the upgraded `federation-graphql-java-support` library.
https://github.com/smallrye/smallrye-graphql/pull/2010#issuecomment-1902254268

server/implementation-servlet/src/main/java/io/smallrye/graphql/entry/http/ExecutionServlet.java
```diff
@@ -136,6 +142,15 @@ private static JsonObject toJsonObject(String jsonString) {
         }
     }
 
+    private Map<String, Object> createContext(HttpServletRequest request) {
+        Map<String, Object> context = new HashMap<>();
```
jmartisk, line 146: - Could you change this to add an entry with key `httpHeaders` and value of type `Map<String, List<String>>` (containing all http headers)? This is how we currently do it in the Quarkus implementation of this, so it would be a good idea to keep it unified: see https://github.com/quarkusio/quarkus/blob/3.7.0.CR1/extensions/smallrye-graphql/runtime/src/main/java/io/quarkus/smallrye/graphql/runtime/SmallRyeGraphQLAbstractHandler.java#L117
- I think the map needs to be concurrent because it then gets passed to operations that might run in parallel and write into it

jmartisk:
I'm not familiar with this at all, so I'll have to ask - is there a reasonably easy way to create a test for it? It would certainly be nice to have, but I'm not sure how hard it would be.

RoKKim:
> I'm not familiar with this at all, so I'll have to ask - is there a reasonably easy way to create a test for it? It would certainly be nice to have, but I'm not sure how hard it would be.

I think there should be, will try to do so. Thank you for all the feedback.

RoKKim:
@jmartisk I have implemented [‎ExecutionServlet::getMetaData](https://github.com/smallrye/smallrye-graphql/pull/2012/files#diff-271603955a1ffe73e940a2b0889ce8f37a35d7935704ca9b7c46115a4667ae79R142)‎ method and [Federated tracing tests](https://github.com/smallrye/smallrye-graphql/pull/2012/files#diff-d431e3986e31f1eee5fb582017ee0649e1aaf917737b1ca4730ca4fb4276bc5d).

jmartisk (approved):
LGTM. Let's wait for @mskacelik who is going to create a little test for this on the Quarkus side

jmartisk:
Thanks! This will go into smallrye-grapql 2.8 (I think it warrants a minor version bump, it's quite a major new feature)

PR #2013, opened 2024-01-24
Fix Throwable class name with wildcard

rubik-cube-man:
Fixes an issue where is a configured value ends in a wildcard, the character before the wildcard is ignored.

Previously `java.lang*` would match with `java.land.SomeException`.

This also fixes a standalone wildcard (`*`) throwing an exception as it tries to do a substring of index 0 to -1.


t1:
LGTM.

Shouldn't there be tests for this?

jmartisk:
Thanks!

PR #2017, opened 2024-01-28
Deprecated annotation added to IndexInitializer

RoKKim:
The indexer is missing the `Deprecated.class`, and as a result, the schema returned from the `_service` query does not include it.


jmartisk (approved):
Thanks

PR #2019, opened 2024-01-29
Issue deprecated index

RoKKim:
Added imports for `Deprecated` and `Override` to make sure that Java native interfaces won't be used.


jmartisk (approved):
Eerm, not sure how I missed that.. Yeah, thanks for noticing.

PR #2026, opened 2024-02-04
Optimise by reducing calls to ServiceLoader

rubik-cube-man:
I decided this weekend to profile some of our programs and discovered that there seemed to be high CPU usage coming from ServiceLoaders. Below is a screenshot of the profile that I took. (One of the vert.x event loop threads)

ServiceLoaders

More than half the CPU was going towards loading these services multiple times.

This change replaces most of (but not all) the unnecessary ServiceLoader calls.

### Json
Inside the Json class, it gets the provider, which makes a call to the ServiceLoader. This can be resolved using Factories instead, as they can be reused.
There are still quite a few Json.createValue()'s inside the client, which seems pretty slow. The Json library doesn't seem to have a factory for this at the moment, so I have just left it how it is for now.

### MetricsEmitter & EventEmitter

Both of these classes seemed to be used in several places, and as the getInstance() method on both of these doesn't return a single instead but instead creates a new instance each time, converting these to static values proved to be performant.
From my poking, it seems that these implementations were developed to support having a single instance accessed across multiple threads, however, I might be mistaken.

I found an instance where these classes weren't being used, so I removed them. There was also an instance inside AbstractDataFetcher where the MetricsEmitter wasn't being used, but removing it broke Quarkus as it seems to use that. I've just added a variable back so that Quarkus can continue to be happy for now.

### Closing Readers

There were a few JsonReaders that weren't being closed. I traced the code, and this doesn't fix any sort of memory leak (which aligns with my heap dumps). However, this fixes the problem of the char arrays not being recycled and being garbage collected, which will offer a minor performance improvement.

### Logger

I also replaced a non-static logger with a static logger to save it from needing to be reconstructed with every new object.

## Testing

Out of curiosity, I also deployed this to our development servers and saw some pretty decent performance improvements:

image
_(Graph of CPU usage of our Quarkus application running GraphQL, left is the old code and the right is the new code)_

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/event/EventEmitter.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/event/EventEmitter.java
@@ -37,10 +37,11 @@ import io.smallrye.graphql.spi.config.Config;
  */
 public class EventEmitter {
     private static final Logger LOG = Logger.getLogger(EventEmitter.class);
+    private static final ThreadLocal<EventEmitter> eventEmitters = ThreadLocal.withInitial(EventEmitter::new);
     private final List<EventingService> enabledServices;
 
     public static EventEmitter getInstance() {
-        return new EventEmitter();
+        return eventEmitters.get();
     }
 
     private EventEmitter() {
--- a/client/api/src/main/java/io/smallrye/graphql/client/websocket/WebsocketSubprotocol.java
+++ b/client/api/src/main/java/io/smallrye/graphql/client/websocket/WebsocketSubprotocol.java
@@ -5,7 +5,9 @@ public enum WebsocketSubprotocol {
     GRAPHQL_WS("graphql-ws"),
     GRAPHQL_TRANSPORT_WS("graphql-transport-ws");
 
-    private String protocolId;
+    private static final WebsocketSubprotocol[] VALUES = values();
+
+    private final String protocolId;
 
     WebsocketSubprotocol(String protocolId) {
         this.protocolId = protocolId;
@@ -16,7 +18,7 @@ public enum WebsocketSubprotocol {
     }
 
     public static WebsocketSubprotocol fromString(String text) {
-        for (WebsocketSubprotocol b : WebsocketSubprotocol.values()) {
+        for (WebsocketSubprotocol b : VALUES) {
             if (b.protocolId.equalsIgnoreCase(text)) {
                 return b;
             }
--- a/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
+++ b/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
@@ -315,7 +315,7 @@ class VertxTypesafeGraphQLClientProxy {
     }
 
     private JsonObjectBuilder variables(MethodInvocation method) {
-        JsonObjectBuilder builder = Json.createObjectBuilder();
+        JsonObjectBuilder builder = jsonObjectFactory.createObjectBuilder();
         method.valueParameters().forEach(parameter -> builder.add(parameter.getRawName(), value(parameter.getValue())));
         return builder;
     }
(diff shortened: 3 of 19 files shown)


phillip-kruger:
@jmartisk can you have a look ? I am wondering about the wildfly use case ?

jmartisk:
Yeah, storing emitters in static fields will break multi-deployment applications in WildFly, because the framework classes are loaded by a single class loader for all deployments, and thus values of static fields will be shared. 

If we want to use a static field, it should be wrapped into
```
private static final Map<ClassLoader, WHAT_YOU_WANT_TO_STORE> INSTANCES = new WeakHashMap<>();
```

and then store the objects in it, with the key being the thread context class loader.

rubik-cube-man:
Yeah, I can do! I've got a few things I need to tackle at work, so I will be able to take a look later this week!

rubik-cube-man:
So, I decided to implement the MetricsEmitter and EventEmitter's getInstance method with a ThreadLocal. I was concerned that WeakHashMap could struggle with the concurrency of large applications. This will have a larger memory footprint, but hopefully, this will be worth the CPU reductions.

Whilst poking around with these changes, I made a few more tweaks:
- [Forcing metrics to end even in an error](https://github.com/smallrye/smallrye-graphql/pull/2026/files#diff-5bf9b18cf9f2e6230ac2891288d5f91f16b656127cdd8fea2047919e9111e7e9R65). Since previously it didn't matter if it ended or not as the MetricsEmitter was just GCed, now that the MetricsEmitter is persisted ending the metrics is far more important.
- [Reducing calls to values()](https://github.com/smallrye/smallrye-graphql/pull/2026/files#diff-794f0ca582f23d5c37adfd14c56dc96f7a89f923e0332dbe079a7810bb354e5fR28). Since values() inside enums creates a whole new array, I was able to see a few of these methods in my profiles. This can easily be solved by calling values() once.
- [Construct the DataLoaderRegistry once](https://github.com/smallrye/smallrye-graphql/pull/2026/files#diff-33bf1ec886c73ea51d9e3bb9708cd51f9e093867e8739d37b3507558a435d5e7R93). This was a pretty expensive operation and looked like it didn't particularly hold any state, so by moving it to the constructor I was able to reduce the number of calls to it.

server/implementation/src/main/java/io/smallrye/graphql/execution/event/EventEmitter.java
```diff
@@ -37,10 +37,11 @@
  */
 public class EventEmitter {
     private static final Logger LOG = Logger.getLogger(EventEmitter.class);
+    private static final ThreadLocal<EventEmitter> eventEmitters = ThreadLocal.withInitial(EventEmitter::new);
```
jmartisk, line 40: Hmm.  I'm not feeling particularly safe about this. 
- Are you sure this works properly with reactive pipelines?
- I think we need a mechanism for clearing the value, no? In WildFly, an application can get undeployed. If we don't have a hook in place to clear thread locals during undeploy, this will result in a memory leak. But right, we haven't been putting too much effort in supporting multi-deployment scenarios with WildFly, so maybe it's not that much of a problem.

It's certainly not safe for multi-deployment scenarios, but if it at least has a good performance and doesn't break WildFly completely, then we might as well just go with it.

rubik-cube-man: I just ran a test inside my Quarkus application, added in all the events, and successfully listened to them. From looking at the event code, it's pretty stateless, so it looks like it's fine to reuse the object for multiple event calls.

The MetricsServices, on the other hand are stateful, and as I fixed in they must have the "end" called, otherwise memory will leak as metric points would just be left open otherwise. The other occurrence of this I know is inside Quarkus [here](https://github.com/quarkusio/quarkus/blob/main/extensions/smallrye-graphql/runtime/src/main/java/io/quarkus/smallrye/graphql/runtime/spi/datafetcher/AbstractAsyncDataFetcher.java#L51). I can open a PR to close on an exception here if you wish?

I have assumed here that it is fine to track exceptions against the timings of data points. I can happily change it to just remove the data point so exceptions don't count towards timings.

 Regarding removing from the ThreadLocal, assuming that Wildfly removes the associated threads from an application when it unloaded, all the values should be removed from ThreadLocal automatically. This is what is mentioned in the docs:
 > after a thread goes away, all of its copies of thread-local instances are subject to garbage collection (unless other references to these copies exist).

jmartisk: Ok let's not dive too much into WildFly potential memory leaks, most SmallRye projects don't even claim they support any kind of multi-deployment scenario, including undeploying and redeploying the same app, because we are focused on microservices, not traditional application servers. 

Yeah I think we count "exceptional" executions towards the metrics normally just like the successful ones. If there's some cases where a metric measurement isn't ended, that should be fixed

rubik-cube-man: Awesome! I'll assemble a patch for Quarkus to ensure that the metrics are always ended!

rubik-cube-man: Have created a fix here: https://github.com/quarkusio/quarkus/pull/38686

jmartisk (approved):
Thanks!

Issue #1993, opened 2023-12-22
Parametrized types in the names of the Type in the schema.


mskacelik:
#1986

mskacelik:
gonna have a look at it

PR #2027, opened 2024-02-05
change how the type variables are handled during the creation of the reference

mskacelik:
fix: #1993
/cc @jmartisk 

WIP

The problem in the ReferenceCreator class was that it did not differentiate between the `SomeClass<A>` and `SomeClass extends SomeOtherClass<A>` type variables. These type variables are stored in a map, which will be used in the schema generation process.

So, I have made a new field in the `Reference` class for storing parent's type variables.

I also added tests in the SchemaBuilderTest, which are also WIP.

Tests seems to work fine, but there seems to be a problem from the Quarkus app side:
```
Error running Quarkus: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at io.quarkus.runner.bootstrap.StartupActionImpl$1.run(StartupActionImpl.java:113)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.ExceptionInInitializerError
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:70)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:44)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:124)
	at io.quarkus.runner.GeneratedMain.main(Unknown Source)
	... 6 more
Caused by: java.lang.RuntimeException: Failed to start quarkus
	at io.quarkus.runner.ApplicationImpl.<clinit>(Unknown Source)
	... 15 more
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.Map.put(Object, Object)" because the return value of "io.smallrye.graphql.schema.model.Reference.getAllParametrizedTypes()" is null
	at io.quarkus.deployment.steps.SmallRyeGraphQLProcessor$buildExecutionService1691419614.deploy_2(Unknown Source)
	at io.quarkus.deployment.steps.SmallRyeGraphQLProcessor$buildExecutionService1691419614.deploy(Unknown Source)
	... 16 more
```
reproduced by:
```java
static class Bar {
    int a;
    public Bar() {}
}
static class Foo<T> extends Entity<Bar> {
        T t;
        public Foo() {
        }
    }

    static class Entity<N> {
        private N b;

        public Entity() {
        }
    }
    @Query
    public Foo<String> getFoo() {
        return null;
    }
```

### DIFF
--- a/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Reference.java
+++ b/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Reference.java
@@ -1,9 +1,12 @@
 package io.smallrye.graphql.schema.model;
 
 import java.io.Serializable;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Represents a reference to some other type (type/input/enum/interface) This so that, as we are scanning, we can refer
@@ -20,6 +23,7 @@ public class Reference implements Serializable {
     private AdaptTo adaptTo = null; // If the type is mapped to a scalar
     private AdaptWith adaptWith = null; // If the type is mapped to another type with an adapter
     private Map<String, Reference> classParametrizedTypes;
+    private Map<String, Reference> parentClassParametrizedTypes;
     private boolean addParametrizedTypeNameExtension;
     private List<DirectiveInstance> directiveInstances;
 
@@ -46,6 +50,7 @@ public class Reference implements Serializable {
         this.adaptTo = builder.adaptTo;
         this.adaptWith = builder.adaptWith;
         this.classParametrizedTypes = builder.classParametrizedTypes;
+        this.parentClassParametrizedTypes = builder.parentClassParametrizedTypes;
         this.addParametrizedTypeNameExtension = builder.addParametrizedTypeNameExtension;
         this.directiveInstances = builder.directiveInstances;
         this.wrapper = builder.wrapper;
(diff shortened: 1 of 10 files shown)


common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Reference.java
```diff
@@ -134,6 +138,29 @@ public boolean isAdaptingWith() {
         return this.adaptWith != null;
     }
 
+    public Map<String, Reference> getAllParametrizedTypes() {
+        if (classParametrizedTypes == null) {
+            return extendedClassParametrizedTypes; // if both are 'null', then it returns null
```
jmartisk, line 143: Why should return null here, how about we just combine the two maps, possibly resulting in an empty map?
Neither of the two maps should be null, we should make sure both are at least empty.

mskacelik:
By doing this issue I also find a possible bug.
When defining a class similar to this:
```java
// -------
class SomeClass<T> extends OtherClass<String> {
     T field1;
    //  other fields/methods
}

class SomeOtherClass extends OtherClass<String>  {
    int field;
    // other fields/methods
}

class OtherClass<N> {
    N field2;
    // other fields/methods
}
// -------
```
The `ReferenceCreator#getReference` checks two things. One is if the given class is a parameterized type. 
If so than it saves the Type Variable reference only for its definition and does not look at the Parent's type variables.

So lets say I have Query like this:
```java
@Query
public SomeClass<Integer> someQuery() { /*...*/ }
```
It saves the `T -> Integer.` How ever it does not saves the `N -> String`.

On the other hand, if the given type is not parametrized but is a normal Class type, it saves the Parent.
So something like this is happening:
```java
@Query
public SomeOtherClass someQuery() { /*...*/ }
```
In this case, `SomeOtherClass` is not a parametrized class, so it saves the `N -> String` (which is fine).

# possible solution
I could fix it by that the Parametrized Type also checks for the parent Type Vars. This could work but there could be a problem that the parent class definition also uses the same Type Variable naming.
For example:
```java
class SomeClass<T> extends OtherClass<String> {
     T field1;
    //  other fields/methods
}

class OtherClass<T> {
    T field2;
    // other fields/methods
}
```
and query be:

```java
@Query
public SomeClass<Integer> someQuery() { /*...*/ }
```
Which will save `T -> Integer` and `T -> String` when merging the type variable references. Which will be a problem.
Plus, there also might be some cases like `T -> T`:
```java
class SomeClass<T> extends OtherClass<T> { //... }
```
Gonna have a look at it in a separate issue...

jmartisk (approved):

Issue #2020, opened 2024-01-31
Mapping directive arguments to custom scalar types


jmartisk:
For example, we have the federation-related `@RequiresScopes` whose argument should be `[[Scope!]!]!` where Scope is a string-based scalar, but we have it as `[[String!]!]!`

jmartisk:
fixed by https://github.com/smallrye/smallrye-graphql/pull/2028

PR #2028, opened 2024-02-05
Mapping directive arguments to custom scalar types

RoKKim:
#2020
Added the following scalars:
```graphql
"Scalar for FieldSet"
scalar FieldSet

"Scalar for Policy"
scalar Policy

"Scalar for Scope"
scalar Scope
```
This is a breaking change, since the usage of the following Federation directives is changed:

- `@Key`
- `@Provides`
- `@Requires`
- `@Policy`
- `@RequiresScopes`

For example, this:
```graphql
@Key(fields = "sku package")
```
Changes to this:
```graphql
@Key(fields = @FieldSet("sku package"))
```

Does this need to be a breaking change? Technically, we could change the following methods:

- io.smallrye.graphql.schema.creator.DirectiveTypeCreator#create
- io.smallrye.graphql.schema.helper.Directives#toDirectiveInstance

To detect if any of the above directives are used, and convert them to appropriate scalars. All of them are string based and are easy to convert. We would probably need to fix some other stuff as well, but I find the first option to be much cleaner. This is also aligned with how `@Link` will be implemented, since its [Import](https://specs.apollo.dev/link/v1.0/#@link.import) type is not string based.

Here is how outputted schema looks like:
```graphql
directive @key(fields: FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE

type Product @key(fields : "sku package") @policy(policies : [["policy1", "policy2"], ["policy3"]]) @requiresScopes(scopes : [["scope1", "scope2"], ["scope3"]]) {
...
}

"Scalar for FieldSet"
scalar FieldSet
```

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/Directives.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/Directives.java
@@ -2,10 +2,11 @@ package io.smallrye.graphql.schema.helper;
 
 import static java.util.stream.Collectors.toList;
 import static org.jboss.jandex.AnnotationValue.Kind.ARRAY;
+import static org.jboss.jandex.AnnotationValue.Kind.NESTED;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -74,8 +75,7 @@ public class Directives {
             if (directiveType.getClassName().equals(Policy.class.getName()) ||
                     directiveType.getClassName().equals(RequiresScopes.class.getName())) {
                 // For both of these directives, we need to process the annotation values as nested arrays of strings
-                List<List<String>> valueList = processAnnotationValues((AnnotationValue[]) annotationValue.value());
-                directiveInstance.setValue(annotationValue.name(), valueList);
+                directiveInstance.setValue(annotationValue.name(), valueObjectNestedList(annotationValue));
             } else {
                 directiveInstance.setValue(annotationValue.name(), valueObject(annotationValue));
             }
@@ -84,35 +84,42 @@ public class Directives {
         return directiveInstance;
     }
 
-    private Object valueObject(AnnotationValue annotationValue) {
-        if (annotationValue.kind() == ARRAY) {
-            AnnotationValue[] values = (AnnotationValue[]) annotationValue.value();
-            Object[] objects = new Object[values.length];
-            for (int i = 0; i < values.length; i++) {
-                objects[i] = valueObject(values[i]);
+    private Object valueObject(AnnotationValue value) {
+        if (value.kind() == ARRAY) {
+            AnnotationValue[] annotationValues = (AnnotationValue[]) value.value();
+            Object[] objects = new Object[annotationValues.length];
+            for (int i = 0; i < annotationValues.length; i++) {
+                objects[i] = valueObject(annotationValues[i]);
             }
             return objects;
+        } else if (value.kind() == NESTED) {
+            AnnotationInstance annotationInstance = (AnnotationInstance) value.value();
+            Map<String, Object> values = new LinkedHashMap<>();
+            if (annotationInstance != null) {
+                for (AnnotationValue annotationValue : annotationInstance.values()) {
+                    values.put(annotationValue.name(), valueObject(annotationValue));
+                }
+            }
+            return values;
         }
-        return annotationValue.value();
+        return value.value();
+    }
+
+    // This method is used specifically in cases where we wish to ignore upper nested objects (e.g. PolicyGroup)
+    private List<List<Object>> valueObjectNestedList(AnnotationValue value) {
+        valueObject(value);
+        List<List<Object>> values = new ArrayList<>();
+        for (AnnotationValue annotationValue : value.asArrayList()) {
+            List<Object> nestedValues = new ArrayList<>();
+            for (AnnotationValue nestedAnnotationValue : annotationValue.asNested().values().get(0).asArrayList()) {
+                nestedValues.add(valueObject(nestedAnnotationValue));
+            }
+            values.add(nestedValues);
+        }
+        return values;
     }
 
     public Map<DotName, DirectiveType> getDirectiveTypes() {
         return directiveTypes;
-    }
-
-    private List<List<String>> processAnnotationValues(AnnotationValue[] annotationValues) {
-        if (annotationValues == null || annotationValues.length == 0) {
-            return Collections.emptyList();
-        }
-
-        List<List<String>> valuesList = new ArrayList<>();
-        for (AnnotationValue nestedValue : annotationValues) {
-            List<String> values = new ArrayList<>();
-            for (AnnotationValue value : (AnnotationValue[]) nestedValue.asNested().values().get(0).value()) {
-                values.add(value.asString());
-            }
-            valuesList.add(values);
-        }
-        return valuesList;
     }
 }
--- a/common/schema-model/pom.xml
+++ b/common/schema-model/pom.xml
@@ -10,4 +10,12 @@
     <artifactId>smallrye-graphql-schema-model</artifactId>
     <name>SmallRye: GraphQL Common :: Schema Model</name>
     <description>A model that represents the schema</description>
+
+     <dependencies>
+        <!-- The API -->
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>smallrye-graphql-api</artifactId>
+        </dependency>
+    </dependencies>
 </project>
--- a/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Scalars.java
+++ b/common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Scalars.java
@@ -28,6 +28,10 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
+import io.smallrye.graphql.api.federation.FieldSet;
+import io.smallrye.graphql.api.federation.policy.PolicyItem;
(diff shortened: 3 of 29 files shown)


common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/Directives.java
```diff
@@ -100,16 +100,16 @@ public Map<DotName, DirectiveType> getDirectiveTypes() {
         return directiveTypes;
     }
 
-    private List<List<String>> processAnnotationValues(AnnotationValue[] annotationValues) {
+    private List<List<Object>> processAnnotationValues(AnnotationValue[] annotationValues) {
```
RoKKim, line 103: This method is changed to handle any type of object, not just the string.

common/schema-model/pom.xml
```diff
@@ -10,4 +10,12 @@
...
     <name>SmallRye: GraphQL Common :: Schema Model</name>
     <description>A model that represents the schema</description>
+
+     <dependencies>
+        <!-- The API -->
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>smallrye-graphql-api</artifactId>
+        </dependency>
+    </dependencies>
```
RoKKim, line 20: I had to add this in order to access Federation directives interface classes. If this is not okay, we can just hardcode the class names.

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Scalars.java
```diff
@@ -181,6 +188,12 @@ public static void addUuid() {
         // Void
         populateScalar(Void.class.getName(), VOID, Void.class.getName());
         populateScalar(void.class.getName(), VOID, Void.class.getName());
+
+        if (Boolean.getBoolean("smallrye.graphql.federation.enabled")) {
```
RoKKim, line 192: I wasn't sure if we can use `Config.get().isFederationEnabled()` here.

RoKKim:
Will fix the tests.

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Scalars.java
jmartisk: Correct, MP Config is not available here, this module executes during Quarkus build time.

jmartisk:
Thanks so much for this! I'll look into it asap.
As for the breaking change, I think it's not much of a problem. We will do it with a minor version bump, and the Federation support is marked as experimental, so breaking is expected. It also looks more typesafe and cleaner when you don't magically change a String into something else

server/implementation/pom.xml
```diff
@@ -26,6 +26,12 @@
             <artifactId>smallrye-graphql-schema-model</artifactId>
         </dependency>
 
+        <!-- Jandex indexer -->
+        <dependency>
```
jmartisk, line 30: We shouldn't use Jandex in the runtime module... Unless it's absolutely necessary.
I don't understand why exactly you need this? Can the Coercing classes somehow work with instances of the annotations themselves instead of the Jandex object?

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Scalars.java
```diff
@@ -181,6 +188,11 @@ public static void addUuid() {
         // Void
         populateScalar(Void.class.getName(), VOID, Void.class.getName());
         populateScalar(void.class.getName(), VOID, Void.class.getName());
+
+        // Federation
+        populateScalar(FieldSet.class.getName(), FIELD_SET, FieldSet.class.getName());
+        populateScalar(PolicyItem.class.getName(), POLICY, PolicyItem.class.getName());
+        populateScalar(ScopeItem.class.getName(), SCOPE, ScopeItem.class.getName());
```
RoKKim, line 195: I had to remove the `Boolean.getBoolean("smallrye.graphql.federation.enabled")` check here because variable `io.smallrye.graphql.schema.model.Scalars#scalarMap` is static. When running tests inside `server/implementation/src/test/java/io/smallrye/graphql`, the `Scalars` class is initialized without federation classes, since federation is disabled by default. However, when running `io.smallrye.graphql.schema.SchemaTest#testSchemaWithFederationEnabled`, federation is enabled, but the `scalarMap` remains unchanged.

server/implementation/src/main/java/io/smallrye/graphql/scalar/GraphQLScalarTypes.java
```diff
@@ -97,6 +100,11 @@ public static void registerCustomScalar(
 
         mapType(new VoidScalar()); // Void
 
+        // Federation
+        mapType(new FieldSetScalar());
+        mapType(new PolicyScalar());
+        mapType(new ScopeScalar());
+
```
RoKKim, line 107: Same reason for removing `Config.get().isFederationEnabled()` as in `io.smallrye.graphql.schema.model.Scalars`.

server/implementation/pom.xml
RoKKim: I fixed `io.smallrye.graphql.schema.helper.Directives` to avoid converting to `AnnotationInstances`.
ca0fd64b79a8cd71c16fbcc6642ff489439225e8

RoKKim:
@jmartisk, WildFly and Quarkus tests fail:
```
2024-02-07T11:23:28.5188034Z [ERROR] /home/runner/work/smallrye-graphql/smallrye-graphql/quarkus/extensions/smallrye-graphql/deployment/src/test/java/io/quarkus/smallrye/graphql/deployment/federation/complex/Foo.java:[7,15] annotation value must be an annotation
2024-02-07T11:23:28.5190863Z [ERROR] /home/runner/work/smallrye-graphql/smallrye-graphql/quarkus/extensions/smallrye-graphql/deployment/src/test/java/io/quarkus/smallrye/graphql/deployment/federation/base/Foo.java:[7,15] annotation value must be an annotation
2024-02-07T11:23:28.5193899Z [ERROR] /home/runner/work/smallrye-graphql/smallrye-graphql/quarkus/extensions/smallrye-graphql/deployment/src/test/java/io/quarkus/smallrye/graphql/deployment/federation/batch/Foo.java:[7,15] annotation value must be an annotation
2024-02-07T11:23:28.5197261Z [ERROR] /home/runner/work/smallrye-graphql/smallrye-graphql/quarkus/extensions/smallrye-graphql/deployment/src/test/java/io/quarkus/smallrye/graphql/deployment/federation/complex/Bar.java:[7,15] annotation value must be an annotation
2024-02-07T11:23:28.5199760Z [ERROR] /home/runner/work/smallrye-graphql/smallrye-graphql/quarkus/extensions/smallrye-graphql/deployment/src/test/java/io/quarkus/smallrye/graphql/deployment/federation/complex/Bar.java:[8,15] annotation value must be an annotation
2024-02-07T11:23:28.5202341Z [ERROR] /home/runner/work/smallrye-graphql/smallrye-graphql/quarkus/extensions/smallrye-graphql/deployment/src/test/java/io/quarkus/smallrye/graphql/deployment/federation/complex/Bar.java:[7,1] annotation value must be an annotation
```
Since `Key#fields` is no longer a string, it must be wrapped in `FieldSet`."

jmartisk:
Yeah I'll prepare the necessary commits for the WildFly and Quarkus side and have another look at your PR, hopefully I'll get enough time tomorrow

jmartisk:
Ok this will be needed on the wildfly/quarkus side once we merge this PR:
 https://github.com/jmartisk/wildfly-graphql-feature-pack/tree/srgql-2.8.0 
and
https://github.com/jmartisk/quarkus/tree/srgql2.8.0-federation-annotations-change

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Scalars.java
jmartisk: If you remove it, then these directives will always be in the schema. I'm not sure I understand your explanation for removing it - why exactly? Is it that `io.smallrye.graphql.schema.SchemaTest#testSchemaWithFederationEnabled` doesn't see them even though it sets `smallrye.graphql.federation.enabled=true` at the beginning? I wonder why this was working well before and now it isn't?!

jmartisk: Ah right, now I understand your point. This is a bit tricky because the static initializer will be run just once, unless we run the test with a separate JVM or class loader. Let me think...

jmartisk: Could you split `io.smallrye.graphql.schema.SchemaTest` into two, where you introduce `io.smallrye.graphql.schema.SchemaTestFederationEnabled` that just contains the `testSchemaWithFederationEnabled` method (plus whatever it uses), and make sure that each test runs in a separate JVM by adding `reuseForks=false` into the surefire plugin's configuration (for the `server/implementation` module)?

```
<plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <configuration>
                    <useFile>false</useFile>
                    <trimStackTrace>false</trimStackTrace>
                    <reuseForks>false</reuseForks>
                </configuration>
            </plugin>
        </plugins>
```

Then you can reintroduce the `if (Boolean.getBoolean("smallrye.graphql.federation.enabled"))` condition in the `Scalars` static initializer

RoKKim: Adding `reuseForks=false` worked perfectly, thanks for the idea! Inside `GraphQLScalarTypes`, I also had to change the Federation check to `Boolean.getBoolean("smallrye.graphql.federation.enabled")`, since it seems to be used by the `smallrye-graphql-maven-plugin` when generating the schema.

jmartisk:
Looks good. Please don't push anything now, gonna squash and resolve some conflicts before merging

jmartisk:
.. and send out PRs with the WF and Quarkus changes

jmartisk (approved):

PR #2033, opened 2024-02-07
Added label argument to @override directive

RoKKim:
With Apollo Federation 2.7 optional `label` argument was added to `@override` directive:
https://www.apollographql.com/docs/federation/federated-types/federated-directives/#label


jmartisk (approved):
thanks!

PR #2040, opened 2024-02-10
Support for @link directive

RoKKim:
I have implemented the `@link` directive as per the documentation:

1. https://specs.apollo.dev/link/v1.0/#Import
2. https://www.apollographql.com/docs/federation/federated-types/federated-directives/#importing-directives

Currently, I have made the usage of the directive optional, meaning if the `@link` directive is not defined in the schema, no renaming and checks will be performed. Of course, if needed, we can make the usage of the `@link` directive mandatory.

I have added all the test scenarios I could think of to the `io.smallrye.graphql.schema.SchemaLinkTest`.


common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/DirectiveTypeCreator.java
```diff
@@ -79,8 +79,8 @@ public DirectiveType create(ClassInfo classInfo) {
                 argumentType = method.returnType();
             }
             argument.setReference(referenceCreator.createReferenceForOperationArgument(argumentType, null));
-            argument.setName(method.name());
             Annotations annotationsForMethod = Annotations.getAnnotationsForInterfaceField(method);
+            argument.setName(TypeNameHelper.getMethodName(method, annotationsForMethod));
```
RoKKim, line 83: This is done due to `io.smallrye.graphql.api.federation.link.Link`, as we cannot use 'import' and 'for' as variable names in Java.
```java
@Name("import")
Import[] _import() default {};

@Name("for")
Purpose _for() default Purpose.UNDEFINED;
```
```graphql
directive @link(url: String!, as: String, import: [Import], for: Purpose) repeatable on SCHEMA
```

common/schema-builder/src/main/java/io/smallrye/graphql/schema/helper/TypeNameHelper.java
```diff
@@ -119,6 +120,18 @@ private static String getNameForClassType(ClassInfo classInfo, Annotations annot
         return sb.toString();
     }
 
+    public static String getMethodName(MethodInfo methodInfo, Annotations annotations) {
+        StringBuilder sb = new StringBuilder();
+
+        if (annotations.containsKeyAndValidValue(Annotations.NAME)) {
+            sb.append(annotations.getAnnotationValue(Annotations.NAME).asString().trim());
```
RoKKim, line 127: Currently we only consider `@Name` annotation when renaming argument, but we could later on add some more cases if needed.

server/api/src/main/java/io/smallrye/graphql/api/federation/link/Link.java
```diff
@@ -0,0 +1,53 @@
...
+    @NonNull
+    @Description("The foreign schema’s URL.")
+    String url();
+
+    @Description("Change the namespace prefix assigned to the foreign schema.")
+    String as() default "";
+
+    @Description("A list of elements to import from the foreign schema into the document.")
+    @Name("import")
+    Import[] _import() default {};
```
RoKKim, line 42: `import` cannot be used as a variable name in Java, so the `@Name` annotation has been added instead.

server/api/src/main/java/io/smallrye/graphql/api/federation/link/Link.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Description("An optional purpose for this link. This hints to consumers as to whether they can safely ignore " +
+            "metadata described by a foreign schema.")
+    @Name("for")
+    Purpose _for() default Purpose.UNDEFINED;
```
RoKKim, line 47: `for` cannot be used as a variable name in Java, so the `@Name` annotation has been added instead.

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -133,6 +135,7 @@ public static GraphQLSchema bootstrap(Schema schema, boolean skipInjectionValida
 
     private Bootstrap(Schema schema, boolean skipInjectionValidation) {
         this.schema = schema;
+        this.linkProcessor = new LinkProcessor(schema);
```
jmartisk, line 138: I'd say this should be wrapped into 
`if (Config.get().isFederationEnabled()) {`
and stay null if federation is disabled, no?

jmartisk: And then all usages of the `linkProcessor` should similary be wrapped into a condition.

jmartisk: Hmm, I see you have `Config.get().isFederationEnabled()` inside the `LinkProcessor` class but it would still be better to avoid using that class altogether, for optimization purposes (it can be eliminated as dead code by GraalVM)

RoKKim: In `Bootstrap`, I also use `linkProcessor.newNameDirective` and `linkProcessor.newName`. If `LinkProcessor` is not defined, this operation will fail, so they should probably be checked as well.

jmartisk: Yes, I would wrap usage of `linkProcessor` into an `if`. And only call it when Federation is enabled.
The `LinkProcessor` itself won't have to check whether Federation is enabled anymore, it can assume that it is enabled.

RoKKim: This should be fixed now.

jmartisk (dismissed):
Very nice, thanks a lot!!

jmartisk:
Ah, I see there are some conflicts. Could you please rebase against main and resolve them? Then I think we'll be ready to merge

jmartisk:
(Also I'd prefer if you squashed all the commits into just one, but if you don't, I can do it myself)

jmartisk (dismissed):

MightyPixel:
Hi @RoKKim / @jmartisk  thank you for adding this.

Does this feature allows generating schema with the following line?
`extend schema @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@key", "@interfaceObject"])`

RoKKim:
> Hi @RoKKim / @jmartisk thank you for adding this.
> 
> Does this feature allows generating schema with the following line? `extend schema @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@key", "@interfaceObject"])`
The provided example would result in "Federation v2.4 feature @interfaceObject imported using old Federation v2.3 version
"

For the provided example, you would currently need to add this annotation:
```java
@GraphQLApi
@Link(url = FEDERATION_SPEC_URL + "/v2.3", _import = {
        @Import(name = "@key"),
        @Import(name = "@interfaceObject")
})
```
Which would result in the following schema:
```graphql
schema @link(import : ["@key", "@interfaceObject"], url : "https://specs.apollo.dev/federation/v2.3")
```

MightyPixel:
Thanks @RoKKim! I do not see the `@Link` in the `2.7.0` release - is it planned to be released in the next version?

PR #2049, opened 2024-03-02
Fixed Federation version of @interfaceObject

RoKKim:
As per [documentation](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#interfaceobject), the minimum version that supports the `@interfaceObject` directive is 2.3, not 2.4,


jmartisk (approved):

jmartisk:
Thanks

PR #2055, opened 2024-03-09
fix #2050: allow ErrorOr to contain a null value

### LINKED ISSUE #2050: `ErrorOr` does not support null value and ignorable errors
I'm using the type-safe API and trying to use `ErrorOr`. In my case, the query can return a `null` value (if not found). Sometimes it also returns an error on some random (nested) field, which I would like to log.

When using `ErrorOr` I get an exception `value must not be null`, because ErrorOr uses `requireNonNull`. It also does not allow access to the value if there are any errors.

Since this class really assumes there is either a value or an error, I'm hesitant to patch it to allow null values. It would probably be better to create a new type that allows access to both without making assumptions.

t1:


jmartisk:
I think you've created this wrong, there are no changes contained in this PR

jmartisk:
I'd assume you wanted to create it from this branch: https://github.com/t1/smallrye-graphql/tree/2050-erroror-null
but the PR is from your main

t1:
oops... sorry, by bad. I've created #2056

PR #2061, opened 2024-03-13
Fix overly ambitious caching when using nested queries

rubik-cube-man:
Well, this is awkward, it would seem that I have been too overly ambitious with my caching in my previous pull request!

This fixes the result of nested queries being cached when with the exact same variables.

As I am busy at the moment at work at the moment, I have just reverted back to its previous state (https://github.com/rubik-cube-man/smallrye-graphql/commit/0ad787459b9fdde120563dce68ff1c7306a30359#diff-33bf1ec886c73ea51d9e3bb9708cd51f9e093867e8739d37b3507558a435d5e7L74)

Apologies for that!

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionService.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionService.java
@@ -68,7 +68,6 @@ public class ExecutionService {
     private final BatchLoaderHelper batchLoaderHelper = new BatchLoaderHelper();
     private final DataFetcherFactory dataFetcherFactory = new DataFetcherFactory();
     private final Schema schema;
-    private final DataLoaderRegistry dataLoaderRegistry;
 
     private final EventEmitter eventEmitter = EventEmitter.getInstance();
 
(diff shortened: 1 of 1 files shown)


rubik-cube-man:
Out of curiosity, would there be any opposition to me adding nightly builds? It would mean that I could start testing the latest changes on our dev servers before a new version drops! Currently, I manually deploy the latest SmallRye GraphQL from my computer. However, this slowly returns to the latest release as we redeploy our applications via CI/CD.

t1:
Where would those nightly builds be deployed to? Maven Central?

jmartisk:
For deploying nightly builds, I think we should consult @radcortez as he's in charge of the release automation scripts.
Generally, I'm not opposed to it, but I don't have a very good grasp of how exactly it would work.

rubik-cube-man:
I was thinking of pushing to GitHub packages similarly to how Keycloak does it: https://github.com/keycloak/keycloak/releases

I'm happy to hold off to ensure that it's done correctly. Let me know if you'd rather me write something that pushes elsewhere.

jmartisk:
That would mean uploading separate files, which would be hard to consume from Maven-based apps, no?
Maybe you could just build a pipeline somewhere on your side that builds the current main, and passes it as a zipped Maven repo to where you need it?

rubik-cube-man:
Yeah, no worries! I'll make something work on my end for now!

PR #2082, opened 2024-04-10
Track metrics of active GraphQL subscriptions

cristianonicolai:


server/implementation/src/main/java/io/smallrye/graphql/spi/MetricsService.java
```diff
@@ -6,4 +6,8 @@ public interface MetricsService {
     void start(Long measurementId, Context context);
 
     void end(Long measurementId);
+
+    void up(Context context);
```
jmartisk, line 10: Could we make it more clear what this is for? If it's for subscriptions only, I'd name these methods `subscriptionStart` and `subscriptionEnd`

jmartisk:
@mskacelik would you like to review this?

server/implementation/src/main/java/io/smallrye/graphql/spi/MetricsService.java
cristianonicolai: @jmartisk changes applied

jmartisk:
Thanks!
However, the CI is failing, looks like you need to wrap some code in the javadoc comments

cristianonicolai:
@jmartisk should be fixed now

mskacelik (approved):
+1

jmartisk:
Odd, the javadoc problem is still there and only occurs when compiling with JDK 11 (which we unfortunately still have to support), maybe we need to avoid using lambdas inside code in javadoc comments

cristianonicolai:
@jmartisk perhaps I should remove the sample code?

jmartisk:
> @jmartisk perhaps I should remove the sample code?

Yeah I guess so. It's not something intended to be used by application developers anyway.

cristianonicolai:
@jmartisk done

jmartisk (dismissed):

jmartisk:
There are some failures in Quarkus tests in CI. I think it might be related to how you add `metricsEmitter.start` and `metricsEmitter.end` in `AbstractDataFetcher`, causing duplicate (doubled) metrics measurements, because the `QuarkusDataFetcher` does the same too. I think it's not necessary to add here, no? It's not related to subscriptions, so I don't see the reason for it.

(to reproduce, you need to build a snapshot of smallrye-graphql, update Quarkus main branch to use the snapshot, and execute tests in https://github.com/quarkusio/quarkus/tree/main/extensions/smallrye-graphql/deployment)

cristianonicolai:
@jmartisk removed the `AbstractDataFetcher` changes; although I feel it would be better to consolidate this logic over here, I'm not completely aware if that would work how Quarkus needs.

jmartisk (approved):

jmartisk:
Thanks!

PR #2088, opened 2024-04-17
Added hint if schema build is failing due conflicting field name

suchwerk:
This just adds some hint how to resolve a naming conflict due  schema build is failing. 

The original error message (io.smallrye.graphql.schema.SchemaBuilderException):
```
Type 'MyEntity' already contains field named 'items' so source field, with the same name, cannot be applied
```
would become:
```
Type 'MyEntity' already contains field named 'items' so source field, with the same name, cannot be applied. You can resolve this conflict using @Ignore on the type's field.
```


jmartisk (approved):

mskacelik:
Thank you for making the changes. Could you please squash all the commits into one after your final finishing touches?

jmartisk:
No problem, I squashed them. Thanks @suchwerk

PR #2093, opened 2024-04-21
1479 client union

t1:
Support `@Union` in typesafe client.


client/implementation/pom.xml
```diff
@@ -67,8 +67,8 @@
         </dependency>
 
         <dependency>
-            <groupId>jakarta.json</groupId>
```
t1, line 70: This dependency was repeated. There are a few other dependencies where I'm not sure about their scope (some should be provided)

jmartisk:
This adds jandex into the runtime dependencies, which is a big no-no. 
Could this be instead integrated with the newer approach that avoids runtime reflection (the `client/model` module which builds a complete representation of the client API at build time from Jandex)? @mskacelik can provide some explanations around it I assume. The original approach where the client API is scanned using reflection at runtime is to be deprecated.

jmartisk:
> I've seen that, but it currently just looks like a PoC. Is there any documentation? I'd be very interested in understanding where we're going here!

It's not quite PoC, it's already the default behavior with Quarkus 3.9, but @mskacelik is working on "documenting" it as his university thesis. He's going to add some notes here in a bit. Or you can connect on zulip too

t1:
> Or you can connect on zulip too

I'd rather not use too many communication channels... let's keep it here 😁

mskacelik:
Hi @t1,
Two new modules have been added to the Type-Safe client (Quarkus): `smallrye-graphql-client-model` and `smallrye-graphql-client-model-builder`.  The `schema` and `schema-builder` on the server side heavily inspired this approach.

The `model` module works both in the build-time and runtime. In this module there is the [ClientModel](https://github.com/smallrye/smallrye-graphql/blob/main/client/model/src/main/java/io/smallrye/graphql/client/model/ClientModel.java) which represents the specific scanned API @GraphQLClientAPI. It contains a map of all the operations, with the key being the method signature and the value being the query generated by Jandex scanning. There is also the class ClientModels, which are all the scanned APIs. (Map of configKeys and ClientModel instances)
The bytecode recorder then stores this ClientModels class, and its instances are then used in the Proxy class during the GraphQL request building.

 For the model-builder, this module works only in build-time and builds the undermentioned ClientModel(s).  [ClientModelBuilder](https://github.com/smallrye/smallrye-graphql/blob/main/client/model-builder/src/main/java/io/smallrye/graphql/client/model/ClientModelBuilder.java) scans all the client APIs and generates their operation queries.  The scanning technology used here is Jandex, and it should only be used in this module during build-time.  The code in the model module is very similar to the reflect approach in the runtime, as the [QueryBuilder](https://github.com/smallrye/smallrye-graphql/blob/main/client/model-builder/src/main/java/io/smallrye/graphql/client/model/QueryBuilder.java) class builds the queries. It also contains  model classes representing fields, parameters, operations, and types (as in the operation's return type). But unlike the reflect approach, it uses Jandex scanning.
 
   With that in mind, this code is only responsible for doing the “serialization” and creating the GraphQL request for the proxy. “Deserialization” and handling of the GraphQL response stays the same (using the java.lang.reflect APIs).

You can try debug things in Quarkus version 3.9.4.

I hope that makes it more clear :D
If you have any other questions let me know.

jmartisk:
I have to add that it would be good to keep the current implementation (extending the original reflection-based approach) too for now, but we have to get rid of the Jandex dependency. I'm not yet completely sure how best to do that, but it sounds like it will have to be implemented in one of the build-time modules (`client/model`) and recorded via the `SmallRyeGraphQLClientRecorder` (or set as some sort of static field for non-Quarkus cases)

client/model/src/main/java/io/smallrye/graphql/client/model/TypeModels.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This is just a first step into having a more complete model of the GraphQL types in the client.
+ * It currently only holds the subtypes of interfaces that are used to model unions.
+ * In the long run, this should contain a complete model of all Java types relevant for the typesafe
+ * GraphQL client.
+ */
+public class TypeModels {
```
jmartisk, line 12: I don't fully understand the purpose of this class, can't this data be clumped together inside https://github.com/smallrye/smallrye-graphql/blob/ceb755444bbc4993d3436e2897103d28ecc2a513/client/model/src/main/java/io/smallrye/graphql/client/model/ClientModel.java which is registered a CDI bean (wrapped in https://github.com/smallrye/smallrye-graphql/blob/ceb755444bbc4993d3436e2897103d28ecc2a513/client/model/src/main/java/io/smallrye/graphql/client/model/ClientModels.java)? 
That would also avoid having to use static fields, which is a major problem for wildfly

jmartisk: Or is keeping reference to implementations necessary at all? I guess it could just be reflected in the query string that is pre-generated for each graphql operation

t1: I like that idea: we could produce all query strings at startup also for the reflection-based approach. Nice! I'm not sure how much effort that would be. I had thought that adding support for Unions would be a quick thing, but it doesn't look like it will be 😁

About the `static`: I had forgotten. Do you have a suggestion how to fix it? Should we use the `CDI` class to find an application-scoped instance of the `TypeModels`, of if we'd put it into `ClientModels`, make that application-scoped?!?

t1:
@mskacelik: would it be possible to have only one query string generator, but two meta data providers, one for Quarkus and one via reflection? The former would run at build time, while the latter at startup (or maybe on-demand). And when a request has to be built, both would use the same static query storage with the method as a key.

jmartisk:
> I have another problem with this approach: If I have a Java SE client, e.g. in a test, the Jandex has not been scanned. I would love to use sealed classes for this instead, but they are JDK 17, not 11 😿

Yeah I fear we will have to, for now, just keep supporting both approaches (ClientModels and reflection) until we find a reasonable way way to create a ClientModel in pure Java SE apps, but without making Jandex a required dependency - so, probably a separate artifact that serves as the tool to create it...

client/model/src/main/java/io/smallrye/graphql/client/model/TypeModels.java
t1: Wouldn't it be enough to annotate the class as `@ApplicationScoped` and add a `beans.xml`? In the `TypeInfo` we could then use the `CDI` class.

jmartisk: That would mean the CDI container would create an instance, but we currently build an instance manually at build-time, and then run it through a bytecode recorder and register the instance as a synthetic bean. Potentially we could rework it to a regular bean created by CDI and somehow set the value during startup, or something... but why wouldn't you be able to use the `CDI` class this way? It's still a bean in the CDI container, that will work normally.

t1:
What do you think of this idea:

1. We create an interface for the `VertxTypesafeGraphQLClientProxy#queryCache` named `OperationStore`. It provides the query string for a `MethodKey` (which should BTW really be immutable and become a record in the future). `MethodInvocation#getKey` would become obsolete. We pass that interface as a constructor parameter instead of the `ClientModel`.
2. The `VertxTypesafeGraphQLClientBuilder` either passes an adapter for the `ClientModel` that fetches the query string from the map, or a reflection-based adapter that calculates and caches the operation string dynamically. I will see how we can do this in a good way.
3. Now comes the interesting part: we move all the classes that are used to build operation strings via reflection into a new module `model-reflection` (or so). This would be excluded in Quarkus. There is no standard for the typesafe client, so we we are free to go ahead and compile this module with Java 17, so we can use reflection on sealed classes to provide the subclasses of a Union instead of Jandex.

Can you already see any major flaws or should I give it a try, so we can talk about real code?

client/model/src/main/java/io/smallrye/graphql/client/model/TypeModels.java
t1: The suggestion I just postet in the PR would make this unnecessary.

jmartisk:
Hm WildFly still compiles with Java 11 and supports applications running on it, so by switching something to 17 we would break it :( 

Also regarding the sealed interfaces: Do I understand it that you would require the union interface to be sealed?  What if it's not? Or, what if it has any implementations that the user didn't intend to be part of the union? Maybe things would be much easier if we required the user to explicitly specify the list of classes that should be part of the union. Then you wouldn't need jandex to find the implementations either.

By the way, we shouldn't exclude the reflection-based approach in Quarkus (at least not yet). Firstly, we've just added a config property that reverts to the reflection-based behavior, so until we've had enough bake time and we're sure that the ClientModel approach works well, we should keep it.. Secondly, the reflection approach is still needed in case that somebody creates clients dynamically using the builder at runtime - we don't have the pre-built ClientModel available in that case.

t1:
Current WildFly versions still support Java 11, true. But they've also been supporting Java 17 for a while; I can't recall since when exactly... do you happen to know? I personally could live with a new version breaking on Java 11. An alternative would be to extract that part to another, optional module, and load it, e.g., with a `ServiceLoader`.

I think a very elegant way to specify the allowed subclasses is sealed interfaces. We could use a different mechanism, but that would be cumbersome, add additional APIs or configs outside of the code. Sealed interfaces would also make switching on the different subtypes super nice. So I think this would be easier and more aligned to where Java is going.

> we shouldn't exclude the reflection-based approach in Quarkus

Okay. And it also still supports Java 11. So we have the same issue as we have with WildFly. So I guess we really need that optional Java 17 module.

jmartisk:
The whole SmallRye org uses one common strategy and build system that runs with JDK 11 (and it will stay like this until WildFly moves to 17+), I don't see it worthwhile to break it for one feature that needs JDK 17 :/ Also the need to use sealed interfaces seems to put quite a burden on application developers...
 I'd go with the Jandex approach that scans it within the `model-builder` module, or have the user explicitly list the classes that should be part of the union...

t1:
Okay, so we'll stick to 11 for now. I still think that sealed classes would not put a burden on developers, but if we support both options (because it will be hard to remove an option later), we'll see what people actually use.

I've just pushed step one and two of my suggestion: I've introduced the `OperationRepository`. Would you like to take a look before I continue with extracting the reflection-based operation building?

t1:
> I'd go with the Jandex approach that scans it within the model-builder module

IIUC, the model-builder is only used in Quarkus, and it has a dependency on Jandex, so we can't use it, can we?

Or, to put it in other words: even if we introduce a new module, as long as it is not optional, it can't depend on Jandex. So we'll _have to_ introduce some sort of API, maybe an annotation like `@UnionTypes({Foo.class, Bar.class})`?

jmartisk:
> IIUC, the model-builder is only used in Quarkus, and it has a dependency on Jandex, so we can't use it, can we?

Currently, it is only in Quarkus, but we'd like to integrate it in WildFly too, same as the server-side schema-builder is used there too. I'm just not yet sure when that will be (unless you want to do it). It should, at deployment time, build a ClientModels instance and register it (via a CDI extension most likely) to be available to the app, just like Quarkus does it. 

I do think that `@UnionTypes({Foo.class, Bar.class})` with an explicit list of implementations would be a good way, avoiding the complexity of scanning the app (including dependencies, to make it worse) and the risk that there will be some implementations that are not meant to be members of the union.

client/model/src/main/java/io/smallrye/graphql/client/model/MethodKey.java
```diff
@@ -4,9 +4,9 @@
 import java.util.Objects;
 
 public class MethodKey {
-    private Class<?> declaringClass;
-    private String methodName;
-    private Class<?>[] parameterTypes;
+    private final Class<?> declaringClass;
```
mskacelik, line 7: @jmartisk Do you think that this will work with the bytecode recorder? (removing set methods)

jmartisk: It definitely must not be final and must have getters and setters to work with Quarkus bytecode recorder

mskacelik: Oh yeah, I was planning to add some Javadoc in the model classes (ClientModels, ClientModel, MethodKey), so that's partially my fault for not explaining the reason for having the set methods.

t1: I just reverted that; and added javadoc and Deprecated annotations

t1:
The WildFly Feature Pack doesn't find the `MethodKey` class. How can we fix that?

t1:
BTW: probably WF 26 will (probably) drop support for Java 11, because it will (probably) support Jakarta 11, which requires Java 17. Maybe we should wait with the Unions for that to happen?

t1:
I'd like to give this another try... this time with just simple reflection using the JSON-B `@JsonbTypeInfo/@JsonbSubtype` annotations. Works pretty well without much code. And it's straight forward to communicate: you want polymorphic behavior, just add the corresponding JSON-B annotations.

I converted into a draft, because I'd like to get early feedback, while there are still some things left to do:

- [ ] add the same behavior for GraphQL interfaces
- [ ] use the `@JsonbSubtype#alias` (maybe we don't have to)
- [ ] use the `@JsonbTypeInfo#key` and fail on the default `@type` (maybe we don't have to)
- [ ] something else?

client/implementation-vertx/src/test/java/test/tck/VertxTypesafeGraphQLClientFixture.java
```diff
@@ -64,7 +64,7 @@ public <T> T build(Class<T> apiClass) {
 
     @Override
     public TypesafeGraphQLClientBuilder builder() {
-        boolean clientModelCase = System.getProperty("clientModelCase").equals("true");
+        boolean clientModelCase = "true".equals(System.getProperty("clientModelCase"));
```
t1, line 67: make null-safe

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/FieldInfo.java
```diff
@@ -111,16 +113,9 @@ public void set(Object instance, Object value) {
...
-                    Thread.currentThread().getContextClassLoader());
-        } catch (ClassNotFoundException e) {
-            /* IN CASE THE CLASS IS NOT IMPORTED */
-        }
-        return field.isAnnotationPresent(NonNull.class)
-                || (jakartaNotNullClass != null && field.isAnnotationPresent(jakartaNotNullClass))
-                || getType().isPrimitive();
+        return getType().isPrimitive()
+                || field.isAnnotationPresent(NonNull.class)
+                || isJakartaNotNull(field);
```
t1, line 118: Reuse the same code from TypeInfo

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
```diff
@@ -30,17 +32,37 @@
...
+        try {
+            //noinspection unchecked
+            return (Class<? extends Annotation>) Class.forName(className, false,
+                    Thread.currentThread().getContextClassLoader());
+        } catch (ClassNotFoundException e) {
+            /* IN CASE THE CLASS IS NOT IMPORTED */
+            return null;
+        }
+    }
+
```
t1, line 65: Actually off-topic, but we can do this stuff only once and not clutter the calling code.

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
```diff
@@ -98,6 +120,18 @@ public String getTypeName() {
...
+        if (isAnnotated(org.eclipse.microprofile.graphql.Type.class))
+            return getAnnotation(org.eclipse.microprofile.graphql.Type.class).value();
+        return getSimpleName();
+    }
+
+    public String getSimpleName() {
+        if (type instanceof Class)
+            return ((Class<?>) type).getSimpleName();
+        return getTypeName();
+    }
```
t1, line 133: I just moved this method further up

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
```diff
@@ -167,24 +201,15 @@ private Field[] getDeclaredFields(Class<?> type) {
...
     private boolean isGraphQlField(Field field) {
-        Class jsonIgnoreClass = null;
-        try {
-            jsonIgnoreClass = Class.forName("com.fasterxml.jackson.annotation.JsonIgnore", false,
-                    Thread.currentThread().getContextClassLoader());
-        } catch (ClassNotFoundException e) {
-            /* IN CASE THE CLASS IS NOT IMPORTED */ }
         return !isStatic(field.getModifiers()) && !isSynthetic(field.getModifiers()) && !isTransient(field.getModifiers())
-                && !isAnnotatedBy(field, Ignore.class, JsonbTransient.class, jsonIgnoreClass);
+                && !isAnnotatedBy(field, Ignore.class, JsonbTransient.class, JACKSON_JSON_IGNORE);
```
t1, line 205: Actually off-topic. See above.

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
```diff
@@ -167,24 +201,15 @@ private Field[] getDeclaredFields(Class<?> type) {
...
-        for (Class<? extends Annotation> annotationClass : annotationClasses) {
-            if (annotationClass != null
-                    && field.isAnnotationPresent(annotationClass)) {
-                return true;
-            }
-        }
-        return false;
+        return Stream.of(annotationClasses)
+                .filter(Objects::nonNull)
+                .anyMatch(field::isAnnotationPresent);
```
t1, line 212: Also off-topic, but nicer code

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
```diff
@@ -313,33 +342,18 @@ private Constructor<?> getDeclaredConstructor(Class<?> type, Class<?>[] paramete
...
         if (annotatedType != null)
-            return annotatedType.isAnnotationPresent(NonNull.class) ||
-                    (jakartaNotNullClass != null && annotatedType.isAnnotationPresent(jakartaNotNullClass));
+            return annotatedType.isAnnotationPresent(NonNull.class) || isJakartaNotNull(annotatedType);
         if (container == null || !container.isCollection() || container.annotatedType == null)
             return false; // TODO test
-        return container.annotatedType.isAnnotationPresent(NonNull.class) ||
-                (jakartaNotNullClass != null && container.annotatedType.isAnnotationPresent(jakartaNotNullClass));
+        return container.annotatedType.isAnnotationPresent(NonNull.class) || isJakartaNotNull(container.annotatedType);
+    }
```
t1, line 353: this is also that off-topic topic

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
```diff
@@ -313,33 +342,18 @@ private Constructor<?> getDeclaredConstructor(Class<?> type, Class<?>[] paramete
...
 
+    static boolean isJakartaNotNull(AnnotatedElement c) {
+        return JAKARTA_NOT_NULL != null && c.isAnnotationPresent(JAKARTA_NOT_NULL);
```
t1, line 356: this is also that off-topic topic

client/model-builder/src/main/java/io/smallrye/graphql/client/model/ClientModelBuilder.java
```diff
@@ -60,7 +60,7 @@ private ClientModels generateClientModels() {
             ClientModel operationMap = new ClientModel();
             ClassInfo apiClass = graphQLApiAnnotation.target().asClass();
             List<MethodInfo> methods = getAllMethodsIncludingFromSuperClasses(apiClass);
-            methods.stream().forEach(method -> {
+            methods.forEach(method -> {
```
t1, line 63: minor off-topic... couldn't resist to fix this warning.

t1:
Adding support for GraphQL interfaces was even easier than I had thought!

mskacelik:
Also, I have added https://github.com/smallrye/smallrye-graphql/pull/2160 fix for the `TypesafeTckClientModelSuite` issue where it executed the `*Behavior` tests with the Java Reflection implementation (via `mvn clean install`). That was the reason thy `InterfaceBehaviour` and `UnionBehaviour` tests were working with `TypesafeTckClientModelSuite` despite not having implementation yet.

jmartisk:
I'm trying to run the tests properly, but I'm getting a completely weird error when building Quarkus. I'll get back to this once I figure that out

jmartisk:
@t1 , I think I've fixed the CI pipeline, could you rebase please?

t1:
If you don't have any further suggestions, great!

I re-thought about checkmarks I left "open": it's probably best to not implement them. It would just make things more complicated.

But I guess, we should also add some documentation. I'll try to find time for that and create a new PR.

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
```diff
@@ -447,4 +466,21 @@ public boolean isAnnotated(Class<? extends Annotation> type) {
...
+
+    public TypeInfo subtype(String typename) {
+        return subtypes()
+                .filter(c -> c.getGraphQlTypeName().equals(typename))
+                .findAny()
+                .orElseThrow(() -> SmallRyeGraphQLClientMessages.msg.cannotInstantiateDomainObject(typename, null));
+    }
+
+    public Stream<TypeInfo> subtypes() {
+        var jsonbTypeInfo = getAnnotation(JsonbTypeInfo.class);
```
jmartisk, line 478: Could we support the `@JsonTypeInfo`  from Jackson too? In Quarkus, there's a general movement to move away from JSON-B/P in favor of Jackson. For most annotations, we support the JSON-B as well as Jackson ones.

jmartisk:
If you don't mind, I'd prefer having at least some basic docs in the initial PR :)

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
t1: We talked about that before. I understand that Yasson is notorious for its slowness. But wouldn't it be wrong to blame that on JSON-B? That's just the API, and the implementation doesn't _have to be_ slow, does it?

Is there an open discussion about this somewhere?

jmartisk:
And how about the variant when a pre-built `ClientModel` is used (which is the default behavior in Quarkus now)? It would be nice to have straight away, but we can add it later if necessary.

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/TypeInfo.java
jmartisk: See https://github.com/quarkusio/quarkus/issues/42748 
But yeah this is probably not that important right now, the move to a general SPI is still to be thought through and AFAIK we don't have a compile-time Jackson dependency, so outside the `client-model-builder` module, without Jandex, we would have to do it reflectively.

t1:
> And how about the variant when a pre-built `ClientModel` is used (which is the default behavior in Quarkus now and @mskacelik is working on doing that for WF too)? It would be nice to have straight away, but we can add it later if necessary.

I understand that @mskacelik has been working on this. See https://github.com/smallrye/smallrye-graphql/pull/2093#issuecomment-2304834999

t1:
> If you don't mind, I'd prefer having at least some basic docs in the initial PR :)

done :)

jmartisk:
> I understand that @mskacelik has been working on this. See [#2093 (comment)](https://github.com/smallrye/smallrye-graphql/pull/2093#issuecomment-2304834999)

Not really, the point was that he fixed an issue that made the TCK run with the reflection approach when it was supposed to run with ClientModel

t1:
> It would be nice to have straight away, but we can add it later if necessary.

Yeah, let’s add this later, when we change all the reflection code

mskacelik:
I can implement it; I was just making sure that the reflection-side code was in the finished "state".

jmartisk (approved):
Ok let's go then! Thanks!

Issue #2097, opened 2024-04-26
Wrongly inferred query name when typesafe client API extends an interface


jmartisk:
When a typesafe client API extends from an interface like this:

```
public interface InventoryClient {
    List<Car> allCars();
}

@GraphQLClientApi(configKey = "inventory")
public interface GraphQLInventoryClient extends InventoryClient {
    @Query("cars")
    List<Car> allCars();
}
```

Then the client thinks the query is named `allCars` instead of `cars`. The `@Query` annotation is ignored.

PR #2098, opened 2024-04-26
Fix `Method Key` class, which had unnecessary attribute.

mskacelik:
fix:#2097
/cc @jmartisk 
I also added some __javadocs__.

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/MethodInvocation.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/MethodInvocation.java
@@ -54,7 +54,7 @@ public class MethodInvocation implements NamedElement {
     }
 
     public MethodKey getMethodKey() {
-        return new MethodKey(method.getReturnType(), method.getName(), method.getParameterTypes());
+        return new MethodKey(method.getName(), method.getParameterTypes());
     }
 
     public OperationType getOperationType() {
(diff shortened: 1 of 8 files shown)


client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
```diff
@@ -305,6 +305,7 @@ private JsonObject request(MethodInvocation method) {
         if (clientModel == null) {
             query = queryCache.computeIfAbsent(method.getKey(), key -> new QueryBuilder(method).build());
         } else {
+            clientModel.getOperationMap().keySet().forEach(System.err::println);
```
jmartisk, line 308: Please remove this line, or use a logger  on debug level

jmartisk: Ah, this is logged for each request, so rather remove. This is something that only needs to be logged once, if at all

mskacelik: Oh yeah, my bad. I left it while I was debugging.

jmartisk (approved):

jmartisk:
Nice, thanks!

PR #2103, opened 2024-05-05
Fixed dependencies, removed Sys. println

mskacelik:
/cc @jmartisk

### DIFF
--- a/client/api/pom.xml
+++ b/client/api/pom.xml
@@ -30,16 +30,10 @@
             <artifactId>microprofile-config-api</artifactId>
             <scope>provided</scope>
         </dependency>
-
         <dependency>
             <groupId>io.smallrye.reactive</groupId>
             <artifactId>mutiny</artifactId>
             <scope>provided</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>io.smallrye</groupId>
-            <artifactId>smallrye-graphql-client-model</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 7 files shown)


client/model-builder/src/main/java/io/smallrye/graphql/client/model/SmallRyeGraphQLClientMessages.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.jboss.logging.Messages;
+import org.jboss.logging.annotations.Cause;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageBundle;
+
+import io.smallrye.graphql.client.InvalidResponseException;
+
+@MessageBundle(projectCode = "SRGQLDC")
+public interface SmallRyeGraphQLClientMessages {
```
jmartisk, line 14: I don't understand the purpose of this. These are errors that happen at runtime, but you add this to a module that is only present at build time (and you're also leaving the original copy in the `client/implementation` module which means we would have this duplicated

mskacelik: The idea was that I did not want to have a dependency on `implementation` in the `model-builder.`
We use one of the messages for recursion exception during query building.

jmartisk: Then we should have this class only contain the message relevant to recursion, and change the numbers to not overlap with the original logging class. Let's switch from 35000-39999 to 40000-44999, that range isn't used right now.And the message about recursion will have an ID of 40000

mskacelik:
new quarkus branch: https://github.com/mskacelik/quarkus/tree/srgql-2.8.4

jmartisk (approved):

Issue #2111, opened 2024-05-17
Error generating query $null parameter


vitomanu96:
I'm using Quarkus 3.10.1 with SmallRye-GraphQL-Client library.

I have to fetch data from this query:

`
query($listaIdTitoli: [BigInteger])  {
  loadProvvigioniSuTitoli(listaIdTitoli: $listaIdTitoli){
    id
  }
}
`

But when Quarkus start print this generated query:

`Query created: query loadProvvigioniSuTitoli($null: [BigInteger]) { loadProvvigioniSuTitoli(listaIdTitoli: $null) {id} }`

**Quarkus set $null insted of $listaIdTitoli**

Here my interface definition:

```
@GraphQLClientApi(configKey = "p2l")
public interface P2LGraphQLClient {

    /**
     * Recupero delle provvigioni di secondo livello a partire da un idTitolo
     * @param listaIdTitoli lista degli id dei titoli tecnici
     * @return la lista delle provvigioni per titolo
     */
    List<ProvvigioneTitolo> loadProvvigioniSuTitoli(@Name("listaIdTitoli") List<Long> listaIdTitoli);
}
```

mskacelik:
Very strange, gonna look into it

ErtiRakipaj:
> Very strange, gonna look into it

A solution that worked was removing the `@Name` annotation from the parameter on the interface definition and adding: 
`<arg>-parameters</arg>` under `<compileArgs>` on the POM, if you're using `maven-compiler-plugin`.  Like this for example: 
```
               <configuration> 
                    <compilerArgs>
                        --other args that may be here
                        <arg>-parameters</arg>  -- this is the arg to add
                    </compilerArgs>
                </configuration>
```
Basically passing the parameters to the java compiler via maven, allowing access at runtime.  When you remove the @ Name annotation this is a solution that is also on the logs, which says:
`You can either annotate all parameteres with @Name, or compile your source code with the -parameters options, so the parameter names are compiled into the class file and available at runtime.`

mskacelik:
Oh yes, you are right, thank you @ErtiRakipaj :). With that said, this _exact_ reproducer (with removed configuration) does not log out the given error message. So, I will create a PR with the fix in a few.

mskacelik:
I understand now; the Java Reflection approach previously returned the parameter argument names as arg0, arg1, arg2,... 
[...]

PR #2112, opened 2024-05-18
Added default name for non-runtime included parameters (for JANDEX API), client-side

mskacelik:
/cc @jmartisk 
fixes: #2111
Added a default name for parameters not configured to be part of the runtime. The default names are equivalent to the Java Reflection API names (arg0, arg1, arg2,...), so they can be consistent in the GraphQL Request building phase.

I am unsure if tests are necessary since I don't think we have tests for such cases, but I can make them if you wish.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,8 +96,8 @@ jobs:
       - name: checkout Quarkus repository
         uses: actions/checkout@v2
         with:
-          repository: mskacelik/quarkus
-          ref: srgql-2.8.2
+          repository: quarkusio/quarkus
+          ref: main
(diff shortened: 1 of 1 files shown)


mskacelik:
Not really sure why Quarkus's tests failed

jmartisk:
@mskacelik isn't that because the CI tests are still configured to run against your 2.8.2 branch? https://github.com/smallrye/smallrye-graphql/blob/c757c099162d8d4674e7884418693375c990cc33/.github/workflows/build.yml#L99 could you try switching it to the official Quarkus main?

jmartisk (approved):

PR #2116, opened 2024-05-24
Issue/2068

robp94:


mskacelik:
I tried to play around with the new implementation in the given reproducer, and it seems that everything works as it should. +1

mskacelik (approved):

jmartisk:
Thanks @robp94  and  @mskacelik !!

mskacelik:
Out of curiosity, does the federation require the interface and its implementors to have the same key identifier?

robp94:
> Out of curiosity, does the federation require the interface and its implementors to have the same key identifier?

We use it that way, but I never tried it with different keys.

robp94:
@jmartisk could we get this in quarkus 3.12.0?

jmartisk:
@robp94 Sorry, there are some issues with the release pipeline that need fixing first, and 3.12 is already in CR so it's probably too late (the changes in smallrye-graphql 2.9 will be quite big) so I'm rather targeting 3.13 now

Issue #2096, opened 2024-04-25
how to add Alias to returned field name?


AzaliyaUfa:
with io.smallrye.graphql.client.core.utils.validation.NameValidation.validateFieldName(NameValidation.java:85) it became impossible to return "required_name: name" construction in Field.field

mskacelik:
Hmm, it seems that the cause of the exception is the white space after the colon. So deleting it would fix the issue. 

That being said GraphQL specification has listed white space as part of the Ignored Tokens https://spec.graphql.org/draft/#sec-Language.Source-Text.Ignored-Tokens.

So, If you want, we can add this feature so the name validation will ignore all the ignored tokens around the Lexical Tokens.

AzaliyaUfa:
thank you so much, it really fixed the problem! 
but adding the option to set an Alias to the field would being great))

jmartisk:
fixed by https://github.com/smallrye/smallrye-graphql/pull/2118

PR #2118, opened 2024-05-29
added some ignored tokens for name validations (dynamic)

mskacelik:
/cc @jmartisk 
fix:#2096

see: https://spec.graphql.org/draft/#sec-Language.Source-Text.Ignored-Tokens

I only added commas and whitespaces.

Afaik, these ignored tokens are ignored by the graphql runtime.


jmartisk (approved):
Nice!

PR #2122, opened 2024-05-31
Schema Builder doesn't support inheriting interface field with subtype

zche-theScore:
Suppose we have the GraphQL Schema below
```
interface FieldInterface {...}
type FieldType implements FieldInterface {...}
interface ContainerInterface {
  field: FieldInterface
}
```
According to GraphQL [spec](https://spec.graphql.org/October2021/#sec-Objects.Type-Validation), we can have a type
```
type ContainerType {
  field: FieldType
}
```
However, as my test cases show, if I define the Java Classes/Interfaces the same way as the GraphQL type relationship, the schema builder would generate `field: FieldInterface` instead of `field: FieldType` in `ContainerType`.

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/TypeCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/TypeCreator.java
@@ -40,32 +40,34 @@ public class TypeCreator extends AbstractCreator {
     @Override
     protected void addFields(Type type, ClassInfo classInfo, Reference reference) {
         // Fields
-        Map<String, MethodInfo> allMethods = new HashMap<>();
+        List<List<MethodInfo>> allMethods = new LinkedList<>();
         Map<String, FieldInfo> allFields = new HashMap<>();
 
         // Find all methods and properties up the tree
         for (ClassInfo c = classInfo; c != null; c = ScanningContext.getIndex().getClassByName(c.superName())) {
+            List<MethodInfo> listMethods = new ArrayList<>();
+
             if (InterfaceCreator.canAddInterfaceIntoScheme(c.toString())) { // Not java objects
                 List<MethodInfo> classMethods = filterOutBridgeMethod(c.methods());
-                for (MethodInfo classMethod : classMethods) {
-                    allMethods.putIfAbsent(classMethod.name(), classMethod);
-                }
-                for (MethodInfo interfaceMethod : getAllInterfaceMethods(c, classMethods
+                listMethods.addAll(getAllInterfaceMethods(c, classMethods
                         .stream()
-                        .map(MethodInfo::name)
-                        .collect(Collectors.toSet()))) {
-                    allMethods.putIfAbsent(interfaceMethod.name(), interfaceMethod);
-                }
+                        .map(MethodInfo::toString)
+                        .collect(Collectors.toSet())));
+                listMethods.addAll(classMethods);
                 for (FieldInfo fieldInfo : c.fields()) {
                     allFields.putIfAbsent(fieldInfo.name(), fieldInfo);
                 }
             }
+
+            allMethods.add(0, listMethods);
         }
-        for (MethodInfo methodInfo : allMethods.values()) {
-            if (MethodHelper.isPropertyMethod(Direction.OUT, methodInfo)) {
-                String fieldName = MethodHelper.getPropertyName(Direction.OUT, methodInfo.name());
-                FieldInfo fieldInfo = allFields.remove(fieldName);
-                fieldCreator.createFieldForPojo(Direction.OUT, fieldInfo, methodInfo, reference).ifPresent(type::addField);
+        for (List<MethodInfo> listMethods : allMethods) {
+            for (MethodInfo methodInfo : listMethods) {
+                if (MethodHelper.isPropertyMethod(Direction.OUT, methodInfo)) {
+                    String fieldName = MethodHelper.getPropertyName(Direction.OUT, methodInfo.name());
+                    FieldInfo fieldInfo = allFields.remove(fieldName);
+                    fieldCreator.createFieldForPojo(Direction.OUT, fieldInfo, methodInfo, reference).ifPresent(type::addField);
+                }
             }
         }
 
@@ -100,7 +102,7 @@ public class TypeCreator extends AbstractCreator {
                                 parentInterfaceInfo
                                         .methods())
                                 .stream()
-                                .filter(method -> isNotGenericType(method) && methodMemory.add(method.name())),
+                                .filter(method -> isNotGenericType(method) && methodMemory.add(method.toString())),
                         getAllInterfaceMethods(parentInterfaceInfo, methodMemory).stream()))
                 .collect(Collectors.toList());
     }
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/TypeCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/TypeCreator.java
@@ -1,11 +1,6 @@
 package io.smallrye.graphql.schema.creator.type;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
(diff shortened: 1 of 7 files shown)


common/schema-builder/src/test/java/io/smallrye/graphql/index/SchemaBuilderTest.java
```diff
@@ -171,6 +171,28 @@ public void testSchemaWithBatchSourceFieldNameDuplicates() {
...
+        // Get Types
+        Map<String, Type> types = schema.getTypes();
+        Map<String, Type> interfaces = schema.getInterfaces();
+        Type containerType = types.get("ContainerType");
+        Field containerTypeField = containerType.getFields().get("inheritField");
+        Type containerInterface = interfaces.get("ContainerInterface");
+        Field containerInterfaceField = containerInterface.getFields().get("inheritField");
+
+        // TODO: Should be FieldInterface according to GraphQL Spec: https://spec.graphql.org/October2021/#sec-Objects.Type-Validation
+        assertEquals("io.smallrye.graphql.index.inherit.FieldInterface",
```
jmartisk, line 190: Just to make sure I understand correctly - you think this should be `FieldType` instead of `FieldInterface`?

zche-theScore:
I PRed a simple fix here
https://github.com/zche-theScore/smallrye-graphql-inherit-field-with-subtype/pull/1

jmartisk:
Could you add the contents of that PR here? The fix should be in the same PR as the tests :)
Thanks!

zche-theScore:
Implementation done. May I get some reviews? Thanks
cc. @mskacelik @phillip-kruger @ybroeker

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/TypeCreator.java
```diff
@@ -45,24 +40,28 @@ public TypeCreator(ReferenceCreator referenceCreator, FieldCreator fieldCreator,
...
         Map<String, FieldInfo> allFields = new HashMap<>();
 
         // Find all methods and properties up the tree
         for (ClassInfo c = classInfo; c != null; c = ScanningContext.getIndex().getClassByName(c.superName())) {
             if (InterfaceCreator.canAddInterfaceIntoScheme(c.toString())) { // Not java objects
                 List<MethodInfo> classMethods = filterOutBridgeMethod(c.methods());
-                allMethods.addAll(classMethods);
-                allMethods.addAll(getAllInterfaceMethods(c, classMethods
+                for (MethodInfo classMethod : classMethods) {
+                    allMethods.putIfAbsent(classMethod.name(), classMethod);
```
jmartisk, line 51: Hold on, why exactly do you need to turn the List into a Map and use `classMethod.name()` as a key? That will clash in case that the method has overrides, no?

zche-theScore: From my understanding, GraphQL doesn't support overloading (both param and return type), i.e. in each GraphQL Object, there cannot be two field with the same name even if they have different param types. Therefore, I think having `MethodInfo.toString()` can be replaced by `MethodInfo.name()`.

In the case of Java methods with the same name, the current implementation puts them both in the `allMethods` list. Then in the following loop
```
        for (MethodInfo methodInfo : allMethods.values()) {
            if (MethodHelper.isPropertyMethod(Direction.OUT, methodInfo)) {
                String fieldName = MethodHelper.getPropertyName(Direction.OUT, methodInfo.name());
                FieldInfo fieldInfo = allFields.remove(fieldName);
                fieldCreator.createFieldForPojo(Direction.OUT, fieldInfo, methodInfo, reference).ifPresent(type::addField);
            }
        }
```
Both of the two Java methods get a call to `type::addField` whose implementation is basically `LinkedHashMap::put` with key `fieldName`, therefore, the second method in the list will override the previous one implicitly.

My change from `List<MethodInfo>` to `Map<String, MethodInfo>` aims to bring this implicit overwrite explicit by maintaining one method per fieldName when we are looping over the Java Inheritance Tree.

My intentioned code behavioural change is
1. the same order for Java Inheritance Tree walk,
2. for methods with the same fieldName, instead of the last encounter overwrites, the earliest encounter takes priority.

jmartisk: GraphQL doesn't have overloading, but Java does, and here we're talking about the name of the Java method, not of the resulting GraphQL field. We can have multiple methods with the same name, differentiated using a `@Name` annotation to denote the name of the GraphQL field to be created from them

jmartisk: There's some failures in the tests, please try to run the test suite, it's probably related

zche-theScore: > GraphQL doesn't have overloading, but Java does, and here we're talking about the name of the Java method, not of the resulting GraphQL field. We can have multiple methods with the same name, differentiated using a `@Name` annotation to denote the name of the GraphQL field to be created from them

Indeed I didn't know about `@Name` directive. Now that makes sense to me. Then another possible fix is that instead of `List<MethodInfo> allMethods`, we put them in different categories depending how many layers between the method to the class, i.e. `List<List<MethodInfo>> allMethods`. And then we process the layers furtherest according to my logic above.

I will add a commit for it.

> There's some failures in the tests, please try to run the test suite, it's probably related

The failures are because I added the methods according to the order `for (MethodInfo methodInfo : allMethods.values())`, which is not determined order for HashMap. And the test relies on a specific order of adding fields.

Additional Note, very unfortunate we need to do one message per day, I live in the EST zone btw.

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/TypeCreator.java
```diff
@@ -45,28 +40,34 @@ public TypeCreator(ReferenceCreator referenceCreator, FieldCreator fieldCreator,
...
                         .stream()
                         .map(MethodInfo::toString)
                         .collect(Collectors.toSet())));
+                listMethods.addAll(classMethods);
                 for (FieldInfo fieldInfo : c.fields()) {
                     allFields.putIfAbsent(fieldInfo.name(), fieldInfo);
                 }
             }
+
+            allMethods.add(0, listMethods);
```
jmartisk, line 62: Always `0`? I assume you want some `i` that is incremented in each execution of the enclosing `for` loop

jmartisk: Or just simply call `add(listMethods)` without specifying an index

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/TypeCreator.java
jmartisk: Cool, it looks good to me now, I'll open more conversations if I find more issues.
I am in CET and unfortunately usually finish work quite early, but I can try to have a look in the evening too, hopefully

common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/type/TypeCreator.java
jmartisk: Ah, my bad, I thought this was always overwriting the element on that index, I had to re-read the javadoc.

jmartisk (approved):
Yeah I think it's good now. Thanks a lot!

zche-theScore:
Hi @jmartisk, I would like to know if there will be a planned release soon?

jmartisk:
@zche-theScore there are some issues currently with the release pipeline for SmallRye projects, but I'll see what can be done

zche-theScore:
Hi, I see the latest release doesn't contain this. Is there any issues with these changes?

jmartisk:
@zche-theScore I'm planning to release 2.9.0 to be included in Quarkus 3.13, which is coming out sometime in late July. I'll do the SmallRye release probably in the first half of July to allow some time for the upgrade (and vacations etc), if that's fine with you.

PR #2130, opened 2024-06-22
fix #2094: replace the dependency on yasson with jsonb and jsonp

### LINKED ISSUE #2094: Dependency on yasson
I just found the following code in `RequestImpl` that necessitates the dependency on yasson, which we should get rid of.

```java
try (Jsonb jsonb = JsonbBuilder.create()) {
    JsonStructure struct = ((JsonBinding) jsonb).toJsonStructure(v);
    varBuilder.add(k, struct);
} catch (Exception ignore) {
}
```

Why is this not simply `varBuilder.add(k, jsonb.toJson(v));` and why is the exception ignored?

If there is a valid reason, this deserves a comment. Anyone any idea? @hogmuzzle maybe?

t1:
I had completely forgotten about this 🤪


jmartisk (approved):
thanks!

Issue #1366, opened 2022-04-21
Problems with annotations like nonnull and lists


robp94:
I found more problems with annotations and lists.

### Java
Code like this
```
    @Query
    public String testNull(List<String> list)

    @Query
    public String test1(@NonNull List<String> list)

    @Query
    public String test2(List<@NonNull String> list)

    @Query
    public String test3(@NonNull List<@NonNull String> list)
```
results in a schema like this:
```
type Query {
  test1(list: [String]!): String
  test2(list: [String!]!): String
  test3(list: [String!]!): String
  testNull(list: [String]): String
}
```

which seems not correct, even so I don't know if anybody would need 
```
test2(list: [String!]): String
```
instead of the actual result.

The problem is that the typesafe client seems to handle it correctly.

With an api like:

```
    @Query
    public String test2(List<@NonNull String> list);
```
which calls test2 from above, the result is

```
Variable type '[String!]' doesn't match expected type '[String!]!' @ 'test2', locations=[{line=1, column=45}], path=null, extensions={classification=ValidationError}})
```

### Kotlin
There seems to be another discrepancy between client and server

```
    @Query
    fun test2(list: List<@NonNull String>): String

    @Query
    fun test3(list: @NonNull List<@NonNull String>): String
```
result in schema 
```
  test2(list: [String!]!): String
  test3(list: [String!]!): String
```

But both test fail with 
```
Variable type '[String!]' doesn't match expected type '[String!]!
```
I think that both server and client should fit together.

### Furthermore, test1 is wrong in both client and server, but could be some kotlin problem.

```
    @Query
    fun test1(list: @NonNull List<String>): String
```
Graphql:
```
test1(list: [String]): String
```

Note kotlin is a bit special with annotations https://youtrack.jetbrains.com/issue/KT-35843 we need an extra flag. 

Similar apis for client and server in java and kotlin.

[code-with-quarkus-nonnull.zip](https://github.com/smallrye/smallrye-graphql/files/8532439/code-with-quarkus-nonnull.zip)
[code-with-quarkus-nonnull-kotlin.zip](https://github.com/smallrye/smallrye-graphql/files/8532440/code-with-quarkus-nonnull-kotlin.zip)

phillip-kruger:
I think `test2(list: [String!]): String` means that the list itself can be null, but if not null, the elements in the list can not be null. Something like that.

[...]

PR #2141, opened 2024-07-03
Fixed wrapper nonnull types

mskacelik:
fixes:#1366
`NonNull` annotation is only applied now where it needs to be (in wrapper hierarchy)
There are some maybe weird behaviors when it comes to arrays: `@NonNull String[] => [String!]!` even tho it should be `[String]!`.

some other changes:
- primitive types that are elements of collection/arrays -> are now `nonNull` by default

/cc @jmartisk

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/Annotations.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/Annotations.java
@@ -512,17 +512,10 @@ public class Annotations {
 
     private static Map<DotName, AnnotationInstance> getAnnotations(org.jboss.jandex.Type type) {
         Map<DotName, AnnotationInstance> annotationMap = new HashMap<>();
-
-        if (type.kind().equals(org.jboss.jandex.Type.Kind.PARAMETERIZED_TYPE)) {
-            org.jboss.jandex.Type typeInCollection = type.asParameterizedType().arguments().get(0);
-            annotationMap.putAll(getAnnotations(typeInCollection));
-        } else {
-            List<AnnotationInstance> annotations = type.annotations();
-            for (AnnotationInstance annotationInstance : annotations) {
-                annotationMap.put(annotationInstance.name(), annotationInstance);
-            }
+        List<AnnotationInstance> annotations = type.annotations();
+        for (AnnotationInstance annotationInstance : annotations) {
+            annotationMap.put(annotationInstance.name(), annotationInstance);
         }
-
         return annotationMap;
     }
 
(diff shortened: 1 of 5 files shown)


jmartisk:
cc @robp94 if you want to help us verify this fix

robp94:
I will try to test it, since our services are still on quarkus 3.8.x I am not sure if I can test it with them.

jmartisk:
It should be possible to cherry-pick this to the 2.8.x branch (which is used by Quarkus 3.8.x).

Or use the branch of this PR, and build Quarkus from Marek's branch: https://github.com/mskacelik/quarkus/tree/srgql-2.9.0 - that one should be compatible with the smallrye-graphql main (you'll need to set the `smallrye-graphql.version` in `bom/application/pom.xml` to `2.9.0-SNAPSHOT`, build a Quarkus snapshot, and then build an application against the correct Quarkus snapshot version (which will be `999-SNAPSHOT`)

robp94:
I updated my old reproducers:
[code-with-quarkus-nonnull.zip](https://github.com/user-attachments/files/16084977/code-with-quarkus-nonnull.zip)
[code-with-quarkus-nonnull-kotlin.zip](https://github.com/user-attachments/files/16084981/code-with-quarkus-nonnull-kotlin.zip)

I had some errors building quarkus on Marek's branch, but probably not relevant for the reproducers. (Formatings on some kotlin modules)

The java reproducer now looks good, the kotlin one still fails. Should the kotlin one be fixed?

mskacelik:
Dont think so. I only fixed Java version.

jmartisk:
The code for handling Kotlin nullability is a different story, let's treat it separately. Could you file a separate issue please? (Maybe just copy the Kotlin-related stuff from the original one, and we will close that one)
Thanks for verifying!

brainlag:
Why was @AdaptWith support for input models dropped with this MR?

jmartisk:
@brainlag could you elaborate on what exactly used to work and now doesn't?

brainlag:
So we have ModelA with a field foo of ModelB . This field is annotated with @AdaptWith to change the type from ModelB to String which is the unique ID of  the object. But with this MR in `Bootstrap.java` around line 913 it was changed from  `referenceGraphQLInputType(field);` to  `getGraphQLInputType(field.getReference());` which bypasses the handling of all @AdaptXXX annotations. In the end the schema is wrong and we have `foo: ModelBInput` instead of `foo: String`.

jmartisk:
Ah you may be right, I'm not sure if that change on 913 was necessary, maybe @mskacelik remembers better.

Btw you may also want to try the newer `@CustomScalar` approach to map ModelB to a string scalar ( see https://smallrye.io/smallrye-graphql/latest/custom-scalar/)

brainlag:
If I understand it correctly @CustomScalar does not work for input models. Also I'm not sure it would work for all your use-cases.

jmartisk:
It should work for inputs (if you decide to give it a try and you encounter any problems, please let us know!)
Regarding AdaptWith though, I do think it is probably a bug to fix. Let's wait for @mskacelik's opinion

mskacelik:
Honestly, I am not sure why I made such a change on line 391. But I am more surprised that we didn't have tests for that. Thanks for reporting. Going to fix it ASAP.

Issue #2148, opened 2024-07-11
TCK failure with 2.9.0


jmartisk:
Seems 2.9.0 introduced is some issue with applying NumberFormat on items in (nested) lists.

The TCK test `trackHeroLongLat` is failing: https://github.com/eclipse/microprofile-graphql/blob/2.0/tck/src/main/java/org/eclipse/microprofile/graphql/tck/apps/superhero/api/HeroFinder.java#L419
When this mutation is called with `coordinates:[["-25.926804 longlat", "28.203392 longlat"],["-26.926804 longlat", "27.203392 longlat"],["-27.926804 longlat", "26.203392 longlat"]]`, it throws a validation error.

The same applies here: https://github.com/eclipse/microprofile-graphql/blob/2.0/tck/src/main/java/org/eclipse/microprofile/graphql/tck/apps/superhero/api/HeroFinder.java#L512C1-L512C111 when called with `dates:["01/14/2020","02/15/2021"]`

Weirdly, the TCKs only on the Quarkus side, not in SmallRye (this fact is probably worth investigating too). I added temporary overrides to have the test pass: see https://github.com/quarkusio/quarkus/blob/24326d05c1cb65cbee4a0a26080e9463c716d30c/tcks/microprofile-graphql/src/main/resources/overrides/importantDatesUS/output.json and https://github.com/quarkusio/quarkus/blob/24326d05c1cb65cbee4a0a26080e9463c716d30c/tcks/microprofile-graphql/src/main/resources/overrides/trackLongLat/output.json

To reproduce with Quarkus, remove the temporarily added `tcks/microprofile-graphql/src/main/resources/overrides/trackLongLat/output.json` and `tcks/microprofile-graphql/src/main/resources/overrides/importantDatesUS/output.json` files in the Quarkus repo and run the tests in the `tcks/microprofile-graphql` module.

mskacelik:
https://github.com/eclipse/microprofile-graphql/blob/93c98a656cdb176427e1c767bf1bd085b706791c/tck/src/main/java/org/eclipse/microprofile/graphql/tck/apps/superhero/api/HeroFinder.java#L418
It seems that this annotation has not been ignored. If I replace the `longlat` with `latlong` it works.

mskacelik:
it seems to be due to fa02098cabe9db9be480d16425e0336e7532680c.

PR #2152, opened 2024-07-25
fix TCK issue

mskacelik:
fix:#2148

The problem was that `getAnnotations(type)` only looked for the most outer wrapper; for example, for type `@A List<@B C>,` it only returned `A` but not `A, B`. This was because I changed it in the https://github.com/smallrye/smallrye-graphql/commit/fa02098cabe9db9be480d16425e0336e7532680c commit.

That said, I am not sure why the SRGQL TCK did not fail and Quarkus did...

I will delete the overridden configuration (outputs.json) on Quarkus when a new version of SRGQL is released...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,8 +96,8 @@ jobs:
       - name: checkout Quarkus repository
         uses: actions/checkout@v2
         with:
-          repository: quarkusio/quarkus
-          ref: main
+          repository: mskacelik/quarkus
+          ref: srgql-tck-fix
           path: quarkus
 
       - uses: actions/setup-java@v3.10.0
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -96,8 +96,8 @@ jobs:
       - name: checkout Quarkus repository
         uses: actions/checkout@v2
         with:
-          repository: quarkusio/quarkus
-          ref: main
+          repository: mskacelik/quarkus
+          ref: srgql-tck-fix
```
jmartisk, line 100: could you name it `srgql-2.9.1` or something? It's quite likely that we will need to add more changes to that branch...

PR #2163, opened 2024-08-26
changed back Quarkus test source, since the 2.9.1 is already released

mskacelik:
/cc @jmartisk


jmartisk:
thanks, I already have https://github.com/smallrye/smallrye-graphql/pull/2162 where I'm also trying to figure out why the quarkus build is failing (it's using an old Maven version, I think)

PR #2169, opened 2024-08-27
Add api group names for typesafe clients

RoMiRoSSaN:
See #2173

Server implementation allows create api groups
```
@Name("Group")
@GraphQLApi
public class Api {
    /////
}
```
and create queries like
```
query {
    Group {
        .......
    }
}
```
But typesafe client has not this. This code 
```
@Name("User")
@GraphQLClientApi
public interface SuperHeroesApi {
    List<User> users(List<Long> id);
}
```
build not correct query (it not conains User wrapper on methods)

This PR fix it

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/QueryBuilder.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/QueryBuilder.java
@@ -27,6 +27,12 @@ public class QueryBuilder {
         if (method.hasValueParameters())
             request.append(method.valueParameters().map(this::declare).collect(joining(", ", "(", ")")));
 
+        String groupName = method.getGroupName();
+        if (groupName != null) {
+            request.append(" { ");
+            request.append(groupName);
+        }
+
         if (method.isSingle()) {
             request.append(" { ");
             request.append(method.getName());
@@ -40,6 +46,9 @@ public class QueryBuilder {
 
         if (method.isSingle())
             request.append(" }");
+
+        if (groupName != null)
+            request.append(" } ");
 
         return request.toString();
     }
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/ResultBuilder.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/ResultBuilder.java
@@ -92,7 +92,10 @@ public class ResultBuilder {
     private JsonObject readData() {
         if (!response.containsKey("data") || response.isNull("data"))
             return null;
-        JsonObject data = response.getJsonObject("data");
+        String groupName = method.getGroupName();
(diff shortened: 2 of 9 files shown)


jmartisk:
Hi, thanks for your PR. But TBH I'm not yet completely sure we want this.

The grouping feature on the server side probably didn't prove very useful (it's not how GraphQL schemas are usually created) - maybe @phillip-kruger has more thoughts on that, as he implemented it? We also removed it from the docs because we don't really recommend using it.

On the client side, this would be closely tied to how specifically the SmallRye GraphQL server side does it, but mostly useless for calling GraphQL services based on anything else.

RoMiRoSSaN:
Hi Jan. I can't agree with. Grouping api by any type can create structured api. For example, in rest we can create api like 
/crm/product/create
/admin/user/create
But by default in smallrye-graphql if not do groups (namespaces) methods can have unreadeble names (crmProductCreate, adminUserCreate)

Also for example, in nodejs you can do schema like (or more nesting)
```
type Query {
   crm: Crm
}

type Crm {
   user: User
}

type User {
   create: Response
}
```

I think it may be very useful for big api if create it right. Becouse in my project there are already about 1k endpoints... For example, there is methods like getCustomerRequestsManagementServiceCustomerRequestPositionsById. This is pain))

The Graphql spec allows namespaces. And I think that removed it may be not best idea, becouse it can be someone use.
As for this PR.  we using third party api, which uses namespace types. And absence of this on typesafe client forces us using dynamic client.

jmartisk:
Could you add a note into the documentation please? (probably into `typesafe-client-usage.md`)

RoMiRoSSaN:
@jmartisk yes. of course

RoMiRoSSaN:
@jmartisk please, check it

jmartisk:
I can't see any documentation changes, maybe you didn't push it?
Also, I know I'm probably a bit annoying, but could you add a simple test? :) For example `ClientModelBuilderTest` may be a good place to do this without it getting too complex
Otherside, looks great!

RoMiRoSSaN:
@jmartisk please, sorry. I made a mistake with PR 😑. 
Yes, I all wiil do here. I think I wiil do it today

RoMiRoSSaN:
@jmartisk I have added what I think is enough detail about namespaces for both server and client. Please take a look. Maybe you'll tell me that everything is bad =)

jmartisk (approved):
Absolutely outstanding! Thanks!

jmartisk:
Btw I'm planning to release SR-GQL 2.10 around this Friday, to catch the Quarkus 3.15 release train, so if you have any more stuff that you think should be included, please let me know, I may forget things...

RoMiRoSSaN:
@jmartisk #2172 maybe too?

jmartisk:
Yeah, merged that one too, thanks

PR #2177, opened 2024-09-02
Client model union/interface implementation

mskacelik:
/cc @jmartisk @t1 

Client model implementation of Union and Interfaces

### DIFF
--- a/client/implementation-vertx/pom.xml
+++ b/client/implementation-vertx/pom.xml
@@ -71,4 +71,33 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <executions>
+                    <execution>
+                         <!--This configuration separates the execution of the `TypesafeTckClientModelSuite` tests
+                         from the default test execution (TypesafeTckSuite) to avoid conflicts or issues that might
+                         arise when executing the same tests multiple times in different test suites.-->
+                        <id>default-test</id>
+                        <configuration>
+                            <excludes>TypesafeTckClientModelSuite</excludes>
+                        </configuration>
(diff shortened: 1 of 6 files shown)


client/implementation-vertx/pom.xml
```diff
@@ -71,4 +71,31 @@
...
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <configuration>
+                            <excludes>TypesafeTckClientModelSuite</excludes>
+                        </configuration>
+                    </execution>
+                    <execution>
```
jmartisk, line 86: Please add a comment explaining why we're doing this, otherwise I won't remember it after two weeks 
(that junit doesn't seem to like executing the same tests separately in multiple suites or something)

client/model-builder/src/main/java/io/smallrye/graphql/client/model/helper/TypeModel.java
```diff
@@ -361,6 +365,23 @@ public boolean isTypeVariable() {
         return Classes.isTypeVariable(type);
     }
 
+    /**
+     * Checks if the type is a Union.
+     *
+     * @return {@code true} if the type is a Union, otherwise {@code false}.
+     */
```
t1, line 372: This javadoc explains nothing that isn't already in the method name. I'd rather delete it. Or link to the documentation what a GraphQL Union is: https://spec.graphql.org/draft/#sec-Unions

t1:
```suggestion
    /**
     * @see <a href="https://spec.graphql.org/draft/#sec-Unions">GraphQL Union</a>
     */
```

t1 (approved):
LGTM. Thanks!

jmartisk (approved):

PR #2179, opened 2024-09-04
bump version of graphql-java 22.2

mskacelik:
/cc @jmartisk 

There still is a need to update the Quarkus/WildFly branch later – I am not sure if this commit is going to be part of 2.9.3 or 2.10.0.


jmartisk:
Thanks! The current `main` is already meant for 2.10, for potential 2.9.3 we would have to backport later

jmartisk (approved):

Issue #2110, opened 2024-05-16
Unable to generate Federation schema with no @Query functions


nicholassmith:
Maybe this is a bit edge case but it's valid for federation and I can't quite figure out how to achieve it. I would like to generate a schema like:

```
union _Entity = Shelf

type Books {
   name: String!
   shelf: Shelf
....

"Query root"
type Query {
   _entities(representations: [_Any!]!): [_Entity]!
   _service: _Service!
}

...
```

However with smallrye-graphql I need at least 1 query for it to build the schema and then invoke the Federation transformer to add the additional `entities` query that allows for this usecase. The only thing I can think of is that when it checks `schema.hasOperations())`  in `Bootstrap.java` it should also check to see if it will generate the `entities` query.

RoMiRoSSaN:
@jmartisk 
Hi Jan, I tagged you, becouse you created PR before.

I have encountered this problem too. At the moment samllrye graphql does not support working with federations in full.
The solution #2113 does not work, because the schema is validated before its transformation.
I came up with a slightly strange solution to get around this, but it definitely works and does not break anything.
Only at the moment it makes no sense, because, as it seems to me, support for federation is not fully implemented.

An example of a schema using federation (maybe not entirely correct, but it reflects the essence)
```
schema {
    query: Query
}

union _Entity = Role

type Query {
    _entities(representations: [_Any!]!): [_Entity]!
    _service: _Service
}

type _Service {
    sdl: String!
}

type Role @extends @key(fields : "id") {
    id: String @external
    nameFromOtherService: String
}

scalar _Any
```

The option when only id information is returned in this type definitely works. When Role is a field of another type, from the request.

However, according to the federation specification, the schema may not contain any requests. But it can extend types, such as Role.

At the moment, this can be done a little incorrectly. For example, add an explicit request.
```
// User service
@GraphQLApi
public class Resource {
    @Query
    public User byId(Long id){
        ///
    }
    
    // This query should not exist
    @Query
    public Role findExtendedRole(Long id) {
        // get role by id
        // set needed fields
        // return
    }
}
```

Most likely, this type of variant will not work

```
type Role @extends @key(fields : "id") {
    id: String @external
    name: String @exteranl
    .... more external provides required
    nameFromOtherService: String
}
```

Or there is nothing about this in the documentation and I just did not find / did not understand how it works. Or it really is not implemented.
This variant definitely does not work. Information about the Role type does not get into the schema, fetcher is not created.
```
// User service
[...]

PR #2180, opened 2024-09-04
Handling federation queries without creating @Query

RoMiRoSSaN:
See #2110


server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -214,9 +219,20 @@ private void generateGraphQLSchema() {
...
             log.enableFederation();
+
+            if (!(schema.hasQueries() || schema.hasResolvers() || schema.hasMutations() || schema.hasSubscriptions())) {
+                throw new RuntimeException("You must define at least one method marked with one of the annotations - "
+                        + "@Query, @Mutation, @Subscription, @Resolver. An empty diagram has no meaning");
+            }
+
+            // hack! For schema build success if queries are empty.
+            // It will be overrides in Federation transformation
+            addSdlQueryToSchema(schemaBuilder, queryRootType);
```
t1, line 230: I‘m not sure, if I got this right: graphql-java federation transformation is going to overwrite the `_service` query that we add here? It‘s just a dummy that is required to make the schema builder happy? Maybe it would make the code flow more understandable, if the method was named `addDummyQuery`. What do you think?

t1 (approved):
The code looks good to me. I just don‘t quite get, what the resolvers are needed for. I’m happy to hear that you want to add documentation, Javadoc and tests.

RoMiRoSSaN:
I found this #1585, #521 
Here how I understand was annotation FederationSource, but now smallrye graphql not contains it. The logic there intersects with my PR. For example, in NodeJS federation router itself sends requests about what and what fields it needs. And there it can be any object (list of parameters). 
At the moment I have not yet come up with how creeate a universal wrapper object (method must have all required 'primitive' params. see test). With these changes, the router can obtain the required data. And most importantly, queries are not published to the schema. Exmaple, service extends external type other service
```
extend type Type key('id') {
  id: String external
  name: String external
  value: String requries('name')
}
```
If you request a field `value` will be sended request like
```
_entities(representations: { id: "id", name: "name", __typename: "Type" }) {
  __typename
  ... on ExtendedType {
    value
  }
}
```
value extended field. It turns out you need to make 2 requests, by id (default and may be query), and id and name (value requires field name). 
And if you publish so many queries to the schema, it can become very large. And it may not be necessary for requests to be available exclusively to the federation router

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
RoMiRoSSaN: yes, you right, i rename this

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -124,7 +125,7 @@ public static GraphQLSchema bootstrap(Schema schema) {
     }
 
     public static GraphQLSchema bootstrap(Schema schema, boolean skipInjectionValidation) {
-        if (schema != null && (schema.hasOperations())) {
+        if (schema != null && (schema.hasOperations()) || Config.get().isFederationEnabled()) {
```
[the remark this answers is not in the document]
RoMiRoSSaN: I was wrong here  (...). I fixed it

RoMiRoSSaN:
It would probably be nice to have such an option.
```
record Vars(String id, String name){}
@Resolver
Product find(Vars vars){
   ...
}
```
and allow pass queries like 
```
_entities(representations: { id: "id", name: "name", __typename: "Product " }) { __typename  ... on ExtendedType { id name value } }

_entities(representations: { id: "id", __typename: "Product " }) { __typename  ... on ExtendedType { id} }
```
in one method.

RoMiRoSSaN:
Unfortunately, I couldn’t figure out how to make one general type. Too much magic

Needs simple changes in quarkus (add resolvers operations [here](https://github.com/quarkusio/quarkus/blob/main/extensions/smallrye-graphql/deployment/src/main/java/io/quarkus/smallrye/graphql/deployment/SmallRyeGraphQLProcessor.java#L490))

RoMiRoSSaN:
Hi @jmartisk, @t1. What do you think about it?
For example, [This example in JS](https://github.com/apollographql/federation-demo).  And in [Spring](https://github.com/apollographql/federation-jvm-spring-example/tree/main)

jmartisk:
Hi, sorry a bit busy, but I'll try to review at the beginning of next week...

RoMiRoSSaN:
This will require changes that are present in https://github.com/smallrye/smallrye-graphql/pull/2184

jmartisk:
Ok, would you also please create a PR to my Quarkus branch (https://github.com/jmartisk/quarkus/tree/smallrye-graphql-2.11) with the Quarkus-side change and preferably also a simple test that runs with Quarkus?

RoMiRoSSaN:
Added changes and a couple of simple tests to your fork and create PR

jmartisk:
Ok, merged the Quarkus part, now re-running the CI here... Thanks!

RoMiRoSSaN:
@jmartisk It seems like I did everything, made all the necessary corrections.

jmartisk (approved):

jmartisk:
Thanks!

Issue #2173, opened 2024-08-28
Graphql namespaces


RoMiRoSSaN:
I decided open issues after discussion with @jmartisk in PR #2169 
Jan said that using namespaces are not recomended. However grpahql spec allows it
Default behavior of graphql and graphql-client does not raise questions. 

Server part allows now using annotation Name for grouping apis. I think it is not bad if someone whant grouping api in same namespaces.

At this moment I found some bugs related with namespaces

1. Uncorrect description - #2168 
2. Typesafe client can not working with Name annotation - #2169 
3. FederationDataFetcher also working not corretly with namespaces - #2172

May be I can found something else..

So, I want discuss them. What are think about it?
Thanks

phillip-kruger:
> However grpahql spec allows it

Can you point to where in the spec this is defined ?

RoMiRoSSaN:
Hi. I carefully reviewed the specification and did not find a description of this. However, there is a fairly old discussion on this topic. As well as documentation from Apollo on this topic.
https://github.com/graphql/graphql-spec/issues/163
https://www.apollographql.com/docs/technotes/TN0012-namespacing-by-separation-of-concern

phillip-kruger:
@jmartisk - your call.

jmartisk:
I don't know man, I'm a bit hesitant to go through with this approach to namespaces when there's an open proposal to add namespacing to GraphQL - see https://github.com/graphql/graphql-spec/issues/163 and https://medium.com/@oleg.ilyenko/graphql-namespaces-proposal-2a097ce81d2a - the approach in the second link is quite different as it uses a `/` syntax to denote namespaces, so it may happen that this will be added to GraphQL in the future, and we would be then using a different, incompatible approach :/

Btw I can imagine a way to do this with the typesafe client too:

With a query like

```
Namespace {
   customersList(someArgument: Int) {
     name
   }
}
```

you, I think, could do

```
class Namespace {
  // empty
}

@GraphQLClientApi
public class MyClient {

  @Query('Namespace')
  Namespace namespace();
  
  List<Customer> customersList(@Source Namespace namespace, Integer someArgument);
  
}
```

RoMiRoSSaN:
Hi. I tested it now. Yes, it can work. 
1 - namespaces can be created with using empty classes and Source methods. It is not bad example of how nested namespaces can be done, but it has complexity. Here you need to explicitly separate the code into query/mutations, which can add more code and confusion
[...]

PR #2184, opened 2024-09-06
Add @Namespace for GraphQLApi and GraphQLClientApi

RoMiRoSSaN:
This PR is only open to see what changes may be needed to changes suggestion by @t1 from this discussion #2173.
For viewing only, not for merge.


common/schema-model/src/main/java/io/smallrye/graphql/schema/model/GroupContainer.java
```diff
@@ -0,0 +1,93 @@
...
+            add(name, description, queue, operation);
+        } else {
+            throw new RuntimeException("Namespaces can't be empty");
+        }
+    }
+
+    private void add(String name, String description, ArrayDeque<String> queue, Operation operation) {
+        this.name = name;
+
+        if (!queue.isEmpty()) {
```
t1, line 76: I don't like not-operations that are not absolutely necessary, so maybe switch the `if`?

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Schema.java
```diff
@@ -96,51 +97,57 @@ public boolean hasSubscriptions() {
     }
 
     public Map<Group, Set<Operation>> getGroupedQueries() {
-        return groupedQueries;
+        Set<Operation> operations = namespaceQueries.values().stream()
+                .map(GroupContainer::getAllOperations)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet());
+        return Map.of(new Group(), operations);
```
t1, line 104: I understand that you moved the nesting logic into the `GroupContainer` class, but I don't really get, why we then still need a `Map` with an "empty" `Group`... wouldn't it be enough to somehow work with a `GroupContainer`? Maybe this could simplify a lot of things in here.

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Schema.java
```diff
@@ -22,9 +24,8 @@ public final class Schema implements Serializable {
     private Set<Operation> mutations = new HashSet<>();
     private Set<Operation> subscriptions = new HashSet<>();
 
-    private Map<Group, Set<Operation>> groupedQueries = new HashMap<>();
-    private Map<Group, Set<Operation>> groupedMutations = new HashMap<>();
-    private Map<Group, Set<Operation>> groupedSubscriptions = new HashMap<>();
+    private Map<String, GroupContainer> namespaceQueries = new HashMap<>();
+    private Map<String, GroupContainer> namespaceMutations = new HashMap<>();
```
t1, line 28: `namespaceX` → `namespacedX` (with the `d`) would make more sense

server/api/src/main/java/io/smallrye/graphql/api/Namespace.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.smallrye.common.annotation.Experimental;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+@Documented
+@Experimental("Experimental namespaces on api class. Separate namespaces via '/' (example - admin/users)")
```
t1, line 14: I would move the second sentence into Javadoc. And add a hint that the class name is being used as the default `value`.

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -389,18 +386,30 @@ private void addRootObject(GraphQLObjectType.Builder rootBuilder, Set<Operation>
...
     }
 
-    private void addGroupedRootObject(GraphQLObjectType.Builder rootBuilder,
-            Map<Group, Set<Operation>> operationMap, String rootName) {
-        Set<Map.Entry<Group, Set<Operation>>> operationsSet = operationMap.entrySet();
+    private String prettyString(String value) {
+        return value.substring(0, 1).toUpperCase() + value.substring(1);
+    }
+
+    private void addNamespacedRootObject(GraphQLObjectType.Builder mutationBuilder,
```
t1, line 393: The parameters can also be for queries, can't they? Maybe better rename it to, e.g., `operationBuilder`/`namespaceOperations`/`operationType`?

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -389,18 +386,30 @@ private void addRootObject(GraphQLObjectType.Builder rootBuilder, Set<Operation>
...
-            Group group = operationsEntry.getKey();
-            Set<Operation> operations = operationsEntry.getValue();
+    private List<GraphQLFieldDefinition> addNamespacedRootObject(GraphQLObjectType.Builder rootBuilder,
+            GroupContainer groupContainer, String rootName, String suffix) {
+        List<GraphQLFieldDefinition> graphQLFieldDefinitions = groupContainer.getContainer().isEmpty()
+                ? List.of()
+                : getGraphQLFieldDefinition(groupContainer, rootName, suffix);
 
-            GraphQLObjectType namedType = createNamedType(rootName, group, operations);
+        if (!groupContainer.getOperations().isEmpty() || !graphQLFieldDefinitions.isEmpty()) {
```
t1, line 405:
```suggestion
        if (groupContainer.getOperations().isEmpty() && graphQLFieldDefinitions.isEmpty()) {
            return List.of();
        }
```

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
```diff
@@ -409,28 +418,53 @@ private void addGroupedRootObject(GraphQLObjectType.Builder rootBuilder,
...
             operation = eventEmitter.fireCreateOperation(operation);
 
-            GraphQLFieldDefinition graphQLFieldDefinition = createGraphQLFieldDefinitionFromOperation(namedTypeName,
-                    operation);
+            GraphQLFieldDefinition graphQLFieldDefinition = createGraphQLFieldDefinitionFromOperation(
+                    namedTypeName, operation);
             objectTypeBuilder = objectTypeBuilder.field(graphQLFieldDefinition);
         }
 
+        objectTypeBuilder.fields(graphQLFieldDefinitions);
```
t1, line 467: isn't the second one unnecessary?

server/implementation/src/test/java/io/smallrye/graphql/test/namespace/SourceNamespaceTestApi.java
```diff
@@ -2,26 +2,14 @@
 
 import org.eclipse.microprofile.graphql.GraphQLApi;
 import org.eclipse.microprofile.graphql.Query;
-import org.eclipse.microprofile.graphql.Source;
+
+import io.smallrye.graphql.api.Namespace;
 
 @GraphQLApi
+@Namespace("first/second")
```
t1, line 9: This is _so_ much nicer! 😁

t1 (approved):
I like it!

Most of the changes are _not_ from the `@Name` → `@Namespace` change, but from using "namespace" instead of "group" and the fact that they can be nested.

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/MethodInvocation.java
```diff
@@ -262,18 +266,25 @@ public String getOperationTypeAsString() {
...
+            return Arrays.stream(annotation.value().split("/"))
+                    .map(String::trim)
+                    .collect(Collectors.toList());
         }
-        return null;
+        return List.of();
+    }
+
+    public String getOperationName() {
+        return namespaces.stream().map(this::prettyString).collect(joining()) + prettyString(getName());
```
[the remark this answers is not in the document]
RoMiRoSSaN: For example
```
@Namespace("first/second/third/someMynotherName")
class Api {
   @Query("findAll")
   ////
}
```

If don create first letter to upper case, operation will firstsecondthirdsomeMynotherNamefindAll. it is unreadable.
And create it only findAll incorrect, becouse in code may be other findAll method. And if we will call just findAll, using some router (with analytics), it can break call analytics.

common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
```diff
@@ -143,11 +144,29 @@ private Schema generateSchema() {
...
+        if (!errors.isEmpty()) {
+            throw new RuntimeException(String.join("\n", errors));
+        }
+
         for (AnnotationInstance graphQLApiAnnotation : graphQLApiAnnotations) {
             ClassInfo apiClass = graphQLApiAnnotation.target().asClass();
             List<MethodInfo> methods = getAllMethodsIncludingFromSuperClasses(apiClass);
-            Optional<Group> group = GroupHelper.getGroup(graphQLApiAnnotation);
-            addOperations(group, schema, methods);
+            Optional<Group> namespace = GroupHelper.getGroup(graphQLApiAnnotation);
```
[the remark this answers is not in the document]
RoMiRoSSaN: I just for working quarkus localy) quarkus using group class. Of course Namespace is correct

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Schema.java
RoMiRoSSaN: just for quarkus

RoMiRoSSaN:
I don't like it in this form yet. I think it would be better to roll back the changes. And expand the existing code. To enable work with Namespace, you need to add a config to switch to working with them. And by default, use what is there. With this approach, Quarkus will not break. And it will be possible to refine it (add additional configs, checks) without any problems.

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/MethodInvocation.java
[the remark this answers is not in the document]
t1: Okay, I think I get it: if the user wants to send multiple operations in one request, their names have to be distinct. But with the typesafe client, you can send multiple request only with a `@Multiple` annotation. The operation name is only scoped within a query string, not globally. I think we can safely skip this. Try to write a test to prove me wrong 😜
BTW, this code _does_ uppercase the very first letter: `FirstSecondThirdSomeMynotherName` ... but the `F` could also remain `f` 😁

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Schema.java
t1: I love Quarkus, but making this code simpler also makes the Quarkus code simpler, wouldn't it?

t1:
I just wanted to repeat it: thanks, @RoMiRoSSaN. You're doing a great job!

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/MethodInvocation.java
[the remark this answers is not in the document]
RoMiRoSSaN: When the Quarkus tests failed, I saw that Quarkus also checks the operation name. 
And maybe it’s worth making sure that the operation name is formed like this:
1. If there is `@Name` or `@Namespace`, capitalize the first letters. 
2. If not, leave it as is.

As a result, the quarkus build will not crash, and there will be a minimum changes in the tests.
As for `@Multiple`, I'll take a look today.

common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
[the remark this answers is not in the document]
RoMiRoSSaN: Just in deployment for build reflection classes [here](https://github.com/quarkusio/quarkus/blob/main/extensions/smallrye-graphql/deployment/src/main/java/io/quarkus/smallrye/graphql/deployment/SmallRyeGraphQLProcessor.java#L490)
As I understand it, if these changes are adopted in the future, namespaced operations will need to be added here.

Or for all operations (queries, grouped queries, namespaced queries, etc..) create a method in the Schema class that will return everything at once in one set. And this place will need change. Then in the future, when adding new types of operations, you will not need to make changes here.

common/schema-model/src/main/java/io/smallrye/graphql/schema/model/Schema.java
RoMiRoSSaN: I returned this part as it was (for Group). 
As for Map<String, NamesapceContainer> - I will try change it for Set

server/implementation/src/main/java/io/smallrye/graphql/bootstrap/Bootstrap.java
RoMiRoSSaN: here passed nested objects. 
`@Namespace("admin")`, `@Namespace("admin/users")`. Each contains some operations. Here builds admin definition (with its opeartions), and in the next adds definitions 'users' and etc..

RoMiRoSSaN:
Hi, @t1 

Summary at this moment
1. You can use `@Name` or `@Namepace` on the server and client code.
2. If both annotations are used - an error.
3. Needs simple changes in quarkus.

I only have a question about the operation names. Quarkus tests fails on the check operation name. How should it be done? 
Doing it by name is not correct if there are namespaces. I made the first letter capitalize. I had to make some minor changes to the tests. Now I think it makes sense to put this back for situations where there are no namespaces. And Quarkus tests will not fail, and there are fewer internal tests to edit.
```
@Namespace("fisrt/second")
@GraphQLClientApi
interface Api {
   @Query
   String get();  // Operation name will be 'FirstSecondGet'
}

@Name("fisrt")
@GraphQLClientApi
interface Api {
   @Query
   String get();  // Operation name will be 'FirstGet'
}

@GraphQLClientApi
interface Api {
   @Query
   String get();  // Operation name will be 'get'
}
```

And I'm waiting for your questions and suggestions

RoMiRoSSaN:
@jmartisk Please, look at this later too. When you have time

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/MethodInvocation.java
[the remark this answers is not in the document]
t1: I'm obviously biased by the Java naming convention that methods start with a lowercase character. That's why I'd like to stick with that, also for operation names. I.e., also if there is a `@Namespace` annotation, the very first character should be lowercase. E.g., `@Namespace("users") ... @Query List<User> all() {...` would result in an operation name `usersAll`.

common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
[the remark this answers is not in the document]
t1: That sounds good. Do you think that would be a terrible lot of work?

t1:
If we would follow my comment above, the names would be `firstSecondGet`, `firstGet`, and `get`.

common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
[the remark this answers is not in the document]
RoMiRoSSaN: No, just a little

client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/MethodInvocation.java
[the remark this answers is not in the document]
RoMiRoSSaN: I just referred to examples from the graphql documentation https://graphql.org/learn/queries/#operation-name
But of course I can change it to your example, no problem.

client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientBuilder.java
```diff
@@ -185,7 +195,8 @@ public <T> T build(Class<T> apiClass) {
                 endpoint,
                 websocketUrl, executeSingleOperationsOverWebsocket, httpClient, webClient, subprotocols,
                 websocketInitializationTimeout,
-                allowUnexpectedResponseFields);
+                allowUnexpectedResponseFields,
+                namespaceAnnotation != null);
```
jmartisk, line 199: Why would you pass a `useNamespace` boolean when you have already looked at the annotation and can now already just pass the information about the particular namespace as a `List<String>`? That would avoid having to use reflection later on.

client/model-builder/src/test/java/io/smallrye/graphql/client/model/ClientModelBuilderTest.java
```diff
@@ -235,13 +236,40 @@ void namedClientModelTest() throws IOException {
...
                 new MethodKey("findAllStringsName", new Class[0]),
-                "query findAll { named { findAll } } ");
+                "query NamedFindAll { named { findAll } }");
         assertOperation(clientModel,
                 new MethodKey("update", new Class[] { String.class }),
-                "mutation update($s: String) { named { update(s: $s) } } ");
+                "mutation NamedUpdate($s: String) { named { update(s: $s) } }");
+    }
+
+    @Namespace("first/second")
```
jmartisk, line 248: I haven't been watching the discussion much, so maybe this was already discussed, but wouldn't it be better to declare the namespace components as an array (`@Namespace({"first","second"})`? That would be more typesafe and avoid having to parse it at runtime

client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientBuilder.java
```diff
@@ -145,6 +147,14 @@ public VertxTypesafeGraphQLClientBuilder websocketInitializationTimeout(Integer
 
     @Override
     public <T> T build(Class<T> apiClass) {
+        Name nameAnnotation = apiClass.getAnnotation(Name.class);
+        Namespace namespaceAnnotation = apiClass.getAnnotation(Namespace.class);
+
+        if (nameAnnotation != null && namespaceAnnotation != null) {
+            throw new RuntimeException("You can only use one of the annotations - @Name or @Namespace " +
+                    "over the GraphQLClientApi interface. Please, fix next interface: " + apiClass.getName());
```
jmartisk, line 155:
```suggestion
                    "over the GraphQLClientApi interface. Please, fix the interface: " + apiClass.getName());
```

common/schema-builder/src/main/java/io/smallrye/graphql/schema/SchemaBuilder.java
```diff
@@ -168,6 +184,76 @@ private Schema generateSchema() {
...
+                        errors.add("class: " + apiClass.name().toString() + ", method: " + methodInfo.name());
+                    }
+                }
+            }
+        }
+
+        if (!errors.isEmpty()) {
+            throw new RuntimeException("Subscriptions can't be nested. " +
+                    "Move your subscriptions to another @GraphQLApi class, not marked @Namespace or @Name. " +
+                    "Check next places: " + String.join("; ", errors));
```
jmartisk, line 240:
```suggestion
                    "Check these places: " + String.join("; ", errors));
```

RoMiRoSSaN:
@jmartisk 
Replaced the string with an array of strings. As a result, it was possible to remove the code for groups and other things without any problems. 
Good point about useNamespace, I've rollback it.
Managed to maintain backward compatibility with `@Name`
I checked these changes in quarkus - minor edits will be required.

client/model-builder/src/main/java/io/smallrye/graphql/client/model/ClientModelBuilder.java
```diff
@@ -75,6 +80,19 @@ private ClientModels generateClientModels() {
...
 
+    private void validateAnnotations(Collection<AnnotationInstance> graphQLApiAnnotations) {
+        List<String> errorInterfaces = graphQLApiAnnotations.stream()
+                .map(annotation -> annotation.target().asClass())
+                .filter(classInfo -> classInfo.hasDeclaredAnnotation(NAMESPACE) && classInfo.hasDeclaredAnnotation(NAME))
+                .map(classInfo -> classInfo.name().toString())
+                .collect(Collectors.toList());
+        if (!errorInterfaces.isEmpty()) {
+            throw new RuntimeException("You can only use one of the annotations - @Name or @Namespace " +
+                    "over the GraphQLClientApi interface. Please, fix next interfaces: " +
```
jmartisk, line 91:
```suggestion
                    "over the GraphQLClientApi interface. Please, fix the following interfaces: " +
```

client/model-builder/src/main/java/io/smallrye/graphql/client/model/ClientModelBuilder.java
```diff
@@ -75,6 +80,19 @@ private ClientModels generateClientModels() {
         return clientModels;
     }
 
+    private void validateAnnotations(Collection<AnnotationInstance> graphQLApiAnnotations) {
```
jmartisk, line 83: Could we name this `validateNamespaceAnnotations` to make it more clear what it does?

jmartisk:
Have you got a branch with the planned Quarkus-side changes too so that I can have a look?

RoMiRoSSaN:
@jmartisk Hi. Here are the changes with comments. They are minimal. The main thing is to remove group processing. 
Just in case - works without adding `@Namesapce` to the index and processing operations. I checked this in the native assembly. But most likely you should add these lines (subtleties of native assemblies, probably)

https://github.com/quarkusio/quarkus/commit/bda9a2f14655f6b406a6b350cc5fdc859bf0dd50

jmartisk:
Yeah let's put the `@Namespace` annotation into the index to be sure,
and also the `classes.addAll(getOperationClassNames(schema.getAllOperations()))` call - I'm not sure what exactly you checked, but it should be necessary, at minimum in cases where GraphQL APIs work with non-scalar classes, I think. We need to make sure that the classes are registered for reflection.
If you could update your Quarkus commit per these comments, I would cherry-pick it and create a Quarkus branch that contains it, so we can later switch the CI (the "Quarkus tests" job) to run against that branch and get a full test run

jmartisk:
@RoMiRoSSaN I took the liberty to copy it into my own branch: https://github.com/jmartisk/quarkus/tree/smallrye-graphql-2.11 this branch will serve to capture any necessary changes that we will do in Quarkus along with upgrading to SmallRye GraphQL 2.11 (when it's released). I've also pushed a commit here to change the CI build so that the Quarkus part of the CI runs with a snapshot from that branch  - so it should reflect the Quarkus-side changes. Let's see what the CI says now.

RoMiRoSSaN:
Everything seems to be fine. 
Please, if you are ready to accept this, look carefully at the changes again, I am ready to accept all comments.

jmartisk:
Ok I think this is pretty much good to go. @RoMiRoSSaN please mark it as ready if you agree

jmartisk (approved):

jmartisk:
I've rebased it and squashed some commits, let's run the CI one more time

Issue #2154, opened 2024-08-01
@link not working for quarkus


robp94:
The GraphQL test works, so maybe the error is somewhere in quarkus/implementation-cdi? But I can not find it.

Quarkus is just the code start with enabled federation via application parameters. Furthermore, I added the @link like in the link tests.

```
@GraphQLApi
@Link(url = FEDERATION_SPEC_URL + "/v2.1", _import = { @Import(name = "@policy") })
public class HelloGraphQLResource {
```
in the Schema builder the link is not included, in the index view there is a link annotation.
![image](https://github.com/user-attachments/assets/ff7bc48d-3edc-4e1f-9f88-3bfe4ed66d04)
![image](https://github.com/user-attachments/assets/95a5a7b5-57e6-472c-969e-ae4ea771efc5)

While in the graphql test, we get the link from the index:
![image](https://github.com/user-attachments/assets/3e8cf659-e3da-42ff-b28c-f314bb42b082)

Furthermore, as far as I understand it without the @link it is not federation v2, but with the recent changes it is neither v1. So federation is currently unusable.

@jmartisk could you have a look pls? Or maybe @RoKKim?

robp94:
I did a bit more research on this, as far as I found out there is apollo federation and open federation. Open federation tries to be compatible with apollo.

For federation v2 apollo needs the @link directive. In wundergraph which uses open federation @link is not needed.

We use rover cli to combine our graphs and test them. Even if we add the @link manually we still need to make sure that the old directives are deleted like @key, and we import the ones from apollo.

In my opinion, we will need more config options to define what federation the user needs. V1? V2? apollo vs open. Since open should be compatible, apollo maybe we start with only that. Furthermore, depending on the version of v2 different directives are supported. As included in @link implementation.

SimFlensted:
Just wanna add to this issue.
We are having problems that I believe are caused by the lack of federation v2 support. We are getting the following build error in federation: `Invalid definition for directive "@key": argument "fields" should have type "_FieldSet!" but found type "FieldSet!"`, which I believe is caused because of a federation v1 schema using a v2 FieldSet type.
We have to use federation v2, and we also cannot get the @Link annotation working. 
@robp94 Did you find a workaround for adding this link, or did you do it manually in the schema (outside the generated one)?

robp94:
[...]

PR #2188, opened 2024-09-13
Fix directives building 

RoMiRoSSaN:
See #2154 

In wundergraph cosmo `@link` is optional. 
Apollo requires it. Quarkus does not pass a complete index containing all directives, resulting in the `@link`and some others being ignored.

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/DirectiveIndexer.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/DirectiveIndexer.java
@@ -1,0 +1,73 @@
+package io.smallrye.graphql.schema;
+
+import java.io.IOException;
+
+import org.jboss.jandex.IndexView;
+import org.jboss.jandex.Indexer;
+
+import io.smallrye.graphql.api.OneOf;
+import io.smallrye.graphql.api.federation.Authenticated;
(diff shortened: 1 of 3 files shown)


RoMiRoSSaN:
Fell due to the fact that the Deprecated and OneOf directives were added. You can remove them, or simply remove 2 checks in SchemaBuilderTest testConcurrentSchemaBuilding

jmartisk:
@RoMiRoSSaN, this change should be done on the Quarkus side, see https://github.com/quarkusio/quarkus/blob/3.15.0.CR1/extensions/smallrye-graphql/deployment/src/main/java/io/quarkus/smallrye/graphql/deployment/SmallRyeGraphQLProcessor.java#L286 - here we add any additional classes that need to be indexed.
The Quarkus integration handles the indexing, we don't do any indexing on the SmallRye side (when using Quarkus... for WildFly it's a bit different)

RoMiRoSSaN:
OK. 
All directives have been added for WildFly. 
I will do PR for quarkus. 
Just a question. Is this really how it should be? After all, this is necessary for the correct construction of a graphql schema. And it turns out that this code is duplicated in two different places. Although partially the schema builder should focus on its own api (annotations, directives, etc.), and not wait for information about it from the outside. Or did I understand something wrong?

PR #2196, opened 2024-09-24
Vertx client implementation requires yasson at runtime

jmini:
Version `1.5.0` was quite behind, I have tried to update to the newest version and I noticed the need for additional dependencies.

### DIFF
--- a/docs/client-standalone.md
+++ b/docs/client-standalone.md
@@ -19,11 +19,10 @@ This is a full script runnable directly with [JBang](https://www.jbang.dev/) tha
 uses a dynamic client for connecting to [countries.trevorblades.com](https://countries.trevorblades.com)
 to obtain a list of countries from its database.
 
-```
+```java
 ///usr/bin/env jbang "$0" "$@" ; exit $?
-//DEPS io.smallrye:smallrye-graphql-client-implementation-vertx:1.5.0
+//DEPS io.smallrye:smallrye-graphql-client-implementation-vertx:RELEASE
 
-import io.smallrye.graphql.client.dynamic.api.DynamicGraphQLClientBuilder;
 import io.smallrye.graphql.client.dynamic.api.DynamicGraphQLClient;
 import io.smallrye.graphql.client.Response;
 import io.smallrye.graphql.client.vertx.dynamic.VertxDynamicGraphQLClientBuilder;
@@ -55,4 +54,4 @@ class Client {
 }
 ```
 
-Save this file as `client.java` and execute with `jbang client.java`.
+Save this file as `Client.java` and execute with `jbang Client.java`.
--- a/client/implementation-vertx/pom.xml
+++ b/client/implementation-vertx/pom.xml
@@ -49,6 +49,11 @@
             <groupId>io.vertx</groupId>
             <artifactId>vertx-web-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse</groupId>
+            <artifactId>yasson</artifactId>
+            <scope>runtime</scope>
+        </dependency>
 
         <dependency>
             <groupId>io.smallrye</groupId>
@@ -63,11 +68,6 @@
         <dependency>
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-graphql-client-model-builder</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse</groupId>
-            <artifactId>yasson</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>


jmartisk:
Hi @jmini , big thanks for noticing this and sending a PR.
Actually, I tend to think that if some extra dependencies are required, then it's a bug. It seems to be fixed when you remove the `test` scope of Yasson here: https://github.com/smallrye/smallrye-graphql/blob/2.10.0/client/implementation-vertx/pom.xml#L71
After this, I can run the jbang script without any dependencies beyond `smallrye-graphql-client-implementation-vertx`. Only Yasson is missing, Parsson is not required at all.

So I would rather have this fixed instead of updating the documentation.

So, could you perhaps change your PR to
- remove the line that I linked
- in the docs, for now we will have to stick with version 2.10.0 until we put out a new release, so the Yasson dependency will still need to be mentioned - maybe you could add a sentence that 2.10.0 has a bug that requires you to add Yasson, and right after releasing 2.11.0, I would then remove this note?

jmini:
> I tend to think that if some extra dependencies are required, then it's a bug.

Right now it is the only way to have it working.

With this jbang script:

<details><summary>Client1.java</summary>
<p>

```
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS io.smallrye:smallrye-graphql-client-implementation-vertx:2.10.0

//JAVA 17
import io.smallrye.graphql.client.dynamic.api.DynamicGraphQLClient;
import io.smallrye.graphql.client.Response;
import io.smallrye.graphql.client.vertx.dynamic.VertxDynamicGraphQLClientBuilder;

import io.vertx.core.Vertx;

// Has a multiline string literal, requires Java 15+!
class Client1 {
    public static void main(String... args) throws Exception {
        Vertx vertx = Vertx.vertx();
        DynamicGraphQLClient client = new VertxDynamicGraphQLClientBuilder()
            .url("https://countries.trevorblades.com")
            .vertx(vertx)
            .build();
        try {
            Response response = client.executeSync("""
                query {
                  countries {
                    name
                  }
                }
                """);
            System.out.println(response);
        } finally {
            client.close();
            vertx.close();
        }
    }
}
```

</p>
</details> 

You get:
```
Exception in thread "main" java.lang.NoClassDefFoundError: jakarta/json/JsonValue
	at io.smallrye.graphql.client.vertx.dynamic.VertxDynamicGraphQLClient.buildRequest(VertxDynamicGraphQLClient.java:388)
	at io.smallrye.graphql.client.vertx.dynamic.VertxDynamicGraphQLClient.executeSync(VertxDynamicGraphQLClient.java:165)
	at Client.main(Client1.java:20)
Caused by: java.lang.ClassNotFoundException: jakarta.json.JsonValue
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
	... 3 more
```

You are right `org.eclipse:yasson:2.0.4` is enough (it probably has a dependency on some JSON Processing library).
This works:

<details><summary>Client2.java</summary>
<p>

```
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS io.smallrye:smallrye-graphql-client-implementation-vertx:2.10.0
//DEPS org.eclipse:yasson:2.0.4

//JAVA 17
import io.smallrye.graphql.client.dynamic.api.DynamicGraphQLClient;
import io.smallrye.graphql.client.Response;
import io.smallrye.graphql.client.vertx.dynamic.VertxDynamicGraphQLClientBuilder;

import io.vertx.core.Vertx;

// Has a multiline string literal, requires Java 15+!
class Client2 {
    public static void main(String... args) throws Exception {
        Vertx vertx = Vertx.vertx();
        DynamicGraphQLClient client = new VertxDynamicGraphQLClientBuilder()
            .url("https://countries.trevorblades.com")
            .vertx(vertx)
            .build();
        try {
            Response response = client.executeSync("""
                query {
                  countries {
                    name
                  }
                }
                """);
            System.out.println(response);
        } finally {
            client.close();
            vertx.close();
        }
    }
}
```

</p>
</details>

jmini:
I think the correct scope for `org.eclipse:yasson` in `smallrye-graphql-client-parent` is `runtime`, since you do not want to have it at compile time but you need it at test execution and at project runtime time.

I have updated the PR title accordingly.

docs/client-standalone.md
```diff
@@ -19,11 +19,10 @@ This is a full script runnable directly with [JBang](https://www.jbang.dev/) tha
 uses a dynamic client for connecting to [countries.trevorblades.com](https://countries.trevorblades.com)
 to obtain a list of countries from its database.
 
-```
+```java
 ///usr/bin/env jbang "$0" "$@" ; exit $?
-//DEPS io.smallrye:smallrye-graphql-client-implementation-vertx:1.5.0
+//DEPS io.smallrye:smallrye-graphql-client-implementation-vertx:RELEASE
```
jmartisk, line 24: I don't think that JBang supports this version pointer, I get 
```
[jbang] [ERROR] Could not resolve dependencies: Failed to collect dependencies at io.smallrye:smallrye-graphql-client-implementation-vertx:jar:LATEST
```
when running it

jmartisk:
> I think the correct scope for `org.eclipse:yasson` in `smallrye-graphql-client-parent` is `runtime`, since you do not want to have it at compile time but you need it at test execution and at project runtime time.

You're right, good point (I think the project contains some dependencies that could be switched to `runtime` but we don't care too much)

docs/client-standalone.md
jmini: I can revert to:

```
//DEPS io.smallrye:smallrye-graphql-client-implementation-vertx:2.10.0
```

if you want, but for this version it will not work without:

```
//DEPS org.eclipse:yasson:2.0.4
```

And it would be better to have always the latest version in the documentation. Somehow with `RELEASE` it is ever green.

jmini: There are even integration tests in the jbang project using the `RELEASE` notation:

https://github.com/jbangdev/jbang/blob/7af0ad56db2162121ab48b37132008417528a7c0/itests/SankeyPlotTestWithDeps.java#L16-L18

jmartisk: Ok somehow it works for me now with RELEASE (no idea why it didn't yesterday), but it picks up 2.10.0, so it still needs the dependency... But hey, I think we will release 2.11 (or 2.10.1) within a few weeks, so I guess we could just keep they way you have it now, and hopefully it will start working by itself then

jmartisk (approved):

jmartisk:
Thanks!

docs/client-standalone.md
jmartisk: Now that I think of it, the documentation will also only deploy when we do a release, so with the next documentation update, it should just start working

PR #2201, opened 2024-09-30
Gather bean validation violations in a single field error

dpolysiou:
Change BeanValidationError path to go from root to field, in accordance with the spec.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: wildfly-extras/wildfly-graphql-feature-pack
-          ref: main
+          ref: smallrye-graphql-2.11
           path: wildfly-graphql-feature-pack
 
       - uses: actions/setup-java@v3.0.0
--- a/server/implementation/src/main/java/io/smallrye/graphql/validation/BeanValidationError.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/validation/BeanValidationError.java
@@ -6,6 +6,8 @@ import static java.util.stream.Collectors.toList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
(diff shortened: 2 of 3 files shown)


jmartisk:
@dpolysiou could you please rebase, specifically we need the branch to contain https://github.com/smallrye/smallrye-graphql/commit/a076d7dc19d6dcfaee564461317991ef7330625a (which I just merged a few minutes ago) to have the CI tests work properly

dpolysiou:
I saw one test in the [wildfly-graphql](https://github.com/wildfly-extras/wildfly-graphql-feature-pack) repository that breaks (specifically, [this one](https://github.com/wildfly-extras/wildfly-graphql-feature-pack/blob/main/testsuite/integration/src/test/java/org/wildfly/extras/graphql/test/beanvalidation/BeanValidationTestCase.java#L79)), because it has an assertion for the existing `BeanValidationError` extensions. It's a small change, but I don't know how the changes between these repositories should be coordinated.

jmartisk:
If you're willing to help us fix the test too, then this is how we normally do it:
- I create a special branch in the wildfly-graphql-feature-pack repo named `smallrye-graphql-2.11` or so
- That branch will capture all the changes that we plan to apply along with upgrading to SR-GQL 2.11
- We temporarily change the SR-GQL CI here: https://github.com/smallrye/smallrye-graphql/blob/main/.github/workflows/build.yml#L61 to run against that temporary branch instead of main
- After the upgrade, we switch the CI back to wildfly-graphql-feature-pack's main

So if you want to help with that, I can create the `smallrye-graphql-2.11` branch in `wildfly-graphql-feature-pack` and you can send a PR to it. But no problem if you want to let us handle it ourselves.

jmartisk:
Ah, `GraphQLBeanValidationTest` in Quarkus will have the exact same problem. I already have a Quarkus branch (https://github.com/jmartisk/quarkus/tree/smallrye-graphql-2.11) where this would have to be fixed :)

Do you want me to handle the WildFly and Quarkus tests?

dpolysiou:
Thank you, I will try to make a PR to these branches as soon as possible.

jmartisk:
Thanks for all the PRs!
I've updated the CI to run against the temporary branches... Hopefully it will all pass :)

jmartisk:
It worked, thanks a lot!
I'll take care of updating the wildlfly and quarkus repositories after releasing smallrye-graphql 2.11.

schulzp:
@dpolysiou - This fails now, if the `invalidValue` is null, since `Map.of` does not allow null values

PR #2204, opened 2024-10-03
Fix gradle plugin for correct generation federated schema

RoMiRoSSaN:
The current plugin does not fully fulfill its role. This is a small fix for him. If you have the opportunity to check the correctness of its operation, I would be grateful.
Unfortunately, this is the only option that worked as it should.

### DIFF
--- a/tools/gradle-plugin/plugin/src/main/java/io/smallrye/graphql/gradle/tasks/FederationDotNames.java
+++ b/tools/gradle-plugin/plugin/src/main/java/io/smallrye/graphql/gradle/tasks/FederationDotNames.java
@@ -1,0 +1,44 @@
+package io.smallrye.graphql.gradle.tasks;
+
+import io.smallrye.graphql.api.federation.Authenticated;
+import io.smallrye.graphql.api.federation.ComposeDirective;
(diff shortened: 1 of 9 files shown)


jmartisk:
Thanks @RoMiRoSSaN , from a quick glance it makes sense, I'll give it some more proper testing as soon as I get some time

docs/gradle-plugin.md
```diff
@@ -86,3 +86,7 @@ Configuration options
 
 -   `includeIntrospectionTypes` - Include the introspection types in the
     schema. Default false.
+
+-   `federationEnabled` - Enable GraphQL Federation. This is automatically
+    enabled if any Federation-related annotations are found in your application, otherwise the default is false.
```
jmartisk, line 91: The gradle plugin doesn't automatically check for these, does it? (I don't see that anywhere)

RoMiRoSSaN: I followed the example of maven plugin. Let me try to do automatic detection of federation annotations.

RoMiRoSSaN:
@jmartisk I added autodetect of federation directives. Please, check it

jmartisk (approved):
LGTM, thanks :)

Issue #2081, opened 2024-04-08
The field "data" should not be present in the GraphQL Response if a non-partial error occurs.


mskacelik:
Via https://spec.graphql.org/draft/#sec-Response-Format quote:
> If the request included execution, the response map must contain an entry with key data. The value of this entry is described in the “Data” section. **If the request failed before execution due to a syntax error, missing information, or validation error, this entry must not be present.**

Yet SRGQL implementation returns the `data` field:
![image](https://github.com/smallrye/smallrye-graphql/assets/124160830/e7896863-5f21-48db-803a-ee219f28617c)

To fix the issue, there need to be changes both on the server side and also on the client side (typesafe).

mskacelik:
I will look into it this week.

PR #2225, opened 2024-11-11
removed unnecessary "data": null if request failed before execution

mskacelik:
https://spec.graphql.org/draft/#sec-Response-Format
> If the request included execution, the response map must contain an entry with key data. The value of this entry is described in the “Data” section. If the request failed before execution due to a syntax error, missing information, or validation error, this entry must not be present.

The only exception is our bean validation, responses during an error contain this field:
```
"data" : {
    operationName: null
}
```

There was no need to change anything on the client side.

fix:#2081

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,8 +96,8 @@ jobs:
       - name: checkout Quarkus repository
         uses: actions/checkout@v2
         with:
-          repository: quarkusio/quarkus
-          ref: main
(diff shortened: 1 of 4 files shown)


jmartisk (approved):

mskacelik:
Quarkus branch (2.11.1): https://github.com/mskacelik/quarkus/tree/srgql-2.11.1

PR #1731, opened 2023-02-07
Make Optional* work as an input, add tests

mskacelik:
cc @jmartisk

PR #2229, opened 2024-11-21
Make it possible to use Optional on payloads

velo:
This PRs supplements https://github.com/smallrye/smallrye-graphql/pull/1731

Make it possible to have Optional Strings too


t1 (approved):
LGTM. The tests could also test more complex scenarios, e.g. something like `Optional<List<Optional<String>>>`; but I guess it's good enough.

velo:
Thanks a lot.

Any idea when next release will come out?

jmartisk:
AFAIK we can't release right now because @radcortez is migrating to a new release process - is that right? 
Hopefully we will be able to get SmallRye GraphQL 2.12 out for Quarkus 3.18

PR #2236, opened 2024-12-05
Update Release workflow

radcortez:
To use the new release process... I'm unsure if this will work on the first try (probably not :) and will require some adjustments).

This project also releases a Gradle plugin. I did that on the preparation workflow before executing the central one. The files should be updated and tagged:
https://github.com/smallrye/smallrye-graphql/pull/2236/files#diff-ddf2bab74923dcd43f9f8d05e50bb1adc9ee5366a394d151d1f04a54a6079486

There is a separate workflow that executes to do the actual Gradle release after the central one and tag:
https://github.com/smallrye/smallrye-graphql/pull/2236/files#diff-8d232d862a776c1047b335f66544717bd08e86abac7de606a02ffa018563476f

This also seems to require `graphviz` that most likely we don't have available in the central repo:
https://github.com/smallrye/smallrye-graphql/pull/2236/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L38-L39

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,47 @@
+name: SmallRye Prepare Release
+
+on:
+  pull_request:
+    types: [ closed ]
+    paths:
+      - '.github/project.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+
+jobs:
+  prepare-gradle:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: git author
+        run: |
+          git config --global user.name "SmallRye CI"
+          git config --global user.email "smallrye@googlegroups.com"
+
+      - uses: radcortez/project-metadata-action@main
+        name: retrieve project metadata
+        id: metadata
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          metadata-file-path: '.github/project.yml'
+
+      - name: Prepare Gradle
+        run: |
+          echo "version=${{steps.metadata.outputs.current-version}}" > tools/gradle-plugin/gradle.properties
+          git add tools/gradle-plugin/gradle.properties
+          git commit -m "Update Gradle plugin version"
+          git push
+
+  prepare-release:
+    needs: [prepare-gradle]
+    name: Prepare Release
+    if: ${{ github.event.pull_request.merged == true}}
+    uses: smallrye/.github/.github/workflows/prepare-release.yml@main
+    secrets: inherit
--- a/.github/workflows/publish-gradle.yml
+++ b/.github/workflows/publish-gradle.yml
@@ -1,0 +1,30 @@
+name: Publish Gradle
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        description: Tag version to perform release
+        type: string
+
+jobs:
+  publish-gradle:
+    name: Publish Gradle
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        name: checkout ${{inputs.version}}
+        with:
+          ref: ${{inputs.version}}
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 11
+
+      - name: gradle release ${{inputs.version}}
+        run: |
+          mkdir -p ~/.gradle ; echo -e "gradle.publish.key=${{secrets.GRADLE_PUBLISH_KEY}}\ngradle.publish.secret=${{secrets.GRADLE_PUBLISH_SECRET}}" > ~/.gradle/gradle.properties
+          cd tools/gradle-plugin && gradle publishPlugins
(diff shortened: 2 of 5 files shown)


.github/workflows/publish-gradle.yml
```diff
@@ -0,0 +1,30 @@
...
+          ref: ${{inputs.version}}
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 11
+
+      - name: gradle release ${{inputs.version}}
+        run: |
+          mkdir -p ~/.gradle ; echo -e "gradle.publish.key=${{secrets.GRADLE_PUBLISH_KEY}}\ngradle.publish.secret=${{secrets.GRADLE_PUBLISH_SECRET}}" > ~/.gradle/gradle.properties
```
gastaldi, line 29: Can't you set them as environment variables?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,45 @@
...
+        name: retrieve project metadata
+        id: metadata
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          metadata-file-path: '.github/project.yml'
+
+      - name: Prepare Gradle
+          echo "version=${{steps.metadata.outputs.current-version}}" > tools/gradle-plugin/gradle.properties
+          git add tools/gradle-plugin/gradle.properties
+          git commit -m "Update Gradle plugin version"
```
gastaldi, line 38: You need to push the changes somewhere, the `prepare-release` job below won't execute in the same context as this job

gastaldi: I'd recommend having a hook to perform the changes you need in the prepare-release.yml, like in https://github.com/quarkiverse/.github/blob/main/.github/workflows/prepare-release.yml#L118-L124

radcortez: Ah, we need to do a `git push` there.

Then, the job below would just check out the new commit.

radcortez:
> This also seems to require `graphviz` that most likely we don't have available in the central repo:
> https://github.com/smallrye/smallrye-graphql/pull/2236/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L38-L39

@gastaldi what about this?

.github/workflows/prepare-release.yml
gastaldi: For `git push` to work, it depends on the ruleset configured in this repository too (I don't know if the current user is able to bypass the branch restrictions). In any case, I think it would be better if it used the GitHub App token, as in 
https://github.com/smallrye/.github/blob/main/.github/workflows/prepare-release.yml#L35-L40

gastaldi:
@radcortez is `graphviz` required in `gradle publishPlugins` or when running `release:prepare`?  If the latter, it would be better to change the prepare-release.yml in the `smallrye/.github` repository to support hooks, as I mentioned in https://github.com/smallrye/smallrye-graphql/pull/2236/files#r1871984725

gastaldi:
@radcortez see https://github.com/smallrye/.github/pull/5

radcortez:
> @radcortez is `graphviz` required in `gradle publishPlugins` or when running `release:prepare`? If the latter, it would be better to change the prepare-release.yml in the `smallrye/.github` repository to support hooks, as I mentioned in https://github.com/smallrye/smallrye-graphql/pull/2236/files#r1871984725

I think it is required for javadocs:
https://github.com/smallrye/smallrye-graphql/blob/95641c9a49705ca7680cfe561f8f98aa71cdf6fd/pom.xml#L477-L489

jmartisk:
I am not really sure what that is used for, and I think we may not need it at all - but maybe @phillip-kruger  can correct me

phillip-kruger:
Yea it's used in the Javadoc. See https://javadoc.io/doc/io.smallrye/smallrye-graphql/latest/index.html or https://javadoc.io/doc/io.smallrye/smallrye-graphql/latest/io/smallrye/graphql/execution/package-summary.html as an example

jmartisk:
Hm, looking at these diagrams, they look a bit crazy, and TBH I don't know if useful. Maybe if it complicates the release too much, we could remove it

phillip-kruger:
Your call.

radcortez:
It shouldn't complicate the release, but we do have to add https://github.com/smallrye/smallrye-graphql/pull/2236/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L38-L39 to the action running the release. 

I don't see a big issue in it. What do you think @gastaldi?

gastaldi:
@radcortez when https://github.com/smallrye/.github/pull/5 is merged, you can add a `.github/workflows/hooks/pre-prepare-release.sh` file to this repository with the commands to install it

radcortez:
> Hm, looking at these diagrams, they look a bit crazy, and TBH I don't know if useful. Maybe if it complicates the release too much, we could remove it

@gastaldi has a way to install the necessary components to support it. Is this something you want to keep so we can prepare the scripts for it?

jmartisk:
If it's not a big problem, then let's keep it, but I wouldn't mind dropping it either.

radcortez:
I've tried locally, and it seems to build fine without it. Let me try it without.

Issue #2232, opened 2024-11-29
Record with List<T> (de-)serializes differently in Java 21


mskacelik:
The test `testNestedRecordWithListWithMissingFieldInQuery` fails in Java 21 due to the weird behavior of the `JsonB` implementation – record containing a `List<T>` instance. The test works fine on Java 17.

https://github.com/smallrye/smallrye-graphql/blob/c9fc8b9b8bed47d57216dc6ad13e4b874de5731f/server/integration-tests/src/test/java/io/smallrye/graphql/tests/client/dynamic/NestedRecordsTest.java#L72-L74

Issue reported: https://github.com/eclipse-ee4j/yasson/issues/656

mskacelik:
The problem is due to https://bugs.openjdk.org/browse/JDK-8320575

PR #2240, opened 2024-12-11
Ignoring testNestedRecordWithListWithMissingFieldInQuery due to JDK (21) bug

mskacelik:
#2232
https://bugs.openjdk.org/browse/JDK-8320575

### DIFF
--- a/server/integration-tests/src/test/java/io/smallrye/graphql/tests/client/dynamic/NestedRecordsTest.java
+++ b/server/integration-tests/src/test/java/io/smallrye/graphql/tests/client/dynamic/NestedRecordsTest.java
@@ -9,6 +9,7 @@ import static io.smallrye.graphql.client.core.InputObjectField.prop;
 import static io.smallrye.graphql.client.core.Operation.operation;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assume.assumeFalse;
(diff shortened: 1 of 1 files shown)


server/integration-tests/src/test/java/io/smallrye/graphql/tests/client/dynamic/NestedRecordsTest.java
```diff
@@ -69,9 +68,13 @@ public void testNestedRecordWithMissingFieldInQuery() throws Exception {
         }
     }
 
-    @Ignore("In Java 21 this test fails due to a bug (presumably) in the JSON-B implementation – Yasson)")
     @Test
     public void testNestedRecordWithListWithMissingFieldInQuery() throws Exception {
+        if (System.getProperty("java.version").startsWith("21.")) {
```
jmartisk, line 73: This affects 22 too, could we make it safer and do something along the lines of
```
String major = Integer.valueOf(System.getProperty("java.version").split("\\.")[0]);
if(major >= 21 && major <= 22) {
```
? Also, when skipping a test, print an info message into the log

mskacelik: yea could do

jmartisk (approved):

Issue #2253, opened 2024-12-21
`JsonbProperty#nillable` attribute is replaced by `@JsonbNillable`


t1:
I just saw that the `nillable = true` attribute is replaced by `@JsonbNillable`

See: https://jakarta.ee/specifications/jsonb/3.0/apidocs/jakarta.json.bind/jakarta/json/bind/annotation/jsonbproperty#nillable()

So I guess a follow up issue/PR to fix `FieldInfo` is needed.

_Originally posted by @jmini in https://github.com/smallrye/smallrye-graphql/issues/2251#issuecomment-2556840858_

t1:
@jmartisk: I currently experience maven build problems, because some versions seem to be missing. Are you aware or am I doing something wrong?

t1:
Oh, and by the way: the example code in `custom-json-serializers-deserializers` should be fixed, too.

jmartisk:
> @jmartisk: I currently experience maven build problems, because some versions seem to be missing. Are you aware or am I doing something wrong?

What exactly is missing? I know there was an issue with the Gradle plugin, I had to deploy it manually afterwards

jmini:
@t1 I have opened this PR https://github.com/smallrye/smallrye-graphql/pull/2254 for support for `@JsonbNillable`, but I am not really sure this is enough and I wanted to contribute a test case (and I do not see how to add it)

jmini:
> Oh, and by the way: the example code in custom-json-serializers-deserializers should be fixed, too.

This is why docs written with asciidoctor with code snippets included from a test class that is compiled and executed at built time helps to keep the doc up-to-date.

I did not worked on that part

jmartisk:
> > Oh, and by the way: the example code in custom-json-serializers-deserializers should be fixed, too.
> 
> This is why docs written with asciidoctor with code snippets included from a test class that is compiled and executed at built time helps to keep the doc up-to-date.

It's a double-edged sword because it also complicates the build quite a bit by adding new dependencies between modules, and it could lead to unexpected surprises when you change some code and don't realize that it could break the docs. I'm personally not a big fan of this approach.

Issue #2251, opened 2024-12-20
Typed client: force sending null value


jmini:
It seem that the gitlab graphql server has a very specific way of treating null.

I looked at the query created by the js graphQL client when an action (removing the due-date of a workItem) is created in the UI. It's send:

```json
{
    "input": {
        "id": "gid://gitlab/WorkItem/18672",
        "rolledupDatesWidget": {
            "dueDateIsFixed": true,
            "startDateIsFixed": true,
            "dueDateFixed": null,
            "startDateFixed": null
        }
    }
}
```

I was able to reproduce it in graphiQL https://gitlab.com/-/graphql-explorer

<img width="657" alt="Screenshot 2024-12-20 at 09 39 14" src="https://github.com/user-attachments/assets/a8423710-891e-4a52-9619-a95b6a248146" />

Corresponding java code:

```java
WorkItemUpdateInput request = new WorkItemUpdateInput()
        .setId(workItem.getId())
        .setRolledupDatesWidget(new WorkItemWidgetRolledupDatesInput()
                .setDueDateFixed(null)
                .setDueDateIsFixed(true)
                .setStartDateFixed(null)
                .setStartDateIsFixed(true) //
        ) //
;
WorkItemUpdatePayload updateResponse = api.workItemUpdate(request);
```

Library is here: https://github.com/unblu/gitlab-workitem-graphql-client

But this is turned by the typed-client to this request:

```
{"arg0":{"id":"gid://gitlab/WorkItem/18672","rolledupDatesWidget":{"startDateIsFixed":true,"dueDateIsFixed":true}}}
```

The `null` values are not sent, and it seem that this is interpreted differently by the server.

If change graphiQL to:

```json
{
    "input": {
        "id": "gid://gitlab/WorkItem/18672",
        "rolledupDatesWidget": {
            "dueDateIsFixed": true,
            "startDateIsFixed": true
        }
    }
}
```

As with the java typed client, no change is performed by the server.

---

Is it a way to force the GraphQL client to set the `null` values?

jmini:
I somehow need to set something like:

```java
JsonbConfig config = new JsonbConfig()
  .withNullValues(true)
```

jmartisk:
Yeah could you try to register a custom `JsonbConfig` like we do for example in this test:

- https://github.com/smallrye/smallrye-graphql/blob/main/server/integration-tests/src/test/java/io/smallrye/graphql/tests/json/CustomJsonbService.java#L21
[...]

PR #2254, opened 2024-12-30
Add support for @JsonbNillable

jmini:
If you want the type-safe client to send an explicit null value to the server,  you can do this:

```java
    /**
     * Fixed due date for the work item.
     */
    @JsonbProperty(value = "dueDateFixed", nillable = true)
    private Date dueDateFixed;
```

This works fine as discussed in https://github.com/smallrye/smallrye-graphql/issues/2251 but the `nillable` attribute is deprecated in [`@JsonbProperty`](https://jakarta.ee/specifications/jsonb/3.0/apidocs/jakarta.json.bind/jakarta/json/bind/annotation/jsonbproperty#nillable())

This PR adds support for @JsonbNillable that can be used like this:

```java
    /**
     * Fixed due date for the work item.
     */
    @JsonbNillable
    private Date dueDateFixed;
```

Fixes https://github.com/smallrye/smallrye-graphql/issues/2253

### DIFF
--- a/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/FieldInfo.java
+++ b/client/implementation/src/main/java/io/smallrye/graphql/client/impl/typesafe/reflection/FieldInfo.java
@@ -8,6 +8,7 @@ import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Optional;
 
+import jakarta.json.bind.annotation.JsonbNillable;
 import jakarta.json.bind.annotation.JsonbProperty;
 
 import org.eclipse.microprofile.graphql.Name;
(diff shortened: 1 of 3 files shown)


jmini:
I do not see any tests for usage of `@JsonbProperty` on a model class, so I am not sure how I could add a regression-test case for this change in the library.

jmartisk:
Thanks for this!

> I do not see any tests for usage of `@JsonbProperty` on a model class, so I am not sure how I could add a regression-test case for this change in the library.

Yeah we don't have  a test for it yet, but it would be cool to add it. I think the additional TCK would be a good place for it, plenty of tests under `client/tck/src/main/java/tck/graphql/typesafe` could serve as inspiration

jmini:
I have updated the PR to add a test

jmartisk (approved):
Nice, thanks!

PR #2260, opened 2025-01-24
Added better exception messages for dynamic client name validation

mskacelik:


client/api/src/main/java/io/smallrye/graphql/client/core/utils/validation/NameValidation.java
```diff
@@ -83,13 +89,38 @@ public static String validateName(String name) {
...
+                    "Invalid field name: '%s'. Field name does not match the regex pattern %s, please check the GraphQL specification %s"
+                            .formatted(fieldName, _FIELD_NAME_REGEX, "https://spec.graphql.org/draft/#sec-Language.Fields"));
         }
         return fieldName;
     }
 
+    public static String validateTypeName(String typeName) {
+        if (typeName == null) {
+            throw new IllegalArgumentException(
+                    "Invalid type name: '%s'. Type name does not match the regex pattern %s, please check the GraphQL specification %s"
```
jmartisk, line 102: This would print `Invalid type name: 'null'`, right? That's a bit confusing, because a (literal) `null` would actually be a valid type name :) (at least I think so, https://spec.graphql.org/draft/#sec-Names does not forbit it) Rather say `Type name cannot be null` or something

mskacelik: good catch

jmartisk (approved):

Issue #2142, opened 2024-07-03
Problems with annotations like nonnull and lists in kotlin


robp94:
Follow-up issue of https://github.com/smallrye/smallrye-graphql/issues/1366

There seems to be a discrepancy between client and server

    @Query
    fun test2(list: List<@NonNull String>): String

    @Query
    fun test3(list: @NonNull List<@NonNull String>): String

result in schema

  test2(list: [String!]!): String
  test3(list: [String!]!): String

But both test fail with

Variable type '[String!]' doesn't match expected type '[String!]!

I think that both server and client should fit together.
Furthermore, test1 is wrong in both client and server, but could be some kotlin problem.

    @Query
    fun test1(list: @NonNull List<String>): String

Graphql:

test1(list: [String]): String

Note kotlin is a bit special with annotations https://youtrack.jetbrains.com/issue/KT-35843 we need an extra flag.
The reproducer uses a quarkus snapshot based on this pr https://github.com/smallrye/smallrye-graphql/pull/2141
[code-with-quarkus-nonnull-kotlin.zip](https://github.com/user-attachments/files/16085550/code-with-quarkus-nonnull-kotlin.zip)

mskacelik:
A bit of an observation from the reproducer:
### build-time typesafe client – newer implementation for Quarkus

query testNull($list: [String]) { testNull(list: $list) }
query test1($list: [String]!) { test1(list: $list) }
query test2($list: [String!]) { test2(list: $list) }
query test3($list: [String!]!) { test3(list: $list) }
### reflection (client) – outdated for Quarkus
query testNull($list: [String]) { testNull(list: $list) }
query test1($list: [String]) { test1(list: $list) }
query test2($list: [String!]) { test2(list: $list) }
query test3($list: [String!]) { test3(list: $list) }
### schema (server)

type Query {
test1(list: [String]!): String!
test2(list: [String!]!): String!
test3(list: [String!]!): String!
testNull(list: [String]!): String!
}

It is a bit tricky with Kotlin because there are the `@NonNull` annotations and also the implicit non-null types (unless explicitly set via `?`).

The server side checks the nullable `?` types (via kotlin metadata), but the client does not. However, from the look, not even the server-side implementation has the correct behavior.

So, for the sake of it, in my opinion, the correct behavior should look something like this (for both the server and client):
```
testNull(list: [String!]!) : String!
test1(list: [String!]!) : String!
test2(list: [String!]!) : String!
[...]

PR #2263, opened 2025-01-29
Kotlin nullability feature for typesafe client (client model)

mskacelik:
/cc @jmartisk
Ok, so with nullabilities in Kotlin, it is a bit complicated; when not having an `?` nullable operator for types, Kotlin automatically adds its own Kotlin `@NonNull` annotations, but not *ALWAYS*. For example, when having a `List<String>`, the annotation is only applied on the `List<>` part but not the argument String. So, making these two codes indistinguishable (via Jandex):
```kotlin
List<String?> // List<String>
```
On top of that, there are some weird behaviors like annotation is applied only on `MethodParameterInfo` and not the `Type` 

So, I only added a simple Kotlin nullability feature (see tests). So it can be consistent with the server side.
Java reflection implementation of typesafe client still does not have Kotlin nullability checking (todo?).

issue: #2142
@robp94

### DIFF
--- a/client/model-builder/pom.xml
+++ b/client/model-builder/pom.xml
@@ -50,6 +50,12 @@
             <artifactId>smallrye-graphql-api</artifactId>
             <scope>test</scope> <!--subscription annotation-->
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-reflect</artifactId>
+            <version>${version.kotlin}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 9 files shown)


client/model-builder/pom.xml
```diff
@@ -61,6 +67,45 @@
...
             </plugin>
+            <plugin>
+                <artifactId>kotlin-maven-plugin</artifactId>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <version>${version.kotlin.compiler}</version>
+                <executions>
+                    <execution>
+                        <id>compile</id>
+                        <goals>
+                            <goal>compile</goal>
```
jmartisk, line 78: Don't we only need `test-compile` because you're not using Kotlin in the application?

client/model-builder/pom.xml
```diff
@@ -61,6 +67,45 @@
...
+                <configuration>
+                    <compilerPlugins>
+                        <plugin>all-open</plugin>
+                    </compilerPlugins>
+                    <args>
+                        <arg>-java-parameters</arg>
+                    </args>
+                    <pluginOptions>
+                        <!-- Each annotation is placed on its own line -->
+                        <option>all-open:annotation=jakarta.ws.rs.Path</option>
```
jmartisk, line 97: We're not using `@Path` anywhere so I guess we don't need this

jmartisk:
@robp94 would you be willing to give this a try and see if it suits your expectations?

robp94:
Looks good to me.

mskacelik:
I played a bit with the MP `@NonNull` annotations now (with `-Xemit-jvm-type-annotations` flag) and it seems to work with Kotlin now if we explicitly annotate inside of wrappers.
```kotlin
@Mutation
    fun testMatrix(working: List<@NonNull String>): Boolean {
        println("hello world")
        return true
    }
```
schema:
```
"Mutation root"
type Mutation {
  testMatrix(working: [String!]!): Boolean!
}
```

mskacelik:
Ok, it seems to have some problems regarding nested Lists, i.e.: `List<List<Int>>` only generates `[Int]` (only in Kotlin, Java works (server side)).

But I think I found a bigger problem, in Quarkus `@NonNull` annotations seem to be very inconsistent after HOT RELOAD.
For example:
```>quarkus dev
List<@NonNull String> => generates [String!]!
changes code to Array<@NonNull String> => generates [String]!
...closes app...
> quarkus dev
Array<@NonNull String> => generates [String!]!
changes code to List<@NonNull String> => generates [String]!
```
Very strange

jmartisk (approved):

PR #2277, opened 2025-03-24
Add support for the client supporting serializing enum values

rubik-cube-man:
Previously, if you tried to use an enum as a variable, it would error, as it would try to be serialized to JSON. As there was no applicable JSON serializer associated with this enum it would fail.

This changes makes sure that enums are serialized as a string.


jmartisk (approved):
Nice, thanks!

PR #2283, opened 2025-04-25
Ensure that large messages sent over websockets are reassembled on the client

rubik-cube-man:
Currently, there is an issue when a message is sent over a websocket which is over the max frame size, that it is not correctly reassembled first. The current behaviour processes each frame of a message as a response, as opposed to joining the frames together first.

This change switches the handler() method to textMessageHandler() instead. This new method only passes the assembled message through.


jmartisk (approved):

jmartisk:
Thanks!

PR #2294, opened 2025-08-22
Make sure that exceptions are propagated when a WebSocket fails to open

rubik-cube-man:
If the WebSocket is closed before the CONNECTION_ACK is received, then the emitter is never notified, and anything trying to execute a query will hang indefinitely (unless a timeout is specified, but even then, you only get the timeout exception, instead of the actual reason your socket closed).


jmartisk (approved):
Thanks!

PR #2295, opened 2025-08-25
Fix nested collections parameters in Type-safe client

### LINKED ISSUE #2265: Type-safe client (client model implementation) wrong behavior with nested collections parameters
When working with nested collections, i.e.:
```java
    @Query
    List<List<String>> returnStringListList(List<List<String>> someStrings);
```
Expected: `query returnStringListList($someStrings: [[String]]) { returnStringListList(someStrings: $someStrings) }`
Actual: `query returnStringListList($someStrings: [ListInput]) { returnStringListList(someStrings: $someStrings) }`

The cause of this problem is probably here (no recursion):
https://github.com/smallrye/smallrye-graphql/blob/7a7a3aba6ef544bb9d5f3761753694f05c372401/client/model-builder/src/main/java/io/smallrye/graphql/client/model/helper/ParameterModel.java#L161-L177

mskacelik:
Fixes #2265 using a recursion for collections (both for Jandex and Java Reflection implementation).
Plus a bit of code refactoring.


jmartisk:
@mskacelik please rebase on top of the latest CI fixes, thanks :)

PR #2298, opened 2025-08-31
Deferred completionStage when executing HTTP Async calls in VertxTypesafeGraphQLClientProxy

### LINKED ISSUE #2293: Retries don’t re-run HTTP call in VertxTypesafeGraphQLClientProxy (eager CompletionStage)
**What’s going on**
When you call a typesafe client method and add `.onFailure().retry()` on the returned `Uni`, the HTTP request is not sent again on retry. You just get the same failure replayed.

This happens in the HTTP path of
`client/implementation-vertx/.../VertxTypesafeGraphQLClientProxy.java`, method
`executeSingleResultOperationOverHttpAsync(...)`.

If there are **no dynamic headers**, the code builds the `Uni` like this:

```java
return Uni.createFrom()
  .completionStage(postAsync(request.toString(), allHeaders))
  .map(response -> new ResultBuilder(...).read());
```

`postAsync(...)` creates a `CompletionStage` eagerly, so by the time the `Uni` exists, the HTTP call has already been started. On retry, Mutiny re-subscribes to the **same, already-completed stage**, so the request is not re-executed.

If there **are dynamic headers**, the code uses `transformToUni(...)`, which runs the lambda per subscription. In that branch a new `CompletionStage` is created on each retry, so the HTTP call is sent again. That means behavior diffesr depending on whether dynamic headers are present.

**Why this matters**
Retries are expected to trigger a fresh attempt. Right now they don’t (unless you happen to use dynamic headers), which breaks backoff/retry policies and makes transient errors permanent.

**How to reproduce (minimal)**

```java
Uni<String> u = client.someQuery(arg) // typesafe client
  .onFailure().retry().atMost(3);

u.await().indefinitely();
```

Point the client at a testt endpoint that fails the first two times and suceeds the third. With the current code you’ll see only one HTTP attempt.

**What I expected**
Each retry creates a new HTTP request.

**What actually happens**
Only the first attempt hits the network; retries replay the same failure.

**Proposed fix**
Defer `CompletionStage` creation by using the supplier overload so a new stage (and HTTP request) is created per subscription.

lacoonte:
Fixes #2293 and #2300

### DIFF
--- a/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
+++ b/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/typesafe/VertxTypesafeGraphQLClientProxy.java
@@ -2,7 +2,6 @@ package io.smallrye.graphql.client.vertx.typesafe;
 
 import static io.smallrye.graphql.client.impl.JsonProviderHolder.JSON_PROVIDER;
 import static java.util.stream.Collectors.*;
-import static java.util.stream.Collectors.toList;
 
 import java.lang.reflect.Array;
 import java.math.BigDecimal;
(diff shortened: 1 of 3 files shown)


jmartisk:
The code looks good IMO (except you need to run the formatter plugin, the CI is complaining).

About testing... Do you have an idea how to easily test this? I think we don't have suitable infrastructure in place for stopping/starting servers, but AFAIU the retry also applies to when the server returns an error, so we could test it with a server that returns an error on the first invocation and succeeds on the second? Maybe you have a better idea...?

jmartisk:
@lacoonte, If you're looking for tests that use a similar architecture and can be used for inspiration, look for example at https://github.com/smallrye/smallrye-graphql/blob/main/server/integration-tests/src/test/java/io/smallrye/graphql/tests/client/typesafe/RecordAsInputToTypesafeClientTest.java

Thanks, much appreciated! :)

lacoonte:
@jmartisk Took me more time than i thought as i had a busy week at work. Added two test cases to cover both issues. They fail on the main branch but pass on this one, let me know if there’s anything else I can improve :)

jmartisk (approved):
Thanks a lot!!!

PR #2301, opened 2025-09-19
Avoid unnecessary conversion of JsonValue objects

dpolysiou:
When returning `JSON` scalar data from a field, we can return the `JsonValue` object as is, without recursively traversing the pojo.

This makes a difference when the field payload is large, as it avoids performing unnecessary work on the event loop thread. [Here](https://github.com/dpolysiou/smallrye-graphql-event-loop-blocking-reproducer) is an example where an unreasonably large field leads to blocking the event loop thread.

### DIFF
--- a/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionResponse.java
+++ b/server/implementation/src/main/java/io/smallrye/graphql/execution/ExecutionResponse.java
@@ -160,7 +160,12 @@ public class ExecutionResponse {
         if (pojo == null) {
             return JsonValue.NULL;
         } else if (pojo instanceof JsonValue) {
-            return (JsonValue) pojo;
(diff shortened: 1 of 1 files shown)


jmartisk:
If you look a few lines below, we have some more customized logic for processing the object, like acting on the `isExcludeNullFieldsInResponses` configuration field, will this not break that functionality? I guess you would skip the removal of nulls from the response if `isExcludeNullFieldsInResponses==true` with this code.

Perhaps a good solution would be to leave it to the original custom logic if `isExcludeNullFieldsInResponses==true` (or even better, implement a faster version of it specifically for `JsonValue` that would only remove nulls) and do your thing of returning the JsonValue directly only if `isExcludeNullFieldsInResponses==false`?

jmartisk:
Thanks!

PR #2444, opened 2026-04-22
Migration to Vert.x 5.0.11

jponge:
This migrates the code to Vert.x 5.0.11.

### DIFF
--- a/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/dynamic/VertxDynamicGraphQLClient.java
+++ b/client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/dynamic/VertxDynamicGraphQLClient.java
@@ -8,6 +8,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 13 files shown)


jponge:
@jmartisk I have opened a PR against `main`, but let me know if this should go against a different branch.

jponge:
Part of the https://github.com/quarkusio/quarkus/issues/51959 effort

client/implementation-vertx/src/main/java/io/smallrye/graphql/client/vertx/dynamic/VertxDynamicGraphQLClient.java
```diff
@@ -394,7 +399,7 @@ private Request buildRequest(String query, Map<String, Object> variables, String
     @Override
     public void close() {
         try {
-            httpClient.close();
+            Future.join(httpClient.close(), webSocketClient.close()).await();
```
jmartisk: Should this have a timeout or it can be safely assumed that this won't get stuck forever?

jponge: There's a variant with a timeout, what value should I use?

jmartisk: Yeah I guess it doesn't make a lot of sense to create a new config property just for that... Maybe just pick a number, like 30 seconds maybe?

jponge: I just took a timeout that was already floating around 😉

jponge: Hum let me put 30s instead of reusing a timeout value

jmartisk (approved):

PR #2515, opened 2026-05-28
Build fixes for the 3.y.z branch

mskacelik:
Wanted to work on one issue, but there seem to be few issues regarding the build, so this PR attempts to fix them.
5f08e6ae17822758a41f0c5294d6705f099238dd and bdeaa693c294f85dc9716d52a8db36c517d14c68 might be worth backporting to 2.x.

Regarding the b6611d43df2c8dad9e6f8af5a726892109ff5caf, it contains the fix for the https://github.com/smallrye/smallrye-stork/issues/1201
Which was not backported 2.x, so either, a) we will ask for backport, or b) we would just use SR Stork 3.x instead, which I think is fine since it aligns with the Vertx 5 migration of SR GQL 3.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
         <version.smallrye-mutiny>3.2.0</version.smallrye-mutiny>
         <version.smallrye-mutiny-zero>1.2.0</version.smallrye-mutiny-zero>
         <version.smallrye-context-propagation>2.3.0</version.smallrye-context-propagation>
-        <version.smallrye-stork>2.7.9</version.smallrye-stork>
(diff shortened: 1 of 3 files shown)


jmartisk:
Thanks a lot!!
Yeah I guess I'll backport https://github.com/smallrye/smallrye-graphql/commit/bdeaa693c294f85dc9716d52a8db36c517d14c68 and https://github.com/smallrye/smallrye-graphql/commit/5f08e6ae17822758a41f0c5294d6705f099238dd

For the stork issue, https://github.com/smallrye/smallrye-stork/issues/1201 now says it's fixed in 2.7.10, which we can use in 2.x

jmartisk (approved):

PR #2533, opened 2026-06-05
Fix type variable resolution for multi-level parameterized hierarchies.

### LINKED ISSUE #2267: Errors with extended generics
I'm trying to write some filter input classes, but it throws errors when generic types include "extends":

```java
public class Filter<FieldType> {
  public FieldType value;
  public FieldType[] values;
  public Operator op=Operator.EQUAL;

  public Predicate toPredicate(...) {
    ... create a predicate ...
  }
}

public class StringFilter extends Filter<String> { // Works
  @Override
  public Predicate toPredicate(...) {
    ... custom predicate options for strings specifically ...
    ... or super.toPredicate(...) ...
  }
}


public class NumberFilter<NumberType extends Number> extends Filter<NumberType> {
  @Override
  public Predicate toPredicate(...) {
    ... custom predicate options for number types specifically ...
    ... or super.toPredicate(...) ...
  }
}

public class IntFilter extends Filter<Integer> {} // works
public class IntFilter extends NumberFilter<Integer> {} // but this one errors:
// io.smallrye.graphql.schema.SchemaBuilderException: 
//   Don't know what to do with [FieldType] of kind [TYPE_VARIABLE] as parent object reference doesn't contain necessary info: 
// Reference{
//   className=<pkg>.v1.query.filters.IntFilter,
//   name=IntFilterInput,
//   type=INPUT,
//   graphQLClassName=null,
//   adaptTo=null,
//   adaptWith=null, 
//   directiveInstances=null,
//   classParametrizedTypesnull,
//   extendedClassParametrizedTypes{
//     NumberType=Reference{
//       className=java.lang.Integer, 
//       name=Int,
//       type=SCALAR,
//       graphQLClassName=java.lang.Integer,
//       adaptTo=null,
//       adaptWith=null, 
//       directiveInstances=null,
//       classParametrizedTypesnull,
//       extendedClassParametrizedTypesnull,
//       wrapper=null
//     }
//   }, wrapper=null
// }
```

Is this a bug, or is there a better way to handle this scenario, without boilerplate code for each number scalar?

mskacelik:
- Fixes #2267

Went by what I described in this [comment](https://github.com/smallrye/smallrye-graphql/issues/2267#issuecomment-3237120332) but upgraded a bit - instead of just using class-qualified identifiers as plain string keys in the same `Map<String, Reference>`, I introduced a `ParametrizedTypeEntry` class that carries `declaringClassName`, identifier, and reference together. 

This also let me merge the two separate map (`classParametrizedTypes` and `parentClassParametrizedTypes`) into a single `Map<String, ParametrizedTypeEntry>` with composite keys (`declaringClassName:identifier`), which removes the need for the dual-map split and the "FIXME merge" logic in `getAllParametrizedTypes()` entirely. 
The `addParametrizedTypeNameExtension` flag that was already there handles the own/inherited distinction for name generation, so no information was lost.

### DIFF
--- a/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/FieldCreator.java
+++ b/common/schema-builder/src/main/java/io/smallrye/graphql/schema/creator/FieldCreator.java
@@ -100,7 +100,8 @@ public class FieldCreator extends ModelCreator {
             Type fieldType = getFieldType(fieldInfo, methodType);
 
             Reference reference = referenceCreator.createReferenceForPojoField(fieldType, methodType,
-                    annotationsForPojo, direction, parentObjectReference);
+                    annotationsForPojo, direction, parentObjectReference,
+                    methodInfo.declaringClass().name().toString());
 
             Field field = new Field(methodInfo.name(), MethodHelper.getPropertyName(direction, methodInfo.name()), name,
                     reference);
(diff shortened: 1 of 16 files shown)


jmartisk (approved):
Super awesome, thanks so much for this!

PR #2544, opened 2026-06-29
Revert "Skip Quarkus CI job on the main (3.x) branch"

mskacelik:
This reverts commit a8b1bfc6e02fc0256097e641ed7a2ff393c4cf91.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,10 +87,6 @@ jobs:
           ./mvnw -B install -Dversion.io.smallrye.graphql=$SMALLRYE_VERSION --file wildfly-graphql-feature-pack/pom.xml
 
   quarkus-tests:
-    # TODO: revert when Quarkus upgrades to Vert.x 5.x
-    if: >-
-      (github.event_name == 'pull_request' && github.base_ref != 'main') ||
-      (github.event_name == 'push' && github.ref_name != 'main')
     runs-on: ubuntu-latest
     strategy:
       matrix:


jmartisk (approved):

jmartisk:
@mskacelik Hmm there's actually a compile error that seems to be related to 082b59682d54dc51162459b0d9e7562f31c0acc3 that probably means we will need a new beta release here and update Quarkus accordingly, right?

jmartisk:
It's because the `Reference.getAllParametrizedTypes` method was removed and Quarkus is using it.

mskacelik:
> It's because the `Reference.getAllParametrizedTypes` method was removed and Quarkus is using it.

Yes, we would need to use the newer API, so I guess we would need to release new Beta. With that said, what about a backport 2.x for the commit?

jmartisk:
> > It's because the `Reference.getAllParametrizedTypes` method was removed and Quarkus is using it.
> 
> Yes, we would need to use the newer API, so I guess we would need to release new Beta. 

I'll release a new beta and let you handle the Quarkus fix :)

> With that said, what about a backport 2.x for the commit?

I don't understand what you mean here. In 2.x we applied https://github.com/smallrye/smallrye-graphql/pull/2545

jmartisk:
I'm releasing 3.0.0.Beta4 (https://github.com/smallrye/smallrye-graphql/pull/2546)

mskacelik:
> I don't understand what you mean here. In 2.x we applied #2545

I meant the https://github.com/smallrye/smallrye-graphql/commit/082b59682d54dc51162459b0d9e7562f31c0acc3, being backported for 2.x.

jmartisk:
Ah ok I guess we can do that,please submit a PR :) (we'll need a release and a Quarkus change too then...)

PR #2587, opened 2026-08-18
Automatic backport for 2.x branch when setting a `triage/backport` and merging the PR into main

### LINKED ISSUE #2584: Automatic backport for 2.x branch when setting a appropriate label
It came to my attention that infinispan and wildfly-clustering projects have a integrated PR backporting feature.

https://github.com/infinispan/infinispan/blob/main/.github/workflows/backport.yaml
https://github.com/wildfly-clustering/wildfly-clustering/blob/main/.github/workflows/backport.yaml

Maybe we can take some inspiration and also use it, instead of doing periodic manual backports in batches, WDYT @jmartisk @phillip-kruger.

Also AFAIK, labels can only be applied by maintainers of SRGQL, so any security risks hopefully should not be a problem.

mskacelik:
Closes #2584

Also added files (technically a complement of a set) into `paths-ignore` - e.g. there is no need for CI jobs when something changes in the [dependabot.yml](https://github.com/smallrye/smallrye-graphql/blob/main/.github/dependabot.yml)

I set the label to `triage/backport` but maybe we can have something more explicit? (e.g. `triage/backport2.x`).

Regarding the workflow, to create automatic backport PR:
1) PR needs to be **OPEN**, target needs to be towards `main` (3.x),  and there needs to be a label of `triage/backport`.
2) After merging the PR, it will automatically (hopefully) create PR with `[2.x] {title-of-the-merged-pr}`

- Description should be a link the old PR. 
- Commits (hashes, commit names, and author(s)) should be the same. 
- Branch name should be in a format `bp-{target-branch}-{commit-sha-range}`.
- Target branch should be `2.x`

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -1,0 +1,24 @@
+name: Automated Backporting
+
+on:
+  pull_request_target:
+    types: [closed]
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  backport:
+    if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'triage/backport')
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Backport to 2.x
+        uses: kiegroup/git-backporting@v4.10.0
+        with:
+          target-branch: 2.x
+          pull-request: ${{ github.event.pull_request.url }}
+          auth: ${{ secrets.GITHUB_TOKEN }}
+          no-squash: true
+          title: "[2.x] ${{ github.event.pull_request.title }}"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,8 @@ on:
       - 'LICENSE'
       - 'NOTICE'
       - 'README*'
+      - '.github/**'
+      - '!.github/workflows/build.yml'
   pull_request:
     paths-ignore:
       - '.gitignore'
@@ -17,6 +19,8 @@ on:
       - 'LICENSE'
       - 'NOTICE'
       - 'README*'
+      - '.github/**'
+      - '!.github/workflows/build.yml'
 
 jobs:
   build:


jmartisk:
Hey, thanks @mskacelik , it looks neat but I'm not quite sure we want it at this point..
3.39 will be released very soon, then 3.40 as a continuation of it will become an LTS and we will be very conservative in what we will backport from this point forward. The number of backports will probably be so small that I don't really mind doing it manually.
Adding new actions from third parties is also something where I want to be conservative.

mskacelik:
Fair enough---I suppose that's up to you. Also, don't forget that WildFly still uses Vert.x 4.x, so it might still be worth supporting the 2.x branch for the feature pack.

On that note, I think the 55c054eaa6b23f7cd0db45da89f342a3685cc7ee commit is still valid and could help with CI usage, at least a little. Do you want me to edit the current PR, or create a new one?

jmartisk:
Hmm, that would skip CI on PRs where I ask for a release, no? That works by updating `.github/project.yml` and sending a PR with that update. If you could make sure that this file is "excluded from the exclusions" then great :) I don't care if it's  a new PR or this one :)

jmartisk:
> Also, don't forget that WildFly still uses Vert.x 4.x, so it might still be worth supporting the 2.x branch for the feature pack.

I plan to support 2.18.x, I hope we won't need a 2.19 release? But yea, who knows.
The GraphQL support in WF is a bit of a best-effort tbh. I don't know if anybody is really using it.

mskacelik:
> Hmm, that would skip CI on PRs where I ask for a release, no? That works by updating `.github/project.yml` and sending a PR with that update. If you could make sure that this file is "excluded from the exclusions" then great :) I don't care if it's a new PR or this one :)

I actually think, that this is still correct. Actual PR (e.g. https://github.com/smallrye/smallrye-graphql/pull/2553) still just changes just the yaml file, not the codebase, so I don't really see a reason for running CI jobs for it.

After the merge, update commits are pushed into main (e.g. https://github.com/smallrye/smallrye-graphql/commit/eb5b5420d3e060f6e609326e8d8bfb82acfdb52a) and they will start the jobs (since they do not trigger any of the `paths-ignore` rules).

jmartisk:
>I actually think, that this is still correct. Actual PR (e.g. https://github.com/smallrye/smallrye-graphql/pull/2553) still just changes just the yaml file, not the codebase, so I don't really see a reason for running CI jobs for it.

That's true, but generally I want to double check that the CI works properly just before starting a release. Something might have broken outside of the codebase since the last commit. For example, we have the Quarkus CI that can also break due to changes in Quarkus.

mskacelik:
That is reasonable, I will close the PR and make a new one. I also thinking of adding:
```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```
to further optimize it a bit.

PR #2615, opened 2026-08-19
Some CI optimizations

mskacelik:
I only left `.github/project.yml` for the PRs, I don't really see a reason for them in pushes, note even if I put it there, it would be canceled via `concurrency` check.


jmartisk (approved):

jmartisk:
Thanks!

